PR #41, opened 2019-04-29
Update README.md

damassi:
Makes the language a little more neutral.


sw-yx:
aww i liked the jared palmer personality in that

jaredpalmer:
It’s back now

damassi:
I mean, but why? It's just kinda aggro and doesn't add anything but edginess to something that is really just a generally useful tool. Not trying to be too PC or anything, but too much personality is distracting from the point. I think @sw-yx was just being funny.

PR #52, opened 2019-04-30
build: remove "prepare" script

aleclarson:
The "prepare" script is executed not only when `yarn` is run, but also when any dependency is installed, so its more annoying than useful.


jaredpalmer:
It's used prior to publishing. However, I agree. I use `np` anyways

PR #54, opened 2019-04-30
Convert source to TypeScript

aleclarson:


src/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { resolveApp } from './utils';
+
+export const paths = {
```
aleclarson: I didn't see where `resolveApp` can throw, so I removed the `try-catch` statement

src/createJestConfig.ts
```diff
@@ -1,5 +1,3 @@
-import { appPackageJson } from './constants';
```
aleclarson: This wasn't being used

src/env.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module 'asyncro';
+declare module 'enquirer';
+declare module 'jpjs';
+declare module 'ora';
+declare module 'sade';
+declare module 'tiny-glob/sync';
```
aleclarson: Here are the package stubs

src/getInstallArgs.ts
```diff
@@ -0,0 +1,10 @@
+import { InstallCmd } from './getInstallCmd';
+
+export default function getInstallArgs(cmd: InstallCmd, packages: string[]) {
+  switch (cmd) {
+    case 'npm':
+      return ['install', ...packages, '--save-dev'];
```
aleclarson: Using `switch` syntax on `"yarn" | "npm"` here to ensure the return type is never `undefined`

src/index.ts
```diff
@@ -43,7 +43,7 @@ try {
   appPackageJson = fs.readJSONSync(resolveApp('package.json'));
 } catch (e) {}
 
-const stat = util.promisify(fs.stat);
+const stat = util.promisify(fs.stat as any);
```
aleclarson: Strangely, `util.promisify` is not inferring the callback type of `fs.stat` correctly (`{}` instead of `fs.Stats`)

src/index.ts
```diff
@@ -196,7 +196,8 @@ prog
 
     const installSpinner = ora(Messages.installing(deps)).start();
     try {
-      await execa(getInstallArgs(getInstallCmd(), deps));
+      const installArgs = getInstallArgs(getInstallCmd(), deps);
+      await execa(installArgs[0], installArgs.slice(1));
```
aleclarson: The typings of `execa` say you can't pass an array of strings as the first and only argument. 🤷‍♂

src/index.ts
```diff
@@ -236,10 +237,7 @@ prog
         module.exports = require('./${safeVariableName(
           opts.name
         )}.cjs.development.js')
-      }`,
-          {
-            overwrite: true,
-          }
```
aleclarson: This isn't specified in `@types/node`. Is this a feature from `fs-extra`?

src/index.ts
```diff
@@ -295,7 +293,7 @@ prog
     const [cjsDev, cjsProd, ...otherConfigs] = createBuildConfigs(opts);
     if (opts.format.includes('cjs')) {
       try {
-        await mkdirp(resolveApp('./dist'));
+        await util.promisify(mkdirp)(resolveApp('./dist'));
```
aleclarson: Forgot to wrap `mkdirp` with `util.promisify`

src/index.ts
```diff
@@ -361,7 +356,9 @@ prog
       argv.push('--watchAll');
     }
 
-    const maybeTestSetupFiledExists = await fs.exists(paths.testsSetup);
+    const maybeTestSetupFiledExists = await util.promisify(fs.exists)(
```
aleclarson: Forgot to wrap `fs.exists` with `util.promisify`

src/output.ts
```diff
@@ -20,11 +20,6 @@ export const success = msg => {
   console.log(`${chalk.green('> Success!')} ${msg}`);
 };
 
-export const time = () => {
-  const start = new Date();
-  return chalk.gray(`[${ms(new Date() - start)}]`);
```
aleclarson: Took this function out cuz it looks useless

tsconfig.json
```diff
@@ -8,13 +10,8 @@
     "module": "commonjs",
     "rootDir": "src",
     "strict": true,
+    "noImplicitAny": false,
```
aleclarson: Using `noImplicitAny: false` makes it easier to convert gradually

src/index.ts
jaredpalmer: Yep

src/index.ts
jaredpalmer: fs-extra is 100% promise

src/index.ts
jaredpalmer: this isn't needed as fs-extra already promisifies everything

src/index.ts
jaredpalmer: It works tho?

aleclarson: https://github.com/sindresorhus/execa/issues/210

src/index.ts
aleclarson: The `@types/fs-extra` package seems to forward many function types through without adding promise overloads. lol

jaredpalmer: That's odd.

aleclarson: I take that back. It was just an inference issue. But the `overwrite` option is missing from the typedefs for sure.

src/index.ts
aleclarson: Added it back 👍

aleclarson:
Okay, should be ready to merge 👍

jaredpalmer:
Reinstall with yarn to update the lockfile

aleclarson:
Ack, I knew I forgot something

aleclarson:
LGTM 😎

src/index.ts
jaredpalmer: overwrite isn't an option. I think i must have copypasted from fs.copy. you can remove

aleclarson: I assume `fs.exists` is missing that promise overload because it's technically deprecated

aleclarson:
LGTM 😎

jaredpalmer:
Update from master

aleclarson:
LGTM 😎

jaredpalmer:
Great

PR #75, opened 2019-05-02
Write Jest configuration file

PabloSzx:
Needed for compatibility with [Visual Studio Code - Jest extension](https://github.com/jest-community/vscode-jest)


agilgur5:
This was removed at some point (not sure when) and needs to be re-added in a non-buggy way: see #270 

@allcontributors please add @PabloSzx for code, bug

PR #81, opened 2019-05-03
Turn off size snapshot

jaredpalmer:


src/constants.ts
```diff
@@ -1,9 +1,21 @@
...
+  appDist: '',
 };
+
+try {
+  paths = {
+    appPackageJson: resolveApp('package.json'),
+    testsSetup: resolveApp('test/setupTests.ts'),
+    appRoot: resolveApp('.'),
+    appSrc: resolveApp('src'),
+    appDist: resolveApp('dist'),
```
aleclarson: How can [`resolveApp`](https://github.com/palmerhq/tsdx/blob/e2d6ba3f64119dc2c87838b8bdf0ffb808a66595/src/utils.ts#L22-L27) throw?

jaredpalmer: Yeah you are right it can't.

PR #86, opened 2019-05-06
Budget 1 char for +/- from diffs

agentofuser:
When looking at git diffs on 80char terminals it's helpful to have lines not
break when they are exactly 80 chars and are prepended with a + or -.


jaredpalmer:
I think this is outside of scope of the package. It is also very easy to augment in user land.

PR #89, opened 2019-05-08
fix external module detection on Windows

### LINKED ISSUE #87: No name was provided for external module
### Current Behavior

My code isn't getting bundled because tsdx treats everything as external modules, the resulting bundle only contains the compiled `index.ts` file.

### Expected behavior

My files shouldn't be treated as external modules.

### Suggested solution(s)

Could it be something related to the `external` function? https://github.com/palmerhq/tsdx/blob/v0.5.7/src/utils.ts#L19

### Additional context

I'm getting the following warning for all my `.ts` files (except the root `index.ts` file):
```
No name was provided for external module 'C:/xyz/xyz/xyz/xyz.ts' in output.globals – guessing 'xyz_ts'
```

### Your environment

| TSDX           | 0.5.7
| TypeScript       | 3.4.5
| Browser          | Chrome Version 74.0.3729.131
| npm/Yarn         | Node v10.11.0, npm 6.4.1
| Operating System | Windows 10

a-type:
Fixes #87 

Absolute path detection was not working for Windows users. I updated the util to use Node's built-in function.

I also created a regression test in the testing suite to ensure that exports defined in other files are correctly included in the final library build.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -17,7 +17,7 @@ export const safePackageName = (name: string) =>
   name.toLowerCase().replace(/((^[^a-zA-Z]+)|[^\w.-])|([^a-zA-Z0-9]+$)/g, '');
 
 export const external = (id: string) =>
-  !id.startsWith('.') && !id.startsWith('/');
+  !id.startsWith('.') && !path.isAbsolute(id);
 
 // Make sure any symlinks in the project folder are resolved:
 // https://github.com/facebookincubator/create-react-app/issues/637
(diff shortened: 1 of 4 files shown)


a-type:
Another problem that's related: the default `tsconfig.json` adds `src/*` to the `paths`, but doing so means that TS will resolve auto-imported modules relative to `src` (so if I auto-imported something from `src/utils.ts`, it would come in as `import { foo } from 'utils'`. This causes these modules to also be marked as external. That's likely to confuse people. Either the default `tsconfig` shouldn't use the absolute resolution in `src`, or the `external` check will need to be made a bit smarter. I consider that outside the scope of this particular issue (Windows support out of the box).

jaredpalmer:
Nice.

agilgur5:
@allcontributors please add @a-type for code, test, bug

PR #93, opened 2019-05-09
add "yarn build" for example

slorber:
having a yarn build command makes it easier to deploy example to CI, as start is watch mode


agilgur5:
@allcontributors please add @slorber for code

PR #96, opened 2019-05-09
Explicitly specify ts and tsx extensions in babel options

jakegavin:
Fixes https://github.com/palmerhq/tsdx/issues/95

>rollup-plugin-typescript2 transpiles the code but doesn't change the
file extensions. Files with ts and tsx extentions aren't processed by
Babel by default so we need to add those extensions to the babel config.

I'm not actually using tsdx on a production app (yet? we need to be able to configure babel though 😕) so I don't have a great way to test this with confidence for a real world project. 

Also I'm seeing...
> Module '"../../../../../../Users/jake/Code/jakegavin/tsdx/node_modules/@types/babel__core"' has no exported member 'DEFAULT_EXTENSIONS'

But its [there](https://github.com/babel/babel/blob/1969e6b6aa7d90be3fbb3aca98ea96849656a55a/packages/babel-core/src/index.js#L39-L45). I guess it's maybe not in the [babel core types](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/babel__core). I have a PR up to add it: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35336

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -1,3 +1,4 @@
+import { DEFAULT_EXTENSIONS } from '@babel/core';
 import { safeVariableName, safePackageName, external } from './utils';
 import { paths } from './constants';
 import { terser } from 'rollup-plugin-terser';
@@ -14,6 +15,7 @@ const replacements = [{ original: 'lodash', replacement: 'lodash-es' }];
 
 const babelOptions = {
   exclude: /node_modules/,
+  extensions: [...DEFAULT_EXTENSIONS, 'ts', 'tsx'],
   plugins: [
     'annotate-pure-calls',
     'dev-expression',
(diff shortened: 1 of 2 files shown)


sw-yx:
hmm. leaving this one for jared.

jaredpalmer:
Will have to test in the morning

jaredpalmer:
Fixed the type issue by patching the types. DEFAULT_EXTENSIONS is indeed exported.
See Line 226 https://unpkg.com/@babel/core@7.4.4/lib/index.js

jaredpalmer:
@jakegavin thanks! Can confirm that lodash opt was not running. For reference, in Formik, we don't use `rollup-plugin-typescript2`, but instead we use plain old `tsc` to spit out  an `esnext` .js module to the `compiled` directory, we then run rollup (with your babel optimizations) against that JavaScript and then copy the types to that output directory...this means i never had to deal with the extensions bit. Great catch!

PR #97, opened 2019-05-10
Document some of the optimizations

### LINKED ISSUE #8: Document how to use lodash/lodash-es optimization

jaredpalmer:
First pass at documenting the built-in optimizations. Did not proof read. So feel free to nitpick typos 

cc @sw-yx 

Closes #8.


sw-yx:
very good writeup! if we were to do the fb minified code thing, we may have to fork `babel-plugin-dev-expression` so that it doesnt quite strip away the arguments, but minifies and appends them to a file instead somehow (probably based on a hash). do i have that intuition right?

PR #98, opened 2019-05-10
document rollup config treeshaking

sw-yx:
i had this in my React Guide PR but it belongs at the top level - submitting as PR for review


jaredpalmer:
word.

PR #99, opened 2019-05-11
Add tsdx lint command

skvale:
- Add `tsdx lint` command
- Use `tsdx lint` to lint itself


jaredpalmer:
Sickkkkk

jaredpalmer:
This looks great. Will give it try in the morning

src/createEslintConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+    env: {
+      node: true,
+      es6: true,
+      jest: true,
+    },
+    parserOptions: {
+      ecmaVersion: 8,
+      ecmaFeatures: {
+        impliedStrict: true,
+      },
```
skvale: This will need to be updated to work with `jsx` syntax in React apps still.

src/createEslintConfig.ts
```diff
@@ -0,0 +1,74 @@
...
+  rootDir,
+  writeFile,
+}: CreateEslintConfigArgs): CLIEngine.Options['baseConfig'] {
+  const config = {
+    extends: [
+      'eslint:recommended',
+      'plugin:@typescript-eslint/recommended',
+      'prettier/@typescript-eslint',
+      'plugin:prettier/recommended',
+      'plugin:react/recommended',
```
skvale: It'd be nice to have an override for `.tsx` files and extend the React specific plugins only for those files.  Would need https://github.com/eslint/eslint/pull/11554 to do that though.

jaredpalmer: I think we just include the react config no matter what.

jaredpalmer:
@skvale I think these rules are a bit too tight. Can we just use `eslint-config-react-app` ?

skvale:
@jaredpalmer I made `createEslintConfig` return
```ts
{
  extends: ['react-app'],
};
```
For tsdx, I ran `yarn lint --write-file` and updated the `.eslintrc.js` to
```js
module.exports = {
  extends: [
    'react-app',
    'prettier/@typescript-eslint',
    'plugin:prettier/recommended',
  ],
};
```
We could document the ways of customizing the rules/extensions through
- `.eslintrc.js`
- `packageJson.eslint`

in the README

jaredpalmer:
Since we install prettier for people, we should setup eslint to play nicely? Does eslint config react app do that?

skvale:
Yep, I'm pretty sure prettier and eslint-config-react-app work well together.

How about adding a `--prettier` optional flag to lint, like `yarn lint --prettier` instead of having to generate a `.eslintrc.js` file or add the rules to the package.json? 4332881

jaredpalmer:
Run it by default. The less flags the better

skvale:
This could wait until eslint@6 is released, since that should come out soon.

jameslnewell:
🎉Nice! 

Can we add a `--watch` option? 

What's the thinking behind providing a separate `lint` command vs running the linter on build or test? (I'd prefer to discover and fix linting issues while I'm working in the file where the problems exist rather than having to come back after I've edited a bunch of files).

skvale:
Waiting for 
- https://github.com/facebook/create-react-app/commit/24489ac0a667af416f1d59dd806dfc2623aabe36 
and 
- https://github.com/typescript-eslint/typescript-eslint/pull/628
to be released to support eslint@6

jaredpalmer:
Let's resolve conflicts and get this merged?

skvale:
@jaredpalmer Sounds good
I changed the [config](https://github.com/palmerhq/tsdx/pull/99/files#diff-e248403f681f0d68a60b6846a5d0d694R26) to account for https://github.com/facebook/create-react-app/commit/24489ac0a667af416f1d59dd806dfc2623aabe36

and bumped to `"@typescript-eslint/parser": "^1.10.3-alpha.13"` for the issue from https://github.com/typescript-eslint/typescript-eslint/pull/628

package.json
```diff
@@ -35,8 +37,11 @@
     "@babel/plugin-proposal-class-properties": "^7.4.4",
     "@babel/polyfill": "^7.4.4",
     "@babel/preset-env": "^7.4.4",
+    "@typescript-eslint/eslint-plugin": "^1.12.0",
+    "@typescript-eslint/parser": "^1.12.0",
     "ansi-escapes": "^3.2.0",
     "asyncro": "^3.0.0",
+    "babel-eslint": "^10.0.2",
```
leohxj: does this necessary? 
I think use the `@typescript-eslint/parser` parser is enough.

skvale:
@leohxj I had to do two things to get eslint to work in vscode.  This page for create-react-app explains it well too: https://facebook.github.io/create-react-app/docs/setting-up-your-editor
- `tsdx lint --write-file` <-- generate the local config file
- Add typescript to my settings file in `eslint.validate`
```json
  "eslint.validate": [
    "javascript",
    "javascriptreact",
    "typescript",
    "typescriptreact"
  ],
```

jaredpalmer:
Is this g2g?

skvale:
yeah, I think so

jaredpalmer:
Not for this PR, but we may want to think through some kind of upgrade and/or validation command. This would check package.json, dotfile config, tsconfig, eslint etc. and fill in (and/or reset) any missing pieces. IIRC, jest has something like this called like jest-validate?.

PR #102, opened 2019-05-11
@babel/preset-env

### LINKED ISSUE #101: Output not being compiled-down
### Current Behavior

Currently the output is not being transpiled so it can be supported on older browsers eg. `class {}`.

### Suggested solution(s)

Use `@babel/preset-env` and default to ES2015 output

### Additional context

It's problematic because some setups don't transpile node_modules. This prevents extending classes from `tsdx` as babel doesn't like es5->es6 interop.

samdenty:
Fixes #101

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -16,6 +16,7 @@ const replacements = [{ original: 'lodash', replacement: 'lodash-es' }];
 const babelOptions = {
   exclude: /node_modules/,
   extensions: [...DEFAULT_EXTENSIONS, 'ts', 'tsx'],
+  presets: ['@babel/preset-env'],
   plugins: [
     'annotate-pure-calls',
     'dev-expression',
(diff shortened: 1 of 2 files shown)


jaredpalmer:
This was fixed with #111

jaredpalmer:
Nvm this is different. I need to think this through

jaredpalmer:
Going to merge and then make this more robust in another PR

src/createRollupConfig.ts
```diff
@@ -16,6 +16,7 @@ const replacements = [{ original: 'lodash', replacement: 'lodash-es' }];
 const babelOptions = {
   exclude: /node_modules/,
   extensions: [...DEFAULT_EXTENSIONS, 'ts', 'tsx'],
+  presets: ['@babel/preset-env'],
```
skvale: Should this have been `presets: [require.resolve('@babel/preset-env')],`?

Otherwise I am getting
```
UnhandledPromiseRejectionWarning: Error: Cannot find module '@babel/preset-env' ...
```
in the package that consumes tsdx

jaredpalmer:
Yep

PR #115, opened 2019-05-28
Create .npmignore

### LINKED ISSUE #105: New project includes yarn-error.log
Steps to recreact
run: npx tsdx create mylib
cd mylib
cat yarn-error.log

See that jared is running 11.8.0 and had a syntax error in /Users/jared/workspace/github/jaredpalmer/tsdx/templates/react/package.json

huchenme:
This does not resolve #105 but could prevent `yarn-error.log` get added in accidentally in future

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,1 @@
+yarn-error.log


TrySound:
yarn-error.log is never published in tsdx package.
https://github.com/palmerhq/tsdx/blob/master/package.json#L28-L31

You need to add `files` to generated `package.json`.

huchenme:
hmm, it got published in [current version](https://unpkg.com/tsdx@0.5.11/templates/react/) and the file is in `templates` folder

TrySound:
I see. It's because templates is included to files. Then this solution is probably ok. Will think more.

jaredpalmer:
This was fixed in the latest release

Issue #116, opened 2019-05-29
Example in React template throws cannot resolve dependency 'react'


kirjai:
### Current Behavior

When creating a new library with `tsdx`, by running `npx tsdx create some-name` and choosing the `react` template, when attempting to run the example, the build fails with a `Cannot resolve dependency 'react'` error.

### Expected behavior
Build should not fail.

### Suggested solution(s)
From what I understand, because the `package.json` inside `example` defines `react` and `react-dom` as aliases to the root directory `node_modules`, and the root `package.json` defines `react` as a peer dependency, we end up in a situation where react is never installed, hence the cannot resolve dependency error.

Reverting the changes introduced in #88 locally, solves the problem. So removing the aliases and defining `react` and `react-dom` as dependencies inside `example/package.json` works as expected.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | `0.5.12`
| TypeScript       | `3.4.5`
| Browser          | n/a
| npm/Yarn         | `1.9.4`
| Operating System | Mac OS `10.14.2`

sw-yx:
hello, the reason we did #88 this is because we need to ensure the same instance of react inside `example` as we do in the hooks lib.

i think the real fix is to have `react` and `react-dom` as devDeps as well as peerDeps at the top level. I'll do that.

sw-yx:
i just tested a freshly created library from v0.5.12, using `yarn` it seems fine. so perhaps this is an `npx` issue.

kirjai:
Thanks for looking into this.

So I did some digging just now, prompted by your comments, and realized that when i was trying to replicate this issue, I was doing so inside a mono repo, managed by `lerna`. So looks like *that's* the reason the `alias` approach doesn't work for me.

As far as i understand, in a mono repo, some dependencies are installed in the root of the repo, that's why `react` and `react-dom` are actually missing from the `node_modules` in a fresh directory created by `tsdx`, they are present in the mono repo root `node_modules`. Therefore, `react` and `react-dom` aren't available in the location point to by the `alias`.

--- 

So not sure if this is a use case that `tsdx` cares about or should care about, but in case it's not, perhaps it's worth to mention this and the potential workaround in the `react` template README?
I'm happy to submit a PR for that if you it would be worth it.

sw-yx:
sure, PR welcome. basically users of tsdx should feel comfortable enough modifying the templated pkg.json to their needs, i dont think we wanna support every use case. but i'm happy to document it.

PR #119, opened 2019-05-30
Document react template usage with lerna

kirjai:
As discussed in #116, added documentation to mention a known caveat when using the react template with lerna


jaredpalmer:
@sw-yx going to open another issue in tsdx about making a monorepo template as well as upgrade palmerhq/monorepo-template too

agilgur5:
@allcontributors please add @kirjai for docs

Issue #114, opened 2019-05-26
Broken build with `tsdx build --format es`


techniq:
### Current Behavior

In an attempt to speed up my builds (and not need the other formats), I attempted to only build for `es` by changing the `build` script in `package.json` to `tsdx build --format es` but in doing so, it fails with

> (node:89396) UnhandledPromiseRejectionWarning: Error: You must supply an options object to rollup

![image](https://user-images.githubusercontent.com/177476/58387234-ac7cc680-7fd9-11e9-83c9-727c368edcca.png)

I thought this might be fixed by this [commit](https://github.com/palmerhq/tsdx/commit/71817a33628983ee8a3adb4d10e852b5895d7a00) in [0.5.10](https://github.com/palmerhq/tsdx/releases/tag/v0.5.10) but even on `0.5.11` it still shows the issue (I did a fresh project with `npx tsdx create tsdx-test` to verify it wasn't anything in my project specifically).

### Expected behavior

To be able to only build a specific format (and possibly any other way so to optimize).  My builds jumped from less than 2 minutes for the full pipeline (not just `npm run build`) to between 7-8 minutes after changing to tsdx (along with migrating to 99% typescript from ~30% prior).  I could see having more Typescript code causing it to be a little slower, but not to this magnitude.  I was only performing 1 build before, so I'm hoping going back to that will improve my `tsdx` build times.

### Suggested solution(s)

I thought this [commit](https://github.com/palmerhq/tsdx/commit/71817a33628983ee8a3adb4d10e852b5895d7a00) was the fix, but not sure where the issue is now.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | `0.5.11`
| TypeScript       | `3.1.6`
| Browser          | `Chrome`
| npm         | `6.8.0`
| Operating System | `Mac`

techniq:
The root issue appears to be if you exclude `cjs` from the formats, so `tsdx build --format cjs,es` works, but not just `tsdx build --format es` due to `inputOptions` being null [here](https://github.com/palmerhq/tsdx/blob/cce9d53cc88fd22675e20697619661738cdc07b5/src/index.ts#L354).

Excluding `umd` using `--format=cjs,es` has dropped a minute or so off my builds, which helps.

jbadeau:
`--format=none`

would also be useful

sw-yx:
@techniq it sounds like you're pretty close to a solution, want to propose a fix?

[...]

PR #125, opened 2019-05-30
Fix: Broken build with `tsdx build --format es`

enesTufekci:
I have added `development` build for `es` format and it fixed the bug #114

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -353,7 +353,6 @@ prog
         .map(
           [cjsDev, cjsProd, ...otherConfigs],
           async (inputOptions: RollupOptions & { output: OutputOptions }) => {
-            console.log(inputOptions);
             let bundle = await rollup(inputOptions);
             await bundle.write(inputOptions.output);
             await moveTypes();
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -351,6 +353,7 @@ prog
         .map(
           [cjsDev, cjsProd, ...otherConfigs],
           async (inputOptions: RollupOptions & { output: OutputOptions }) => {
+            console.log(inputOptions);
```
jaredpalmer: Remove

src/index.ts
```diff
@@ -91,6 +91,8 @@ function createBuildConfigs(opts: any) {
         createRollupConfig('cjs', 'development', { ...opts, input }),
       opts.format.includes('cjs') &&
         createRollupConfig('cjs', 'production', { ...opts, input }),
+      opts.format.includes('es') &&
```
jaredpalmer: This would need an entry point like `index.es.js` just like `index.js`

jaredpalmer:
Closing in favor of #128

Issue #122, opened 2019-05-30
Add monorepo template


natemoo-re:
### Current Behavior

Currently, users can scaffold out a monorepo manually, then add `tsdx` as a dependency at the root of the project.

### Desired Behavior

Users should be able to scaffold out a monorepo project using `npx tsdx create`. Just like the current templates, users could pass a flag or select from the interactive prompt.

### Suggested Solution

Create two new templates, `monorepo-basic` and `monorepo-react`. In both cases, the only additional dependency would be `lerna`. The root `package.json` scripts would use `lerna run command` and any packages would use `tsdx command`.

1) Add these options directly to the `create` template prompt and add CLI flags for each.

2) Add an additional prompt to the `create` command which allows you to select a project structure (`basic` or `monorepo`?) and add a `—monorepo` CLI flag which will automatically resolve the monorepo templates.

### Who does this impact? Who is this for?

This feature would be very beneficial for anyone managing a larger project who wishes to leverage both `tsdx` and `lerna`.

### Describe alternatives you've considered

If this use case seems out of scope, improving the documentation around integrating with `lerna` would be helpful to guide users through manual setup. 

As a possible alternative, `tsdx` could become monorepo-aware so that running `npx tsdx create mylib` from within a monorepo could scaffold out a new package in the `packages/` (or configured in `lerna.json`) directory.

### Additional context

I have created a minimal [tsdx-monorepo](https://github.com/natemoo-re/tsdx-monorepo) example project.

I would be happy to contribute to this feature, but I want to collect feedback and hash out the preferred approach before starting.

jaredpalmer:
Update: I just moved formik into a monorepo powered by tsdx (with both React native and react packages). Should be a good test bed. We DEF want to bake this in because it was huge pain to setup with VSCode.

sw-yx:
what were the pain points?

thaiphan:
@jaredpalmer Is that something you can blog about?

ambroseus:
@thaiphan  example of using `tsdx` in lerna monorepo : https://github.com/DCKit/dckit

mwarger:
@jaredpalmer - any updates on the baked-in monorepo setup?  Interested to see what you encountered...  Thanks.

pcowgill:
> Update: I just moved formik into a monorepo powered by tsdx (with both React native and react packages). Should be a good test bed. We DEF want to bake this in because it was huge pain to setup with VSCode.

[...]

PR #127, opened 2019-05-31
feat(monorepo): improve monorepo support

natemoo-re:
Related to #122. Feedback welcome!

**Goals**
* Simplify integration of new templates
  * Remove hardcoded template references in `src/index.ts`
* Introduce project structures (`basic`, `monorepo`)
* Prompt `tsdx create` users to select project structure (if template supports it)
* Automatically determine if a template supports `monorepo` structure
  * `basic` template and `monorepo` template must share a basename
  * `monorepo` template should be suffixed with `-monorepo`

**Ideas**
* Allow `tsdx create` (or new command?) to scaffold out a new package within an existing monorepo
  * How to determine if run inside a monorepo, where to place package?


aleclarson:
> How to determine if run inside a monorepo, where to place package?

Maybe by finding all directories with a `package.json` and prompting the CLI user to choose between their parent directories? The other route is supporting only Lerna and Yarn workspaces, and checking their configuration.

natemoo-re:
Both good ideas.

As of right now, this PR adds `lerna` to any monorepo created with `tsdx create`. Yarn workspaces are great, but using `lerna` avoids forcing any package manager lock-in.

If that remains the case, I'd lean towards checking `lerna.json` `packages` and prompting the CLI user to select one (if there are multiple.)

natemoo-re:
The current `react` template includes a playground at `/mylib/example`. For the `react-monorepo` template, should the playground be moved to `/examples/basic`? This seems like a common pattern (see [jest](https://github.com/facebook/jest), [storybook](https://github.com/storybookjs/storybook), [next.js](https://github.com/zeit/next.js)).

sw-yx:
no opinion, go for it :)

natemoo-re:
I'll be circling back to this shortly. Does cutting off this PR at adding `basic-monorepo` and `react-monorepo` templates sound okay for a first pass?

I'd like to add support for `tsdx create` within existing monorepos, but I think that can be split out into a separate issue to keep things moving.

jaredpalmer:
What's the status of this?

package-lock.json
```diff
@@ -0,0 +1,7824 @@
+{
+  "name": "tsdx",
```
jaredpalmer: Remove this file. We are on #TeamYarn

templates/react-monorepo/packages/core/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "main": "dist/index.js",
+  "umd:main": "dist/core.umd.production.js",
+  "module": "dist/core.es.production.js",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "start": "tsdx watch",
+    "build": "tsdx build"
```
jaredpalmer: What about test and lint cmd?

PR #130, opened 2019-06-01
feat: Add success/failure hooks to watch

hedgerh:
Resolves https://github.com/palmerhq/tsdx/issues/104

Adds flags to the `tsdx` CLI for running commands on success, first success, and failure of builds when running `watch`.  Usage:

```bash
tsdx watch --onFirstSuccess "echo First successful build" 
tsdx watch --onSuccess "echo Successful build"
tsdx watch --onFailure "echo The build failed" 
```

Still thinking about how to test these, and maybe even the `watch` command in general.

**PS.** I also added a `start` command in the `tsdx` project to run tsc in watch mode to re-bundle on changes.


jaredpalmer:
I agree we should add this. Can you fix the conflicts?

jaredpalmer:
Let's fix conflicts and merge. This would be super useful for a bunch of stuff.

sw-yx:
i think i have rebased

jaredpalmer:
lgtm.

agilgur5:
So the `esnext` target override was changed in this PR -- is that safe?? It doesn't seem related to the other changes.

I'm not 100% sure why the `esnext` target override was set as such to begin with (it seems to exist since some of the first commits / original versions of TSDX), but I _thought_ it existed so that TS would minimally make changes and then Babel (via `preset-env`) would further transpile to the target. I'm not really sure if that assumption is correct, but I think it did that before, so this is potentially breaking if correct.

If that's not correct, this still now makes the `tsconfig` `target` option actually applied, which is a non-trivial change as well and I think also potentially breaking (e.g. if the newly applied target _doesn't_ transpile some features that _were_ transpiled before), but I'm not 100% sure about how the interaction with `preset-env` works.

agilgur5:
@jaredpalmer so this got released in 0.12.1 -- could you comment on the above, re: safety/potential breaking-ness of changing the target override?

sw-yx:
thanks for the great PR! @all-contributors please add @hedgerh for ideas, docs, code, tests

agilgur5:
Argh, found while looking at #669 that this ended up adding these lines to `watch`:

```ts
    opts.name = opts.name || appPackageJson.name;
    opts.input = await getInputs(opts.entry, appPackageJson.source);
```

Which are redundant with the `normalizeOpts` function that is called above it 😕

PR #138, opened 2019-06-10
Add error code extract and transform

jaredpalmer:
Closes https://github.com/palmerhq/tsdx/issues/78

- Rip out React's error workflow
- Add rollup and babel transforms that extract and swap errors
- Creates `./errors/codes.json` in the root of tsdx directory on the first run, anything that uses a function call of `invariant` will be extracted. 
- Outputs dev and prod error message functions at `./errors/ErrorDev.js` and `./errors/ErrorProd.js`. Once generated, user can freely edit them. To regenerate, just delete and run a build or watch.
- Add a flag to `build` called `--extractErrors`


sw-yx:
testing this out

src/errors/extractErrors.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  const errorMapFilePath = opts.errorMapFilePath;
+  let existingErrorMap: any;
+  try {
+    // Using `fs.readFileSync` instead of `require` here, because `require()`
+    // calls are cached, and the cache map is not properly invalidated after
+    // file changes.
+    existingErrorMap = JSON.parse(
+      fs.readFileSync(
+        path.join(__dirname, path.basename(errorMapFilePath)),
```
sw-yx: this is a bug, we should just read from `errorMapFilePath`, exactly how we write.

jaredpalmer:
We need to generate the files and roll it up so it either needs to be a separate runtime package or  we need to inject it So it can be a relative import. The reason we need to generate it is so we can say “Minified Blank” error and have a custom url. Putting it in a dot folder could cause confusion. One possibility is for use to make an errors directory outside of src where this all lives.

sw-yx:
i dont feel strongly enough about it to do anything tbh :) anyway i think this is good to go. can work on more docs in a separate issue.

jaredpalmer:
Slept on it. I think we should move everything to an errors directory

sw-yx:
i still want this i think. any blockers in your mind? i can resolve merge conflicts later

jaredpalmer:
I just not sure it should be the default? Then again, it only does magical stuff if you use invariant() so idk. Maybe we just document the crap out of it

sw-yx:
i think it is already opt in tho? you made that change. needs people to actively run the errorExtraction flag. 

and maybe add a bit more logging so people dont get surprised there is a new file.

jaredpalmer:
Lol I totally forgot. I've been crazy busy working on Formik 2 + the thing I that's not out yet. Yeah let's merge this lil guy

jaredpalmer:
@sw-yx made some changes. I think this is g2g

sw-yx:
Formik World

PR #141, opened 2019-06-12
Create ES dev bundle same as CJS

### LINKED ISSUE #140: Production bundle for ES format is not enough for a development
### Current Behavior

Webpack will use ES format but being in a development mode it means the wrong assumption from 3rd modules like `invariant` which then throws an error.

image

https://github.com/zertosh/invariant/blob/fae1c859f3eec25025dc6019bf7f7fdb78a27307/invariant.js#L26

### Expected behavior

Same behavior as for CJS should be used and provide development bundle along with `index.es.js`.

### Suggested solution(s)

Should be pretty straightforward to add a line in here...

https://github.com/palmerhq/tsdx/blob/943648edb06cd131f2c1751120cd1301804b480b/src/index.ts#L98

And utilize logic for creating index file...

https://github.com/palmerhq/tsdx/blob/943648edb06cd131f2c1751120cd1301804b480b/src/index.ts#L277

### Your environment

| TSDX           | 0.6.1
| TypeScript       | 3.5.1
| Browser          | N/A
| Yarn         | 1.16.0
| Operating System | Windows x64

FredyC:
Fixes #140 

Tested it against a project that has shown the problem and works correctly there.

Btw: It seems that tests setup is wrong, no test file is executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ export const sum = (a: number, b: number) => {
 };
 ```
 
-`tsdx build` will output an ES module file and 3 CommonJS files (dev, prod, and an entry file). If you want to specify a UMD build, you can do that as well. For brevity, let's examine the CommonJS output (comments added for emphasis):
+`tsdx build` will output an 3 files for CommonJS and ES formats (dev, prod, and an entry file). If you want to specify a UMD build, you can do that as well. For brevity, let's examine the CommonJS output (comments added for emphasis):
 
 ```js
 // Entry File
--- a/src/index.ts
+++ b/src/index.ts
@@ -388,9 +388,9 @@ prog
         [
           `'use strict'`,
           `if (process.env.NODE_ENV === 'production') {`,
-          `${baseLine}.production.js`,
+          `${baseLine}.production.js')`,
           '} else {',
-          `${baseLine}.development.js`,
+          `${baseLine}.development.js')`,
           '}',
         ].join('\n')
       );
--- a/src/index.ts
+++ b/src/index.ts
@@ -386,11 +386,11 @@ prog
       return fs.writeFile(
         resolveApp(`./dist/index${format !== 'cjs' ? `.${format}` : ''}.js`),
         [
-          'use strict;',
+          `'use strict'`,
           `if (process.env.NODE_ENV === 'production') {`,
-          `${baseLine}.production.js;`,
+          `${baseLine}.production.js`,
           '} else {',
-          `${baseLine}.development.js;`,
+          `${baseLine}.development.js`,
           '}',
         ].join('\n')
       );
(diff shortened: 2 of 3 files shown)


jaredpalmer:
I don’t think any bundler is setup to use this build though. Only one file can be noted as the esm entry in pkg.json

--
Jared

________________________________

FredyC:
@jaredpalmer Well, the assumption is that you will use `index.es.js` in the `module` field.

TrySound:
This way you may end up with a lot of duplicated code in your bundle.

FredyC:
@TrySound Well yea, but how does that matter? Each bundler/environment will pick whatever they need. It's not like you will use all of them at once.

UMD is the ultimate copy and would forbid for that format to exists if I could :)

FredyC:
Ah, finally it's green. Please squash it, I got a bit overboard with commits here 😎

TrySound:
The point is to drop commonjs completely. It should not be used to import esm. It will break a lot of stuff.

FredyC:
@TrySound That's a bad idea. NodeJS is still not fully ESM capable, especially for people stuck at V8 or something. You cannot ask of them to transpile 3rd modules.

TrySound:
I'm not gonna ask to transpile. Here's a kind of perfect build I recommend.
https://unpkg.com/react-beautiful-dnd@11.0.4/dist/
```js
cjs -> NODE_ENV is not replaced. Here we have react like development/production bundles. It's ok.
esm -> NODE_ENV is not replaced. It should be replaced by user bundler. No duplication for user / no shit from mixing commonjs and esm
umd -> NODE_ENV is replaced with development
min -> NODE_ENV is replaced with production and bundle minified to show kind of final size
```

jaredpalmer:
You cannot conditionally import ESM the way can conditionally require. I confirmed this with @developit. Thus, there is no use case for a ES dev build.

TrySound:
@jaredpalmer Where is the simplifity of early days rollup config? It's full of unreadable conditions now. And how you want users to build their apps with production mode messages?

FredyC:
> You cannot conditionally import ESM the way can conditionally require. I confirmed this with @developit. Thus, there is no use case for a ES dev build.

I see what you mean. That's bad... How to solve the original issue then? It breaks the development builds if a production build is used.

TrySound:
Just remove `replace` plugin from esm pipeline and remove `production`/`development` suffix from esm bundle name.
@jaredpalmer Look, it was perfect before.
https://unpkg.com/formik@1.5.7/dist/formik.esm.js

FredyC:
@TrySound I see what you mean. That makes sense and I assume it's the way microbundle does since it does not have such if/else logic in the output.

Oh well, should have been thinking more before diving into this PR. I suppose let's close it and do it with a clean slate.

FredyC:
Btw, I wonder, does it actually work by default that Webpack would run replace on NODE_ENV? It's been quite a while I've configured Webpack by myself and I do remember some awkward DefinePlugin or whatnot.

TrySound:
@FredyC webpack 4 has `mode=production|development` which uses DefinePlugin internally. It just works.

PR #142, opened 2019-06-12
No env specific bundle for ESM

### LINKED ISSUE #140: Production bundle for ES format is not enough for a development

FredyC:
Fixes #140 

An alternative (and correct) approach based on a discussion in #141 

~~I've kept `index.js` which points to a single `name.cjs.js` file to keep breaking changes at a minimum. It could probably be removed eventually.~~

Besides some slight cleanup, the major change is a single ESM build that isn't env specific. Production bundles also get the `.min` in the name.


FredyC:
Btw, I've noticed that for `watch` command it includes `umd` for format defaults while in `build` only `cjs,es` is by default. Looks like an oversight. Shall I fix it here as well?

src/index.ts
```diff
@@ -267,23 +265,9 @@ prog
     opts.name = opts.name || appPackageJson.name;
     opts.input = await getInputs(opts.entry, appPackageJson.source);
     const buildConfigs = createBuildConfigs(opts);
+    await util.promisify(mkdirp)(resolveApp('dist'));
```
FredyC: I haven't fully understood why this line was originally inside the condition. Why it shouldn't be creating dist folder no matter of what format is chosen?

FredyC:
Perhaps I should also rename those umd bundles as drafted in https://github.com/palmerhq/tsdx/pull/141#issuecomment-501420287? It would be breaking change unless making alias files for that as well.

On the other hand, having `.min` (or not) does not tell the story that it's actually a production build. If something is minified is not that relevant and it's apparent from the first look at the file.

TrySound:
I think UMD story similar to react is ok too.

TrySound:
I would recommend to rename `es` into `esm`. Rollup supports both formats. `es` has different meaning in some projects. For example material-ui uses it for es6 syntax (not only for esm).

FredyC:
@TrySound Well, in that case I would have to change the whole `--format esm` option which is another breaking change. Is that ok?

TrySound:
Yeah, both umd and cjs are ok. The problem only esm bundle right now. @FredyC I think from tsdx point of view it can support both. `es` is an alias.

FredyC:
Alright, another version. The CJS is back in both dev/prod builds. It also adds `min` to a production builds. The ES format now outputs like `dist/build-default.esm.js` without running replace. It's possible to input format as "es" or "esm", it gets normalized.

src/createRollupConfig.ts
```diff
@@ -74,9 +88,7 @@ export function createRollupConfig(
     // Establish Rollup output
     output: {
       // Set filenames of the consumer's package
-      file: `${paths.appDist}/${safePackageName(
-        opts.name
-      )}.${format}.${env}.js`,
+      file: outputName,
       // Pass through the file format
       format,
```
TrySound: You may add the same condition here to highlight migration to `esm`

FredyC: Well, I am not sure to call it a migration. Or at least babel plugin typings expect the "es" only. Strangely enough, the rollup expects "esm" in the format option. I wonder how is it possible it actually worked till now. It's a bit of mess.

https://rollupjs.org/guide/en#command-line-flags

TrySound: Babel do not use `es` value anywhere. The type is defined in this module.

FredyC: Format is being passed to Babel in here ... https://github.com/palmerhq/tsdx/blob/943648edb06cd131f2c1751120cd1301804b480b/src/createRollupConfig.ts#L163

Either way, I've used the same condition for a rollup format.

TrySound: I mean babelOptions is defined here too. Format is only used to check non-commonjs
https://github.com/palmerhq/tsdx/blob/d111c131a5e5cf19d62b98e123060c2ab55c43da/src/createRollupConfig.ts#L16-L50

FredyC: Doh! How could I've missed that :) Ok, it's now consolidated as "esm" everywhere with a check for "es" to avoid breakage.

FredyC:
So...anything else missing here? Would like to get it out soon-ish so we can publish a fixed package. I am kinda surprised that nobody else noticed it yet considering that Webpack grabs the ESM build by default.

TrySound:
Latest release of formik uses old build. I guess that's why no one noticed.

src/index.ts
```diff
@@ -383,6 +345,40 @@ prog
...
+function ensureDistFolder() {
+  return util.promisify(mkdirp)(resolveApp('dist'));
+}
+
+function writeCjsEntryFile(name: string) {
+  // prettier-ignore
+  const baseLine = `  module.exports = require('./${safePackageName(name)}`;
+  return fs.writeFile(
+    resolveApp(`./dist/index.js`),
+    [
```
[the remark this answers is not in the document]
FredyC: I mostly wanted to fix the formatting, before it was like weird indentation all over the place. Do you really insist on such nitpick? :)

jaredpalmer:
Yep

src/index.ts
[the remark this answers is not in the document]
jaredpalmer: Yep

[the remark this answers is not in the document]
jaredpalmer: Yep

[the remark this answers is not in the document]
FredyC: Ok, it's changed. Please try to get it out soon, it's fairly important fix.

FredyC:
Can you release it this week, please?

jaredpalmer:
Yup. On it meow.

agilgur5:
@allcontributors please add @FredyC for code, docs, test, ideas

agilgur5:
@allcontributors please add @TrySound for review, questions, ideas

PR #145, opened 2019-06-16
Fix typo in readme

quentin-sommer:


agilgur5:
@allcontributors please add @quentin-sommer for docs

PR #158, opened 2019-06-24
Provide ability to preserve console output during watch

johnrjj:
Hey there!

Huge fan of `tsdx`, we currently use it to compile packages within our monorepo.

When running tsdx in watch mode, tsdx currently clears the console with no ability to retain the console output. This PR adds an opt-in ability to preserve the console output during watch. This PR behavior aligns with the existing TypeScript compiler option ( https://www.typescriptlang.org/docs/handbook/compiler-options.html, e.g. `tsc -w --preserveWatchOutput`).

As far as use cases, a common workflow inside a monorepo is to run all packages in watch mode using something like [`wsrun`](https://github.com/hfour/wsrun). Clearing the output makes it hard to debug or audit the logs if something goes wrong within the monorepo dependency chain on start up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,12 +278,13 @@ Usage
   $ tsdx watch [options]
 
 Options
-  -i, --entry    Entry module(s)
-  --target       Specify your target environment  (default web)
-  --name         Specify name exposed in UMD builds
-  --format       Specify module format(s)  (default cjs,esm)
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Rename this option to —verbose

agilgur5:
@allcontributors please add @johnrjj for code, docs

PR #159, opened 2019-06-24
Update createJestConfig.ts `testMatch` to allow for other folder structures

kylemh:
I like to pair testing files with my source code files so everything is collocated in a file explorer. This shouldn't be a breaking change and will simply allow me to create test files outside of the initially created `test` folder.


kylemh:
Anything wrong with this @sw-yx ?

kylemh:
@jaredpalmer https://jestjs.io/docs/en/configuration#testpathignorepatterns-array-string

It looks like any `node_modules` is ignored by default. I didn't see `testPathIgnorePatterns` defined in the config, but I could make it explicit in this config to prevent future issues.

nobitagit:
Just chiming in to say that I tested this on one of my projects and indeed it doesn't run node_modules tests, if that helps. 

As @kylemh says one is still free to keep tests under `test` but now has the option to co-locate, so no breaking changes.

jaredpalmer:
Coolz.

PR #160, opened 2019-06-25
Gracefully exit with code 1 when build failed

jooohn:
I found some problems with `build` command.

1. It does exit with code `0` even when the build failed. I think it should be `1`.
2. When it fails, the error won't be caught and that causes `UnhandledPromiseRejectionWarning`.

PR #162, opened 2019-06-28
Exit different than 0 when build is failing.

Pruxis:
I noticed that our CI pipeline could not catch build errors due to that they aren't exitting the process with an error code.


agilgur5:
Duplicate / superseded by #160

PR #163, opened 2019-07-02
Add template flag to create command

wrdls:
I want to run `tsdx create` from another program which means I can't fill in the template prompt.

I added an optional `--template` flag to the create command. 
The flag value is checked with the options available in the prompt. If invalid, the user is prompted to make a choice.


agilgur5:
@allcontributors please add @wrdls for code, idea

PR #164, opened 2019-07-04
Add doctype to react template html file

leonardodino:
Hey there :wave: Thanks for this awesome package!

---

Pretty minor diff, but [`react-live`](https://github.com/FormidableLabs/react-live) was having weird rendering issues without it.

Only discovered because I was porting an already functioning example, without any code changes.

I can also run prettier on this file if preferred.
`vscode` ends up auto-formatting it on save.

### DIFF
--- a/templates/react/example/index.html
+++ b/templates/react/example/index.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <html lang="en">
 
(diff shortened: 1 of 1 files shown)


agilgur5:
@allcontributors please add @leonardodino for code, bug

PR #173, opened 2019-07-31
Build UMD if name was specified

FredyC:
Just a quick idea. I think it makes sense if a name was explicitly specified to have UMD build by default as well.

The issue for a reference: https://github.com/JustFly1984/react-google-maps-api/pull/197

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -381,15 +381,19 @@ prog
   });
 
 async function normalizeOpts(opts: any) {
+  let format = opts.format.split(',')
+  if (opts.name && format.indexOf('umd') === -1) {
+    format.push('umd')
+  }
   return {
     ...opts,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
I feel like this is a magical side effect? Would it be better to throw an error saying that the name is irrelevant unless you specify --format=umd?

FredyC:
The worst thing that happens is that you get extra umd build which means slightly longer build time.

Throwing error feels like an annoyance to a user with no usefulness. He kinda has to double confirm that he really wants that UMD format :)

Besides, currently, the `esm,cjs` is build by default. In case you need `umd`, you have to list all 3 formats there for no good reason. So utilizing the `name` is kinda escape hatch to "over configure" :)

jaredpalmer:
Closing. I just think this is too magical for my taste.

Issue #169, opened 2019-07-28
use babel-plugin-transform-runtime for generator function support


xaviergonz:
### Current Behavior

Right now, when using generator functions, a global import of regenerator-runtime is required, which means that either the library consuming project has to import it itself, or that the library has to import it (which can cause problems due to different versions of the regenerator-runtime being possibly imported.

### Desired Behavior

The library should use its own version of the regenerator-runtime without polluting the global namespace.

### Suggested Solution

https://babeljs.io/docs/en/babel-plugin-transform-runtime

Using that plugin with regenerator set to true + including regenerator-runtime as dependency should ensure that the regenerator runtime is not using a global namespace and can be imported by the own library without problems.

### Who does this impact? Who is this for?

Libraries that make use of generator functions.

agilgur5:
Automatic polyfilling of generators with `babel-plugin-polyfill-regenerator` for `targets` that need a polyfill has been added #795 and will be released in v0.14.0 soon.

PR #183, opened 2019-08-22
V0.9

### LINKED ISSUE #110: extend rollup configuration
### Current Behavior
At the moment there's no way to add other configurations to rollup. In my case I would like to add support to css modules but this is not possible because it require a configuration in rollup as well

### Desired Behavior
Extend rollup config

jaredpalmer:
Closes #110 #169

You can now extend the rollup config via `tsdx.config.js`.

Now you can do this....
```js
// tsdx.config.js
const postcss = require('rollup-plugin-postcss');
const autoprefixer = require('autoprefixer');
const cssnano = require('cssnano');

module.exports = {
  rollup(config, options) {
    config.plugins.push(
      postcss({
        plugins: [
          autoprefixer(),
          cssnano({
            preset: 'default',
          }),
        ].filter(Boolean),
        // only write out CSS for the first bundle (avoids pointless extra files):
        inject: false,
        extract: options.writeMeta,
      })
    );
    return config;
  },
};

```


sw-yx:
got some time to check this out. so `extractErrors` changes from a string to boolean. i feel like i can argue pros and cons of that but you've probably thought this through. either way i'll update some parts of the docs that fell out of date w this pr

sw-yx:
getting "Sourcemap is likely to be incorrect: a plugin was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help" this is probably us...

Andarist:
`rollup-plugin-babel` should generate source maps by default, unless you turn it off - https://github.com/rollup/rollup-plugin-babel/blob/4d8c0c570e35200c3456372494d1029a4e2a2a90/src/index.js#L20

designorant:
Could we consider extending `rollup-plugin-typescript2` configuration in the same way babel one can be extended?

I've been looking into adding assents into my library via `@svgr/rollup` which led me to: https://github.com/ezolenko/rollup-plugin-typescript2/issues/105

The fix seems to be to pass the `objectHashIgnoreUnknownHack: true` option to `rpt` but that isn't quite possible in `tsdx` unless I'm missing something.

sw-yx:
open separate issue with suggestion of what api you’d like to see please

PR #185, opened 2019-08-26
update babel-plugin-transform-async-to-promises dependency

### LINKED ISSUE #172: Name of async recursive function is stripped away
### Current Behavior

Following code:

```ts
export function pipeP(...fns: Function[]) {
  return (value: any) =>
    new Promise((resolve, reject) => {
      ;(async function run([f, ...fns], value: any) {
        try {
          if (f === undefined) resolve(value)
          else run(fns, await f(value))
        } catch (e) {
          reject(e)
        }
      })(fns, value)
    })
}
```

 is being compiled to:

```js
function pipeP() {
  for (var _len = arguments.length, fns = new Array(_len), _key = 0; _key < _len; _key++) {
    fns[_key] = arguments[_key];
  }

  return function (value) {
    return new Promise(function (resolve, reject) {

// function run - its name is stripped
      (function (_ref, value) {
        var f = _ref[0],
            fns = _ref.slice(1);

        try {
          return _catch(function () {
            var _temp = function () {
              if (f === undefined) resolve(value);else return Promise.resolve(f(value)).then(function (_f) {
// calling run fails, because theres's no run defined
                run(fns, _f);
              });
            }();

            if (_temp && _temp.then) return _temp.then(function () {});
          }, function (e) {
            reject(e);
          });
        } catch (e) {
          Promise.reject(e);
        }
      })(fns, value);
    });
  };
}
```

Fails with `run is not defined` - see comments above for explanation.

### Expected behavior

Does not strip function's name

### Your environment

| TSDX           | 0.7.2
| TypeScript       | 3.5.3

honzabrecka:
Fixes #172

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "asyncro": "^3.0.0",
     "babel-plugin-annotate-pure-calls": "^0.4.0",
     "babel-plugin-dev-expression": "^0.2.1",
-    "babel-plugin-transform-async-to-promises": "^0.8.11",
+    "babel-plugin-transform-async-to-promises": "^0.8.14",
     "babel-plugin-transform-rename-import": "^2.3.0",
     "babel-traverse": "^6.26.0",
     "babylon": "^6.18.0",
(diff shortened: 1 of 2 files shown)


agilgur5:
@allcontributors please add @honzabrecka for code, bug

PR #195, opened 2019-09-04
Pass tests in cases where no tests exist

sadsa:


sw-yx:
would you apply this flag to the non react test commands as well?

sadsa:
My answer is Yes but I guess this question is more "Should the test script pass in the absence of tests?". Personally, I found it strange `tsdx test` threw an error because it couldn't find any valid tests. I would expect that it said "I couldn't find any tests to run, oh well, lets move on"

sw-yx:
ok. no strong feelings on this one

agilgur5:
@allcontributors please add @sadsa for code

PR #208, opened 2019-09-13
Support Async/Await in Rollup Plugins 

### LINKED ISSUE #200: Migrate Rollup TS Plugin to '@wessberg/rollup-plugin-ts'
### Current Behavior

TSDX uses `rollup-plugin-typescript2` to compile and check TypeScript files. Due to [certain anomalies](https://github.com/ezolenko/rollup-plugin-typescript2/issues/105) specific to this plugin, tsdx users are unable, _inter alia_, to use additional rollup plugins that contain async/await syntax (_e.g._, [`rollup-plugin-visualizer`](https://github.com/btd/rollup-plugin-visualizer)) without a workaround. _See_ https://github.com/ezolenko/rollup-plugin-typescript2/issues/105; _see also_ https://github.com/palmerhq/tsdx/pull/183#issuecomment-528312690. 

Unfortunately, the workaround, aptly titled, `objectHashIgnoreUnknownHack,` 'can potentially make cache stale,' due to the fact that `rollup-plugin-typescript2` uses a hash of the entire rollup configuration to facilitate cache invalidation. _See_ https://github.com/ezolenko/rollup-plugin-typescript2/issues/105#issuecomment-412986356. As a workaround for the workaround, users are encouraged to bypass the cache completely by setting `clean` to `true` in their rollup configurations. _See_ https://github.com/ezolenko/rollup-plugin-typescript2/issues/105#issuecomment-412955438. 

> The plugin uses rollup config as part of cache key. object-hash is used to generate a hash, but it can't hash certain elements at the moment. Setting this option to true will make object-hash ignore unknowns, at the cost of not invalidating the cache if ignored elements are changed. Only enable this if you need it (Error: Unknown object type "asyncfunction" for example) and make sure to run with clean: true once in a while and definitely before a release. (See #105)

### Desired Behavior

As a tsdx user, I should be able to extend tsdx's rollup configuration to include additional plugins that use async/await syntax without a workaround and, especially, without having to worry whether my cache is stale. 

### Suggested Solution

Replace `rollup-plugin-typescript2` with [`rollup-plugin-ts`](https://github.com/wessberg/rollup-plugin-ts). 

While #183 added support for extending tsdx's rollup configuration, making it easy to add additional rollup-plugins, there is no easy way to swap out tsdx's default plugin choices. _See, e.g._, _infra_, Alternative Solutions. Thus, tsdx users are, more or less, blocked from using any rollup plugin that contains `async/await` code. As the `async/await` syntax continues to gain popularity, it is likely that this issue will affect more and more tsdx users. 

With #183, tsdx users can hack tsdx's default plugin choices (example below) to either (1) to modify the `rollup-plugin-typescript2` configuration as described above, or (2) swap out the plugin for `rollup-plugin-ts`; however, neither approach is ideal. 

```
const ts = require("@wessberg/rollup-plugin-ts");

module.exports = {
  rollup(config, options) {
    config.plugins = config.plugins.map(plugin => {
      if (plugin && plugin.name === "rpt2") {
        return ts({
          tsconfig: tsconfig => ({
            ...tsconfig,
            target: "esnext",
            sourceMap: true,
            declaration: true,
            jsx: "react"
          }),
          transpiler: "babel"
        });
      }
      return plugin;
    });

   ...

    return config;
  }
};

``` 

For diagnostic purposes, a working example using the above swap-out hack can be found here:  https://github.com/medelman17/tsdx-monorepo-template 

### Additional context

The `rollup-plugin-typescript2` plugin appears to have several additional bugs that may cause tsdx users pain. For example, to enable code splitting, users must also employ an additional workaround, setting `rollupCommonJSResolveHack` to `true`. But see: 

> On windows typescript resolver favors POSIX path, while commonjs plugin (and maybe others?) uses native path as module id. This can result in namedExports being ignored if rollup happened to use typescript's resolution. Set to true to pass resolved module path through resolve() to match up with rollup-plugin-commonjs. rollup-plugin-commonjs fixed this in 10.1.0, so projects using this option who update to new version will be broken again. This also works around the similar bug affecting code splitting (see rollup/issues/3094).

- https://github.com/ezolenko/rollup-plugin-typescript2

medelman17:
This is a WIP re #200. Sadly, I cannot, for the life of me, get the last two tests to pass. A screenshot of the issue is included below. It looks related to our babel/typescript relationship-- which is complex. I'm happy to continue digging in, but figured I would surface in case anyone could point me in the right direction. 

FWIW, it looks like the new plugin has a fairly robust integration with babel. I'm not sure re purpose of the custom tsdx babel plugin, but could we accomplish the same with a `.babelrc` or `babel.config.js` file? If so, the new plugin looks capable of managing the babel stage. See [Meaw](https://github.com/wessberg/rollup-plugin-ts#combining-typescript-with-babel). 

Screen Shot 2019-09-12 at 8 23 31 PM
 
closes #200


src/createRollupConfig.ts
```diff
@@ -125,22 +126,16 @@ export function createRollupConfig(opts: TsdxOptions) {
...
-            jsx: 'react',
-          },
-        },
-        tsconfigOverride: {
-          compilerOptions: {
-            target: 'esnext',
-          },
-        },
+      ts({
+        tsconfig: tsconfig => ({
```
tunnckoCore: Best bet would be to check if `opts.tsconfig` is given, if not pass path to the TSDX default tsconfig.json. Something like mine

```json
{
  "$schema": "http://json.schemastore.org/tsconfig",
  "compilerOptions": {
    "skipLibCheck": true,
    "alwaysStrict": true,
    "newLine": "LF",
    "declaration": true,
    "declarationMap": true,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": false,
    "allowSyntheticDefaultImports": true,
    "forceConsistentCasingInFileNames": true,
    "importHelpers": true,
    "incremental": false,
    "jsx": "react",
    "lib": ["dom", "esnext"],
    "module": "esnext",
    "moduleResolution": "node",
    "noEmit": true,
    "strict": true,
    "sourceMap": true,
    "stripInternal": true,
    "noUnusedLocals": true,
    "noImplicitReturns": true,
    "resolveJsonModule": true,
    "noUnusedParameters": true,
    "target": "esnext"
  }
}
```

Some of them will be automatically ignored anyway.

Also would be good to provide some options for "export tsconfig" and "export babel config" to the root of the project that is using TSDX.

src/createRollupConfig.ts
```diff
@@ -125,22 +126,16 @@ export function createRollupConfig(opts: TsdxOptions) {
...
+          ...tsconfig,
+          target: ScriptTarget.ESNext,
+          sourceMap: true,
+          declaration: true,
+          jsx: JsxEmit.React,
+        }),
+        // transpiler: 'babel',      // nothing seems to make last two tests pass.
```
tunnckoCore: Uncomment it, and also pass the desired default babel config as `babelConfig`.

But I'm not sure if you pass babelConfig option does it will support project's babel configurations. This `ts` plugin is doing A TON of magic behind the scenes. It generally will work without config, and even if the project's babel.config.js has some presets it will override some of their options. 

A lot of magic is happening which I don't like. Just look how big the codebase is.

wessberg: Hey there, I'm the author of the package. I'm glad to assist you with any questions you may have during migration. Yes, it's true that some Babel options are forced to ensure maximum interoperability with Rollup. I've documented all of these forced options [here](https://github.com/wessberg/rollup-plugin-ts/blob/master/README.md#ignoredoverridden-babel-options) and [here](https://github.com/wessberg/rollup-plugin-ts/blob/master/README.md#combining-typescript-with-babel).

If there are any forced options that you find questionable, please let me know so I can take a look at it.

medelman17:
Sorry, y'all, for the delayed update, but, with one minor caveat, we should now be good to go, here. The remaining 'hitch' relates to the name of the emitted type declaration files. 

In `tsdx`-generated `package.json` files we include the following configuration relevant here: 
```
{
  ... 
  "main": "dist/index.js", 
  "typings": "dist/index.d.ts"
  ...
}
```

However, the plugin wants to emit declaration files on a (potential) chunk/entry-by-chunk/entry basis using actual filenames, bundle-type-suffixes, etc.. As a consequence, no `index.d.ts` file is ever generated. This appears to be the reason the tests above were failing--i.e., it wasn't that types/declaration files were not being included; rather, our tests were just looking for the wrong file. Maybe @wessberg can shed some additional light here, but I couldn't figure out how to configure `rollup-plugin-ts` to do otherwise. 

Thus, to retain functional parity with respect to typings, I changed the default `package.json` configuration to point to the correct/equivalent file: `dist/${safeName}.esm.d.ts` and amended the tests accordingly. This should work, but it is not as pretty. Alternatively, we could add another build step to copy/rename the declaration file. Otherwise, I think this is baked and ready to merge.

wessberg:
@medelman17, that's correct, `rollup-plugin-ts` will generate declarations for each output chunk, and the declarations will be identically named to that of the output chunks.

So if Rollup writes an output chunk to `dist/index.esm.js`, the declaration will be written to `dist/index.esm.d.ts`.

jaredpalmer:
Nice

timReynolds:
Happened to hit this yesterday and it's annoying to work around without this change, any idea when this could be merged?

tunnckoCore:
Not especially related to the PR, but just my 2cents :)

> @wessberg: So if Rollup writes an output chunk to

And that's kind of strange, but understanding it. If you output 2 formats - cjs and esm (as in my case, in different directories - `dist/cjs/index.js` and `dist/esm/index.js`), it generates basically the same type files (`dist/cjs/index.d.ts` and `dist/esm/index.d.ts`, which is unnecessary. In my case I did [workaround it](https://github.com/tunnckoCore/configs/blob/a90e8183a4a7a678ec0e42dec253277c789a7001/packages/rollup-config/create-config.js#L219-L229) to just generate types at `dist/types/index.d.ts`. But yea, I understand why this is/should happen.

medelman17:
Yes. This is precisely the hitch above. What to do about that?

Since TSDX does both cjs and esm out of the box,  by default we generate two identical .d.ts files (neither of which is named index.d.ts 🙄), but for the file names, using this plugin.

In other words: 😬

I suppose I lean towards, ‘letting it happen,’ but I don’t have enough skills/experience to feel confident about that. That said, even writing an extra build step just to clean up the naming issue purely for aesthetics feels hack-y. So, if, as a practical matter, having the extra types included in the package has a negligible impact, I would accept that as a trade off. But I don’t know RT impact.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tunnckoCore:
Oh sure, I'm sleepy, haha. I'll check it: `do not merge` ;p

> That said, even writing an extra build step just to clean up the naming issue purely for aesthetics feels hack-y

Besides that, feels too much :D

wessberg:
@tunnckocore, @medelman17, if you use `tsc` directly, the declarations will be emitted alongside the emitted JavaScript and SourceMap files, having the same basename, and this is the same default behavior of `rollup-plugin-ts`. It goes for both `tsc` and `rollup-plugin-ts` that you can provide a `declarationDir` option inside `tsconfig.json` to ensure that declarations are writing to a specific folder. However, there's no way of forcing a specific basename.

If this is an important feature requirement for you, I can look into providing a hook to the config that is invoked with the path to a declaration file immediately before it is emitted, which then allows you to rewrite it to something else. That would allow you to always write to the same path and with the same name, no matter the amount of Rollup outputs.

Would you be interested in this?

@tunnckocore The generation of declarations could potentialy have some form of shared caching that may enhance performance, given that the input will be identical.

But I don't see how we could avoid generating declarations for each Rollup output, given that the output configs in Rollup can be very different. For example, an `intro` and `outro` can be added to an output, and the entry- and chunk names may differ.

Let me give an example:

In a multi-chunk setup with two outputs: ESM and CJS, where the chunk name patterns differ, Rollup may convert this:

```typescript
import {Foo} from "./foo";
console.log(Foo);
export type Bar = typeof Foo;
```

To this for ESM

```javascript
// index.esm.js
import {Foo} from "./chunk-0027dabcd....js";
console.log(Foo);
```

And this for CJS:

```javascript
// index.cjs.js
const foo = require("./other-pattern-123456...");
console.log(foo.Foo);
```

Now, if we wanted to generate - and bundle - declarations for these two index files, it becomes clear that they are different.

The `index.esm.d.ts` file could look like this:

```typescript
// index.esm.d.ts
import {Foo} from "./chunk-0027dabcd...";
export declare type Bar = typeof Foo;
```
And the `index.cjs.d.ts` file could look like this:

```typescript
import {Foo} from "./other-pattern-123456...";
export declare type Bar = typeof Foo;
```
So under these circumstances, they would be different.

For by far the most library builds (though _definitely_ not all of them), then yes, the declarations will be identical and only a single chunk will be emitted. I guess we could look into a heuristic saying something like "If all Rollup options like `input`, `output`, `entryFileNames`, `chunkFileNames` are identical across all output configs", then only generate declarations once".

tunnckoCore:
> It goes for both tsc and rollup-plugin-ts that you can provide a declarationDir option inside tsconfig.json to ensure that declarations are writing to a specific folder. However, there's no way of forcing a specific basename.

Yea yea, that's exactly what I did.

> Would you be interested in this?

Don't it's needed at all for this case here - I'll try some things based on this PR. But might be a cool feature for some other cases.

> But I don't see how we could avoid generating declarations for each Rollup output

Yea, I see it now more clearly.

It's possible, and the trick is quite easy. But TSDX does not support/emit/do anything chunks related anyway. So it's not the case here.

Basically the thing is that: for the one output format, use the plugin without providing tsconfig option, which means it will pick it up automatically, which must include (if you want declaration files at all) `declaration: true` and `declarationDir` typescript compiler options; and for the second output, pass a custom tsconfig path (which extends the base one) to the ts plugin, which specifically is adding `declaration: false` - that thing I called `tsconfig.notypes.json`.

tunnckoCore:
So, basically I'll try to change this

https://github.com/medelman17/tsdx/blob/7f796bb66f906ad4182bef6f42b20656f52875d6/src/createRollupConfig.ts#L129-L138

to something like

```js
      opts.format === 'esm'
      ? ts({
        tsconfig: tsconfig => ({
          ...tsconfig,
          target: ScriptTarget.ESNext,
          sourceMap: true,
          declaration: true,
          declarationDir: `${paths.appDist}/types`,
          jsx: JsxEmit.React,
        }),
        transpiler: 'babel',
      })
      : ts({
        tsconfig: tsconfig => ({
          ...tsconfig,
          target: ScriptTarget.ESNext,
          sourceMap: true,
          declaration: false,
          jsx: JsxEmit.React,
        }),
        transpiler: 'babel',
      }),
```

tunnckoCore:
So actually, @wessberg, this hook you are talking about will be in help here.

tunnckoCore:
Anyway, forget about all the stuff I said. The whole thing boils down to that we use "safe package name" for output filename, instead of `index` which is definitely better. And second, the `safePackageName` function isn't that good. It completely strips the scope - in babel hypothetically case dist will be `core.<etc>.js`, in my scoped `@hela/core` package it will be the same, which isn't okay. At least we can include the scope and replace `/` with `-`. But overall, it makes things unpredictable and error/confusion-prone. 

If user doesn't use the `tsdx create` he'll end up wondering what his typings file is. If we have the following output structure it will be a lot easier.

```
dist/index.js - conditional require, the `main` field points here
dist/cjs/index.js - normal
dist/cjs/index.min.js - production
dist/esm/index.js - normal, the `module` field points here
dist/esm/index.min.js - production
``` 

I don't understand why we need the package name, "production", "development" and the format in the filename. It's hard on eyes. Not to mention that if it's "min" it's obviously for production (whatever that may mean in the different cases), right?

To be able to have single `index.d.ts` file we'll need to do the above-mentioned trick, checking what format is and if it's not "min". Otherwise, we'll end up having `dist/cjs/index.d.ts` and `dist/esm/index.d.ts` which will be the same file. And I'm assuming that we won't be able such `.d.ts` file (e.g. `dist/index.d.ts` which loads the cjs or esm d.ts file) as the `dist/index.js`, not to mention it does not make sense to have it, haha.

tunnckoCore:
Seems good, right?

2019-09-30-193623_1280x1024_scrot
2019-09-30-193720_1280x1024_scrot

So, the users can for sure bet on this package.json fields

```json
{
  "main": "dist/index.js",
  "module": "dist/esm/index.js",
  "typings": "dist/types/index.d.ts"
}
```

wessberg:
> So actually, @wessberg, this hook you are talking about will be in help here.

Alright, I've implemented such a hook and released it under v1.1.65.
I've documented it [here](https://github.com/wessberg/rollup-plugin-ts/tree/master#the-outputpath-hook)

tunnckoCore:
@wessberg cool! I'll try it.

---

offtopic: anytime I'm visiting this PR I'm wondering why it's called that way, @medelman17? :thinking: :rofl: I don't get it. What has it to do with async/await? :D

tunnckoCore:
Yup, it works great. But still, don't think it's for our case here :)

src/index.ts
```diff
@@ -243,7 +243,7 @@ prog
         version: '0.1.0',
         main: 'dist/index.js',
         module: `dist/${safeName}.esm.js`,
-        typings: 'dist/index.d.ts',
+        typings: `dist/${safeName}.esm.d.ts`,
```
tunnckoCore: Revert it back to old.

src/createRollupConfig.ts
```diff
@@ -125,22 +126,15 @@ export function createRollupConfig(opts: TsdxOptions) {
...
-            declaration: true,
-            jsx: 'react',
-          },
-        },
-        tsconfigOverride: {
-          compilerOptions: {
-            target: 'esnext',
-          },
-        },
+      ts({
```
tunnckoCore: add

```js
hook: {
  outputPath: (fp, kind) => {
    if (/declaration/.test(kind) && opts.format === 'esm') {
      return path.join(path.dirname(fp), 'index.d.ts');
    }
  }
}
```

test/tests/tsdx-build.test.js
```diff
@@ -29,7 +29,7 @@ describe('tsdx build', () => {
     ).toBeTruthy();
     expect(shell.test('-f', 'dist/build-default.esm.js')).toBeTruthy();
 
-    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+    expect(shell.test('-f', 'dist/build-default.esm.d.ts')).toBeTruthy();
```
tunnckoCore: back to previous

src/createRollupConfig.ts
```diff
@@ -125,22 +126,15 @@ export function createRollupConfig(opts: TsdxOptions) {
...
+        tsconfig: tsconfig => ({
+          ...tsconfig,
+          target: ScriptTarget.ESNext,
+          sourceMap: true,
+          declaration: true,
```
tunnckoCore: replace `true` with `opts.format === 'esm'`

tunnckoCore:
After above requested changes:

```
~/github/medelman17/tsdx/test/fixtures/build-default chore/swap-rollup-ts-plugins* 13s
❯ ls dist    
build-default.cjs.development.js         build-default.esm.js                  build-default.umd.production.min.js
build-default.cjs.development.js.map     build-default.esm.js.map              build-default.umd.production.min.js.map
build-default.cjs.production.min.js      build-default.umd.development.js      index.d.ts
build-default.cjs.production.min.js.map  build-default.umd.development.js.map  index.js
```

So, yea. We can go with this PR or #219.

jaredpalmer:
Which PR are we going with?

tunnckoCore:
Your call. @medelman17 needs to make changes here.

designorant:
> Which PR are we going with?

I'd go with this one purely for the benefit of having a full history here, even though @tunnckoCore has taken everything to a new level, really. Nevertheless, it looks like @medelman17 has been busy lately.

> What it has to do with async/await?

@tunnckoCore It's explained in the first paragraph of #200 it addresses.

medelman17:
I can make the needed changes to mine early  tomorrow. That said, and while it would be great to have my contribution forever immortalized on my GH profile, I’m pretty sure I’m the only one who will notice. :-)

So, let’s just go with whatever makes the most sense for the library and not make it emotional. If that’s the other PR, no problem. To the extent we can merge the two together and that would make sense, awesome. But ... whichever way, let’s get ‘er done. :-)


Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

medelman17:
Alrighty. Feedback/changes from @tunnckoCore have been incorporated. :-)

tunnckoCore:
Cool. But still, I think that we should somehow drop the dependency `rollup-plugin-babel` which mean we should move the tsdxPluginBabel (or how it was named) things as options passed to `ts({ babelConfig })`.

jaredpalmer:
Going with this as it is not a breaking change. We can debate the output/safe-variable name elsewhere.

jaredpalmer:
Thank you @medelman17 @tunnckoCore. You are both awesome.

bitmoji

agilgur5:
@allcontributors please add @medelman17 for code, ideas

agilgur5:
@allcontributors please add @tunnckoCore for reviews, code, ideas

agilgur5:
@allcontributors please add @wessberg for questions

agilgur5:
Thanks everyone for helping with this even if it eventually got rolled back.

To close this out for any readers that are unsure -- can see https://github.com/jaredpalmer/tsdx/issues/294#issuecomment-584456814 and #506 for how I ended up fixing/getting this fixed upstream so that async Rollup plugins can be used.

PR #219, opened 2019-09-30
Use `@wessberg/rollup-plugin-ts` + change output naming structure

tunnckoCore:
Addresses #208 and #200.

- Switch the output naming structure/convention. We are kind of forced to do that to make a better and more predictable experience, and integrating this plugin.
- Update `tsdx create` according to new output structure; Plus, correct "package generation". Currently, if user type `@hela/core` in the prompt it will create `core` folder in the `cwd` and also will add to the generated package.json file a name field to be `core` instead `@hela/core`, which definitely isn't correct behavior. 
- Remove the "utils safePackageName" test.
- Remove the safePackageName function, at all.
- ESLint autofix is enabled on my VSCode, so ESLint removed the `use strict` on the `tsdx-build.test.js`, obviously according to the given eslint config and prettier.
- Immediately before I started anything with just `pnpm i` on fresh branch, the pnpm's lockfile changed.

The only left is to update the `babelPluginTsdx` file, meaning to drop the `rollup-plugin-babel` dependency and just generate babel config and pass it as `babelConfig` option to the ts plugin. But keep in mind the ts plugin also automatically includes some plugins and does merging some default options.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -9,7 +9,7 @@ import replace from 'rollup-plugin-replace';
 import resolve from 'rollup-plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import { ScriptTarget, JsxEmit } from 'typescript';
-import typescript from '@wessberg/rollup-plugin-ts';
+import ts from '@wessberg/rollup-plugin-ts';
 import { extractErrors } from './errors/extractErrors';
 import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
@@ -125,8 +125,8 @@ export function createRollupConfig(opts: TsdxOptions) {
         },
       },
       opts.format === 'esm' && !shouldMinify
-        ? typescript({
-            transpiler: 'typescript',
+        ? ts({
+            transpiler: 'babel',
             tsconfig: tsconfig => ({
               ...tsconfig,
               target: ScriptTarget.ESNext,
@@ -136,8 +136,8 @@ export function createRollupConfig(opts: TsdxOptions) {
               jsx: JsxEmit.React,
             }),
           })
-        : typescript({
-            transpiler: 'typescript',
+        : ts({
+            transpiler: 'babel',
             tsconfig: tsconfig => ({
               ...tsconfig,
               target: ScriptTarget.ESNext,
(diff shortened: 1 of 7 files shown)


src/createRollupConfig.ts
```diff
@@ -30,9 +31,7 @@ export function createRollupConfig(opts: TsdxOptions) {
     opts.minify !== undefined ? opts.minify : opts.env === 'production';
 
   const outputName = [
-    `${paths.appDist}/${safePackageName(opts.name)}`,
-    opts.format,
-    opts.env,
+    `${paths.appDist}/${opts.format}/index`,
```
jaredpalmer: If we were nice we would automagically update folks package.json for them

test/tests/tsdx-build.test.js
```diff
@@ -21,15 +20,11 @@ describe('tsdx build', () => {
...
-      shell.test('-f', 'dist/build-default.cjs.development.js')
-    ).toBeTruthy();
-    expect(
-      shell.test('-f', 'dist/build-default.cjs.production.min.js')
-    ).toBeTruthy();
-    expect(shell.test('-f', 'dist/build-default.esm.js')).toBeTruthy();
-
-    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+    expect(shell.test('-f', 'dist/cjs/index.js')).toBeTruthy();
+    expect(shell.test('-f', 'dist/cjs/index.min.js')).toBeTruthy();
```
jaredpalmer: I am confused. Is .min the production version?

jaredpalmer:
Why are we changing the output format?

src/createRollupConfig.ts
tunnckoCore: Meaning? Codemod thingy? It's could be the next possible step after this whole change.

test/tests/tsdx-build.test.js
tunnckoCore: Sure. Why otherwise one will want minifying? and especially, in cjs/node target.

tunnckoCore:
> Why are we changing the output format?

For "reasons", the other PR has comments around that :) Or at least because such design decisions, problems and "features" appear in some other places/tools (referring the new "hook" option in the `ts` plugin). All this is just one simple example that may seem negligible, but this is happening for years in the ecosystem.

PR #228, opened 2019-10-10
Add system to list of supported formats

mfolnovic:
We need support for SystemJS format because we would like to load generated bundle through SystemJS.

Hopefully, I haven't missed changing something. :blush: 

I've tested just basic scenario and it seems okay. Not sure how to test it on some complex scenario.


jaredpalmer:
Seems fine.

agilgur5:
@allcontributors please add @mfolnovic for code, docs

Issue #150, opened 2019-06-19
Cannot set custom jest.config.{js,ts} even with --config flag in test CLI command


SeedyROM:
### Current Behavior

The `test` command's `--config, -c` flags throw a nondescript error when passed any value.
```
$ yarn run tsdx test --config=./jest.config.js --env=jsdom
...Usage...
argv.config.match is not a function
```

This also doesn't load the `test` command in watch mode as the documentation specified.

### Expected behavior

Either having a `jest.config.{js,ts}` in the project root or specifying one via the config flag in your `tsdx test` should actually be loaded on setup.

### Suggested solution(s)

**Quickfix:**
Just move all your configuration into `package.json` under the "jest" key.

**Long term fix:** 
I can't seem to find the bug first thing in [index.ts](https://github.com/palmerhq/tsdx/blob/6a46b801f86716e0a9047147376e9aafb9fbe0f6/src/index.ts), but this is definitely where things are going wrong.

I will update this issue as I study it more!

### Additional context

The example bug repo is available [here](https://github.com/SeedyROM/tsdx-jest-config-bug).

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.7.1 |
| TypeScript       | 3.5.2 |
| Browser          | NA |
| yarn         | 1.15.2 |
| Operating System | Ubuntu 16.04 |

Aidurber:
@SeedyROM @jaredpalmer I'm taking a quick look at this and I'm unsure what would be the correct fix. 

The easiest is to check if the args already contain a config ([here](https://github.com/jaredpalmer/tsdx/blob/master/src/index.ts#L463)), if it does then don't push the tsdx config. 

Or do we merge in the contents of the supplied config with the tsdx config? It would probably need to be a deep merge if you need to add in custom transforms into your jest config such as CSS transforms.

SeedyROM:
I think a deep merge would be appropriate, but I'm not an expert on this library. So take my opinion with a grain of salt.

agilgur5:
Ran into this myself during some testing (and, notably, TSDX itself uses `jest --config`) and did some further investigation.

This same error, `argv.config.match is not a function` also occurs with CRA (e.g. https://github.com/wmonk/create-react-app-typescript/issues/396) as it uses the same code for config:
[...]

PR #229, opened 2019-10-11
fix(testing): add support for jest.config.js

### LINKED ISSUE #100: tsdx test ignores snapshotSerializers in jest.config.js
### Current Behavior

If I run `tsdx test --env=jsdom --coverage` Jest will not use the snapshot serializers defined in the `jest.config.js` and fail all my snapshots.

### Expected behavior

When I run `tsdx test --env=jsdom --coverage` it should use the snapshot serializers defined in the `jest.config.js`.

### Suggested solution(s)

Pick up the config in the `jest.config.js` or be able to define a snapshot serializers via the cli.

### Additional context

When I run `jest` all my snapshots pass.
I am using `jest-emotion` for the snapshot serializers.

### Your environment

| TSDX             |   0.5.9    |
| TypeScript       |   3.4.5    |
| Browser          |   Firefox  |
| npm/Yarn         |   npm      |
| Operating System | OSX 10.14.4|

Aidurber:
Currently, `jest.config.js` is ignored. My fix is to try and check if there is a `jest.config.js`, if so then we just merge it into the jestConfig. 

This fixes #100 . 

Issue #150 is in a similar vein to this but that needs some more work.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -469,10 +469,7 @@ prog
     };
     try {
       // Allow overriding with jest.config
-      const jestConfigContents = require(path.resolve(
-        paths.appRoot,
-        'jest.config.js'
-      ));
+      const jestConfigContents = require(paths.jestConfig);
       jestConfig = { ...jestConfig, ...jestConfigContents };
     } catch {}
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -460,15 +460,26 @@ prog
     });
 
     const argv = process.argv.slice(2);
+    let jestConfig = {
+      ...createJestConfig(
+        relativePath => path.resolve(__dirname, '..', relativePath),
+        paths.appRoot
+      ),
+      ...appPackageJson.jest,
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -460,15 +460,26 @@ prog
...
+        relativePath => path.resolve(__dirname, '..', relativePath),
+        paths.appRoot
+      ),
+      ...appPackageJson.jest,
+    };
+    try {
+      // Allow overriding with jest.config
+      const jestConfigContents = require(path.resolve(
+        paths.appRoot,
+        'jest.config.js'
```
jaredpalmer: We should add `jestConfig: resolveApp('./jest.config.js)` to `paths` and use it here.

Aidurber: Yep, makes sense 👍

agilgur5:
@allcontributors please add @Aidurber for code

PR #244, opened 2019-10-15
feat: MIT License by default

### LINKED ISSUE #241: License generation
### Current Behavior

package.json has no license on new project init

image

### Desired Behavior

Some way to have tsdx assist in adding a license.

### Suggested Solution

Add ability to select a license to `init`, or, to keep `init` as frictionless as possible, add a second `tsdx` command `tsdx license` which adds a `license` field to `package.json` and inserts the license that was chosen from the CLI.

Anyone making a library to publish with `tsdx` will likely want to license their project.

You could create the license on github when you create a new repository and then add the remote locally and do a `git pull origin master --allow-unrelated-histories`, but that is a somewhat complex workflow that messed up the projects HEAD, and creates an ugly initial history for your project.

-or-

Docs could recommend something like [lice](https://github.com/superkhau/lice) to assist with this

audiolion:
Adds an additional prompt to the CLI, somewhat unavoidable, we need the package author name to place in the LICENSE file. That said, we can double up and use the author name provided to place inside the package.json's author field as well.

image

Closes #241 

If you have other ideas about how to approach this problem @sw-yx @jaredpalmer let me know.

Some other ideas I had:
- Try and parse the user's git config for email / name, still need the CLI though because what if their git config is wrong, doesn't exist, etc. This could be useful to provide a sensible default for the name to make the create flow easier
- `tsdx config` cli option that creates and stores a config object somewhere on the file system and uses those fields as defaults. It just seems like this is flaky though, unless it was part of the users dotfiles where they carried them between systems.


jaredpalmer:
I thought about this last night. I really want to limit the choice presented to the user. I really do not want people to have to think. 

Some ideas:
- Can we pull out the author info from npm config or git config?
- We could create a `~/.tsdx` file in the user's home directory with this information, so subsequent runs don't need to pull it out and we can skip the question (even with `npx`)
- If we had our own `tsdx publish` we could get this info on the first run. This command would just be a thin wrapper around `np`. When it's run for the first time, it would prompt the user for whatever is missing in `package.json`. It could also maybe even check at the `scripts.build` against the presence of `main`, `module`, `umd` fields too.

orta:
We got pretty far with a mix of keychain and git [for CocoaPods](https://github.com/CocoaPods/pod-template/blob/da55e874c536cfc844b5cffcd1f2740458789475/setup/TemplateConfigurator.rb#L196-L204)

audiolion:
> Can we pull out the author info from npm config or git config?

I could update this PR to follow what @orta suggested and if I find anything to just use that, otherwise we prompt them for the information. I get the philosophy you are going for, and the user can always change the name in the license/package.json afterwards if they want something else.

audiolion:
Ok I updated the PR to parse the users global git config and search for their name from that, if not found it prompts. I would like to add npm config as well but I didn't see any existing tools for parsing npm config. I also could add the stuff cocoapods was doing, but I would need to be running  `shelljs` exec commands for that, I am not sure if I should be going down that road or not?

jaredpalmer:
I am cool with adding shelljs.

jaredpalmer:
go for it.

audiolion:
Ok changed out to using shelljs. The unfortunate consequence of this is that the shelljs command outputs to stdout.. not sure how to stop that from happening, I suppose we could really quickly rewrite stdout afterwards to clear it?

image

audiolion:
Ok switched to shelljs and dropped parse-git-config, as it was just another point of failure.

jaredpalmer:
@orta does this look good to you?

orta:
Could be worth setting the username in the npm config after asking for the user (so it only happens once) but other than that, looks good to me?

jaredpalmer:
@audiolion I agree. `npm-config set` or something if IIRC

audiolion:
groovy, done.
 
image

src/index.ts
```diff
@@ -464,6 +496,39 @@ if (process.env.NODE_ENV === 'production') {
...
+  author = shell
+    .exec('npm config get init-author-name', { silent: true })
+    .stdout.trim();
+  if (author) return author;
+
+  author = shell
+    .exec('git config --global user.name', { silent: true })
+    .stdout.trim();
+  if (author) {
+    setAuthorName(author);
```
audiolion: if they did not have an author with npm but do have a user name in git, we set their npm author name to it.

agilgur5:
Hmmm... the set / get author name capabilities added here seem appropriate to split out into a separate package that other libraries could use too

PR #245, opened 2019-10-15
Upgrade rollup-plugin-ts

jaredpalmer:


jaredpalmer:
@sw-yx should we just drop support for node v8?

sw-yx:
ha damn you move fast! i'm up to speed.

PR #247, opened 2019-10-15
Fix greenkeeper going cray

jaredpalmer:
Greenkeeper got a little too zealous. I rolled back some of the changes.


jaredpalmer:
So rollup-plugin-ts is using some jS that isn't supported in node 8

jaredpalmer:
@wessberg can you change `catch {}` to `catch (e) {}` and then update your engines to support node 8?

wessberg:
Yes, thanks for making me aware of this. v1.1.67 has been published. It includes catch-bindings, so it shouldn't break Node 8

PR #248, opened 2019-10-15
Drop node 8

jaredpalmer:
Remove Node 8 from the test matrix because `rollup-plugin-ts` doesn't support it. I tried patching it but there are downstream dependencies that also do not support Node 8 either. :shrug:


jaredpalmer:
@wessberg is updating his packages to support node 8. closing

agilgur5:
Newer version was in #426 . Related to #424

PR #250, opened 2019-10-15
Add back support for node 8

jaredpalmer:
Thank you @wessberg for the help


jaredpalmer:
@wessberg need you bump version of browserlist generator in rollup-plugin-ts and then publish .69

wessberg:
I've already bumped the dependency and published a new version, so we should be good 😊

jaredpalmer:
@wessberg Still failing: https://github.com/jaredpalmer/tsdx/runs/260960301#step:5:7

wessberg:
Ah, didn't change the engine property from >=9.00 to >= 8 inside of browserslist-generator before publish. I'll get a new version published later today and will update here when it's done.

wessberg:
Fixed in v1.1.69!

wessberg:
I see that you're having the issue with another one of my packages on the CI pipeline. Let me fix that one too.

jaredpalmer:
@wessberg https://github.com/jaredpalmer/tsdx/pull/250/checks?check_run_id=261186208#step:5:7

wessberg:
[Fixed](https://github.com/wessberg/rollup-plugin-ts/releases/tag/v1.1.70)

PR #267, opened 2019-10-17
Basic alt forking

tabuckner:
## Value
After reviewing the source code, I noticed that this is not entirely necessary, as the Jest config is set up to already look for all `spec/test` files in the entire project.

It's entirely possible for a user to make use of the current 'Basic' template and then just re-arrange files as they see fit. 

But I'm lazy, and so I've forked it to make me happy. I figured I'd submit a PR anyways, just in the event you guys see value in adding this to the core product.

I've also got a separate branch that just adds a new choice, `basic-alt` (maybe `basic-flat` is a better name?), and its respective template _if_ that better suits the project, _and_ you guys see value in the addition.

## What I Did
* Added a new template that is (_potentially_) more familiar for angular devs.
* Added some logic forking in the prompts for `tsdx create`

## Examples
image
image
image


sw-yx:
thanks for the PR! this is probably not a feature we want to maintain. you’re welcome to keep your fork up to date.

PR #209, opened 2019-09-15
Use `['src', 'test']` as default input files for lint

skvale:
Fixes #196

PR #269, opened 2019-10-18
[feat] Improve defaults for lint

lookfirst:
Specify directories to prevent 'No input files specified, defaulting to src test' warning
Specify all 3 common directories as default for react template.
Add `lint-fix` script.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -289,7 +289,7 @@ prog
           start: 'tsdx watch',
           build: 'tsdx build',
           test: template === 'react' ? 'tsdx test --env=jsdom' : 'tsdx test',
-          lint: 'tsdx lint',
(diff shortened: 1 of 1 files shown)


jaredpalmer:
@sw-yx why wouldn't we default to --write with lint?

lookfirst:
> @sw-yx why wouldn't we default to --write with lint?

works for me. i can adjust the pr if you want.

sw-yx:
@jaredpalmer @lookfirst idk, i'm fine with it :)

agilgur5:
Something similar was done in #209 to have `tsdx lint` internally default to `['src', 'test']` when no patterns are specified (meaning it supersedes adding these to the templates).

sw-yx:
awesome

PR #272, opened 2019-10-18
fix: don't remove dist folder, wipe content instead

### LINKED ISSUE #260: removing dist folder causes issues with parallel lerna processes
I don't know if this is the right place for this, neither if this is a bug.

### Additional context

I am using tsdx in a lerna monorepo. If I start all process at once via `lerna run --parallel start`, my `create-react-app` start process fails with `Module not found: Can't resolve <path/to/module>`.

My guess is the cleanup of the dist folder on watch mode collides with the parallel startup.

If this doesn't belong here, feel free to close it.

### Your environment

| TSDX           | 0.10.0
| TypeScript       | 3.6.3
| Browser          | Chrome Version 77.0.3865.120 (Offizieller Build) (64-Bit)
| npm/Yarn         | npm 5.8.0
| Operating System | Ubuntu 19.04 VM

audiolion:
Potentially fixes #260

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -474,7 +474,7 @@ function ensureDistFolder() {
 
 function cleanDistFolder() {
   if (fs.existsSync(paths.appDist)) {
-    return util.promisify(rimraf)(paths.appDist);
+    return util.promisify(rimraf)(`${paths.appDist}/*`);
   }
 }
 


sw-yx:
@grimnomad can you test if this fixes the issue?

grimnomad:
I tested the fix by manually altering the index.js on my machine. I don't know how to get this PR working with my project. But it doesn't work. CRA still fails, because there is no file at the time it tries to access the file.

This is what CRA prints:
```
Failed to compile.
../common/dist/index.js
Module not found: Can't resolve './common.cjs.development.js' in '<path/to/common>'
```

audiolion:
Bummer, I wonder if this is a different issue then. Could you make an example lerna repo where I could test this out?

grimnomad:
https://github.com/grimnomad/lerna_tsdx

This is a minimal example. From my point of view it is a race condition.

audiolion:
so when I run that example repo it works without issue :( (I had to add a `declare module 'common'` in a `common.d.ts` file under `src` to get rid of the warning).

image

grimnomad:
I am starting the watch task not the build task.

In my opinion the missing module is a consequence of the "error". If I start both tasks manually (first common, than web) everything is ok. If I start them via lerna in parallel I get the module error.

jaredpalmer:
Is this ready?

audiolion:
This change does not fix grimnomads issue, we can make this change if we want, it at least keeps the dist folder in existence if something else is relying/watching for it.

Grim's issue is that he is running a tsdx build process and a CRA start process at the same time, so CRA is trying to use files from another subpackage's dist folder and they are being moved/modified. I dont think there is a solution that we can provide for this, lerna would need to provide a way to sequentially order package scripts.

grimnomad:
@audiolion 
I closed my issue, because the `--noClean` flag is a good enough solution for me.

PR #277, opened 2019-10-21
Added support for eslint report

mfolnovic:
I've added new flag `--report-file report.json` to `tsdx lint` which is the same as `eslint -f json -o report.json`.

e.g. `tsdx lint src --report-file report.json`


sw-yx:
nice

PR #286, opened 2019-10-27
Remove confusing comment

techieshark:
The line removed didn't make sense in the context of this source.

It appears to have been copied from https://github.com/bvaughn/progress-estimator/blob/master/README.md#usage-example (in the context of the README it made more sense).


agilgur5:
@allcontributors please add @techieshark for docs

Issue #285, opened 2019-10-26
Generated typings are broken


z0al:
### Current Behavior

Generated typings are broken (literally) starting from `v0.10.0`. Here are some snippets:

```typescript
    interface Query extends Object {
        id: ID;
        type?: ch;
        ';;
        exp;
    }
```

```typescript
    interface Query extends Object {
        id: ID;
        type?: xport;
        fu;
    }
```

And YES, it duplicated almost every type declaration (see the full file [here](https://gist.github.com/z0al/78c73ab9d799533d79915a53da558697))

### Expected behavior

Typings don't have syntax errors.

### Suggested solution(s)

I think it's a bug in `@wessberg/rollup-plugin-ts`. I tried that lib before in a different project but quickly removed it due to the same issue.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | >= v0.10.0
| TypeScript       | 3.6.4
| Browser          | N/A
| npm/Yarn         |  6.11.3
| Node             | 12.8.1
| Operating System |  Debian Linux

z0al:
I tried updating `@wessberg/rollup-plugin-ts` to the latest version (`v1.1.73` instead of `v1.1.70`) but the output is still the same. I will try to create a minimal reproducible repo later.

samdenty:
For me, it's not exporting all the types. The dts declaration merging is broken


0.9.3 okay:
```ts
import { Type } from 'library'
```

1.10.0 error, type not exported

wessberg:
Sorry about that, do you have a minimal repro I can take a look at? It would be great if you submitted this as an issue over at [rollup-plugin-ts](https://github.com/wessberg/rollup-plugin-ts)

z0al:
No worries. I will do if I could reproduce it in a small repo, thanks.

wessberg:
@z0al, The issue with malformed typings you reported has been fixed in [the latest version of `rollup-plugin-ts`](https://github.com/wessberg/rollup-plugin-ts/releases/tag/v1.1.74).

Issue #274, opened 2019-10-20
Build fails on 0.10.5: Cannot find module @babel/...


Flydiverny:
### Current Behavior

```sh
$ tsdx build
Error: Cannot find module '@babel/plugin-transform-runtime' from '/path-to-pkgs/packages/auth'

    at Function.module.exports [as sync] (/path-to-pkgs/packages/auth/node_modules/resolve/lib/sync.js:74:15)
    at resolveStandardizedName (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/files/plugins.js:101:31)
    at resolvePlugin (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/files/plugins.js:54:10)
    at loadPlugin (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/files/plugins.js:62:20)
    at createDescriptor (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/config-descriptors.js:154:9)
    at /path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/config-descriptors.js:109:50
    at Array.map (<anonymous>)
    at createDescriptors (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/config-descriptors.js:109:29)
    at createPluginDescriptors (/path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/config-descriptors.js:105:10)
    at /path-to-pkgs/packages/auth/node_modules/@wessberg/rollup-plugin-ts/node_modules/@babel/core/lib/config/config-descriptors.js:63:49

error Command failed with exit code 1.
```

### Expected behavior

Build working like in 0.9.3 😄 

### Additional context

Build works fine in 0.9.3 but upgrading tsdx to latest breaks, doing yarn add `@babel/plugin-transform-runtime` leads to the same error with another babel plugin (`@babel/plugin-syntax-dynamic-import`) 

Guessing this issue is related: https://github.com/wessberg/rollup-plugin-ts/issues/37
### Your environment

| Software         | Version(s) |
[...]

PR #287, opened 2019-10-27
Rollback #208: Reinstate `rollup-plugin-typescript2` 

### LINKED ISSUE #268: No .d.ts created after update to 0.10.0
### Current Behavior

//index.ts
import { PrivateRoute, PublicRoute } from "./components";

export {
    PrivateRoute,
    PublicRoute
}
will generate 
//index.d.ts
/// <reference types="react" />
export { PrivateRoute, PublicRoute };
and directories of components with .d.ts arent created
[ss](https://i.imgur.com/2W6kqxM.png)

### Expected behavior

In 0.9.3 the .d.ts are created:
[ss](https://i.imgur.com/RfeuOHv.png)
//index.d.ts
import { PrivateRoute, PublicRoute } from "./components";
export { PrivateRoute, PublicRoute };

### Your environment

| TSDX           | >=0.10.0
| TypeScript       | ^3.6.4
| Browser          | 
| npm/Yarn         | v1.17.3
| Operating System | Arch

medelman17:
As described in #200, `tsdx` previously utilized `rollup-plugin-typescript2` to compile and check TypeScript files. On account of the fact that the above plugin cannot be used alongside other rollup plugins that contain async/await syntax without a workaround, `tsdx` migrated to `@wessberg/rollup-plugin-ts` in #208. However, as discussed in #268, #274, #285 and elsewhere, the above migration caused untoward consequences in certain instances for existing `tsdx` users with custom `babel` setups. 

This PR reverts support `@wessberg/rollup-plugin-ts` and restores `rollup-plugin-typescript2` as compiler in chief. 

Closes #268, #274, #285


markerikson:
Hmm.  We specifically updated Redux Starter Kit to use 0.10.6 because it started correctly merging all our declarations into a single `index.d.ts` file, vs a separate `.d.ts` file per source file.  Does this rollback still do things as a single `index.d.ts` file?

jaredpalmer:
Yes, I believe so. The issue was that the rollup-plugin-ts2 really did not work as advertised and introduced more problematic and inconsistent bugs such as duplicative references directives and incorrect declarations. @markerikson why does merging declarations matter?

markerikson:
I know @phryneas pointed out some issues where that might not work with certain Webpack setups, apparently.  See https://github.com/reduxjs/redux-starter-kit/pull/212#issuecomment-543033641 for discussion.

jaredpalmer:
@medelman17 @markerikson FWIW, my OG formik rollup setup (pre-tsdx) didn't use `rollup-plugin-typescript2` at all, instead it did a 2 pass compile. First pass used `tsc-watch` (which is just `tsc` with an `onSuccess` flag) to compile typescript to a temp directory. Second pass just used rollup to bundle that JS up and move the types to the dist dir. The clear benefit of this setup is that you can just use good ol' `tsc` and let it do it's thing. Then you can let people go buck wild with rollup.

wessberg:
> Yes, I believe so. The issue was that the rollup-plugin-ts2 really did not work as advertised and introduced more problematic and inconsistent bugs such as duplicative references directives and incorrect declarations.

Merging declarations, something that rollup-plugin-typescript2 doesnt do, is non-trivial. I think it's a bit harsh to right out say it is not working as advertised 🙂. I'm putting a lot of effort into fixing whatever issues that remains with declaration bundling and are addressing the issues one at a time.

jaredpalmer:
Didn't mean to come off as rude, nor does this have anything to do with lack of effort. I understand that you're working through bugs, but until those are fixed, we need to ensure that the published version of tsdx still works in the interim. Looking forward to when the declaration merging is stuff is sorted out. It's very cool and exciting.

tunnckoCore:
I was betting it will conflict. :laughing: The customBabelRollup plugin should be removed (entirely dropping `rollup-plugin-babel`) and passed as babelConfig to the `ts` plugin, I believe. But yea, I agree that we need to wait :)

markerikson:
@jaredpalmer : no, it looks like v0.11.0 is _not_ combining types into a single `index.d.ts` file.

I just tried updating Redux Starter Kit and rebuilding, and it's back to individual `.d.ts` files per source file.

smashercosmo:
hm... so, what's the recommended approach then to introduce additional rollup plugins that contain async/await?

Issue #265, opened 2019-10-17
Make things async wherever possible


jaredpalmer:
I noticed some sync file system tasks slipped into .10.x. We should make these async wherever possible.

Example: we should make this async https://github.com/jaredpalmer/tsdx/blob/76d7eb33e946c6dd3ebb656eabc874e392ecf012/src/index.ts#L476

tabuckner:
👀

gndelia:
I would like to work on this one 😄

gndelia:
Do I need to be formally assigned or should I start working on it directly?

sw-yx:
go ahead!

gndelia:
Hi
I have a draft PR, however I am having issues to commit - I am getting eslint errors under `test/tests/lint` - although that folder should be excluded. I get them even without my changes

I am running `yarn lint` (actually it is being run in the `pre-commit` hook)

```
$ yarn && yarn lint
yarn install v1.19.1
[1/4] Resolving packages...
success Already up-to-date.
$ tsc -p tsconfig.json
Done in 3.86s.
yarn run v1.19.1
$ yarn build && yarn lint:post-build
$ tsc -p tsconfig.json
$ node dist/index.js lint src test --ignore-pattern 'test/tests/lint'

C:\Users\DELL\Dev\Personal\tsdx\test\tests\lint\file-with-lint-errors.ts
  1:17  error  Parsing error: ',' expected

C:\Users\DELL\Dev\Personal\tsdx\test\tests\lint\file-with-prettier-lint-errors.ts
  1:18  error  Replace `·=·(···)·=>·⏎!!·('bar')⏎` with `=·()·=>·!!'bar'`  prettier/prettier
  4:1   error  Delete `⏎·`                                                prettier/prettier

C:\Users\DELL\Dev\Personal\tsdx\test\tests\lint\react-file-with-lint-errors.tsx
  4:9  error  Parsing error: Expression expected

✖ 4 problems (4 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```


Any ideas? My understanding is that those fields should be excluded from being analyzed

sw-yx:
@gndelia lets address those windows lint issues separately, and get this PR in as soon as you feel it is ready.

sw-yx:
some progress made on these thanks to @gndelia!

agilgur5:
[...]

PR #291, opened 2019-10-29
Replaced some sync methods for their async version

gndelia:
See #265 

I removed most of the sync functions, there are some scenarios which I am unsure what to do. For instance, those sync calls that are at module level (like `src/index.ts` line 50 or line 60)

Same for `src/errors/transformErrorMessages.ts` - I am unsure if I can make that function async

Still pending one change from `getProjectPath()` in `src/index.ts`

### DIFF
--- a/src/createEslintConfig.ts
+++ b/src/createEslintConfig.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import util from 'util';
 import { CLIEngine } from 'eslint';
 import { PackageJson } from './types';
 import { getReactVersion } from './utils';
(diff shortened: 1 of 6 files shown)


gndelia:
There are some options I was unable to migrate - I think it's not possible. Like the following https://github.com/jaredpalmer/tsdx/blob/master/src/index.ts#L50 and a couple more that are code part of a module and therefore can't be async (that's my understanding)

sw-yx:
mm i think we had some merge conflicts from other PRs. if you have time to address them, be my guest, if not i'll try to update when i free up

gndelia:
@sw-yx  I rebased and solve the conflicts :)

sw-yx:
Thank you!

agilgur5:
@all-contributors please add @gndelia for code

PR #307, opened 2019-11-06
feat: Add transpileOnly flag to watch and build command

### LINKED ISSUE #243: Compilation time grows with project size
### Current Behavior

When projects grow, transpilation time of TypeScript tends to grow with it because of the type checking. And while having types is awesome, especially when developing the longer compilation time is meh. Especially because most of the time your editor will tell you anyway that you messed up.

### Desired Behavior

Possibility to opt out of type checking.

### Suggested Solution

Add option (e.g. `--transpileOnly`) to skip type checking all together.
This can be directly passed to the rollup plugin as an option. [See docs](https://github.com/wessberg/rollup-plugin-ts#transpileonly).

In addition, we could also add a cli command that does only the type checking (e.g. `tsdx typecheck`). So people who are opted out, can still easily run diagnostics.

Everyone who runs into slow builds.

I tried to switch a project to tsdx and the build time grew to *6min* (sometime I even got a heap overflow error from V8 .... uuuuups). The project is very small, but the types seems to be very expensive. Removing type checking dropped the build time to **10s**.

Not using tsdx and configuring rollup ourselves. Noooes!

### Additional context

If this proposal sounds good, I am more than happy to work on it 🙂

sebald:
Adds `--transpileOnly` (alias `-T`) flag to `watch` and `build` commands. This will set the `check` option of roll up's typescript plugin to false.

Closes #243


README.md
```diff
@@ -17,7 +17,7 @@ Despite all the recent hype, setting up a new TypeScript (x React) library can b
   - [Development-only Expressions + Treeshaking](#development-only-expressions--treeshaking)
     - [Rollup Treeshaking](#rollup-treeshaking)
     - [Advanced `babel-plugin-dev-expressions`](#advanced-babel-plugin-dev-expressions)
-      - [`__DEV__`](#__dev__)
+      - [`__DEV__`](#dev)
```
sebald: I think prettier messed this up?

README.md
```diff
@@ -389,6 +392,7 @@ Options
   --tsconfig            Specify your custom tsconfig path (default <root-folder>/tsconfig.json)
   --verbose             Keep outdated console output in watch mode instead of clearing the screen
   --noClean             Don't clean the dist folder
+  -T, --transpileOnly   Skip type checking
```
jaredpalmer: Can we omit the short hand actually? We don't use it on others aside from help. I think we should open a new issue / pr and do them all at once.

sebald: Of course, my idea was to use the same args as `ts-node` so people feel familiar 🙂

sebald: PS: will change it tomorrow (it's late in 🇩🇪)

README.md
```diff
@@ -9,18 +9,18 @@ Despite all the recent hype, setting up a new TypeScript (x React) library can b
 
 - [Features](#features)
 - [Quick Start](#quick-start)
-  - [`npm start` or `yarn start`](#npm-start-or-yarn-start)
```
jaredpalmer: weird that the ` marks got removed?

jaredpalmer:
lgtm. Thanks! 

bitmoji

sw-yx:
thanks for the great PR! @all-contributors please add @sebald for docs, code, tests

PR #308, opened 2019-11-06
fix(lint): do not output warning about react on non-react projects

### LINKED ISSUE #279: Option to opt-out of eslint-plugin-react
### Current Behavior

`tsdx` supports `eslint-plugin-react` outof the box. If React is not used this warning pops up: 
```
Warning: React version was set to "detect" in eslint-plugin-react settings, but the "react" package is not installed. Assuming latest React version for linting.
```

### Desired Behavior

Opt-out of `eslint` plugins that are not in used. For example, I'm using `tsdx` to build a logger plugin for a specific library. I don't need `.eslintrc` extending`react-app`, because my work is only `node` related. 

### Suggested Solution

My suggestion is to remove any dedicated `eslint` plugin except the ones that are `typescript` and `prettier` related. User would manually add additional rules to the project if needed, for example `eslint-plugin-vue`, and configure it in the local `.eslintrc`.  This way `tsdx` remains decoupled from any specific library.

TypeScript users that develop plugins, libraries for NodeJS.

arthurdenner:
**Current behavior:**
We get a warning about detecting React version using the `lint` command in non-react projects.

**New behavior:**
Using the shared settings and the same value used by `eslint-plugin-react` [internally](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/util/version.js#L29) - it could be almost any value, but I chose this one, we don't get the warning on non-react projects. Closes #279.

**How to reproduce:**
- Use `tsdx lint` inside a react and a non-react project;
- No warning is displayed on both cases.

**Notes:**
N/A.

### DIFF
--- a/src/createEslintConfig.ts
+++ b/src/createEslintConfig.ts
@@ -1,15 +1,23 @@
 import fs from 'fs';
 import path from 'path';
 import { CLIEngine } from 'eslint';
+import { PackageJson } from './types';
+import { getReactVersion } from './utils';
 
 interface CreateEslintConfigArgs {
+  pkg: PackageJson;
   rootDir: string;
   writeFile: boolean;
 }
 export function createEslintConfig({
+  pkg,
   rootDir,
   writeFile,
 }: CreateEslintConfigArgs): CLIEngine.Options['baseConfig'] {
+  const isReactLibrary = Boolean(getReactVersion(pkg));
+
+  console.log(isReactLibrary);
+
   const config = {
     extends: [
       'react-app',
--- a/src/createEslintConfig.ts
+++ b/src/createEslintConfig.ts
@@ -25,7 +25,7 @@ export function createEslintConfig({
     settings: {
       react: {
         // Fix for https://github.com/jaredpalmer/tsdx/issues/279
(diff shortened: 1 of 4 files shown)


sw-yx:
looks good enough. ready to merge?

arthurdenner:
@sw-yx, I believe so. :smiley:

sw-yx:
thanks for the great pr! will go out with the next release

pcowgill:
Would you consider shipping a tiny patch release with this change? Or have some bigger changes already made it into master that make that tricky? Not a big deal either way, of course - just curious. Thanks!

sw-yx:
its up to @jaredpalmer - he controls the releases

EthianWong:
When to push this change to npm？i think i need it. @jaredpalmer

sw-yx:
thanks for the great PR! @all-contributors please add @arthurdenner for bugs, code

pcowgill:
Do you know if this change has been released yet? Thanks!

arthurdenner:
@pcowgill, yes, it was released on [`v0.12.0`](https://github.com/jaredpalmer/tsdx/releases/tag/v0.12.0).

PR #314, opened 2019-11-09
Fixed path in example

thchia:
I'm not sure if I'm missing something, but the import in `index.tsx` in the react example directory does not point to a valid location. It points to the package root, when I think it should point to either `src/index.tsx` or `dist/index.tsx` (but probably the former).


Carl-Foster:
When the path points to the package root, it loads the package from the `dist` folder, as defined in `package.json`. This means that the example code runs the compiled components which is ideal for testing.
It's noted in the docs to run both the `start` command for tsdx and the `example` package separately to enable this.

PR #315, opened 2019-11-09
Move most deps to devDependencies

FredyC:
There is no reason to install all those dev dependencies on consumer machines.

Besides, we don't want to rely on a global package across the team, so tsdx is part of a project deps and bloats node_modules quite heavily. Also spits a lot a peerDeps warnings for most of those deps.

The list could be probably trimmed down even more, but this is a start of the obvious ones. I don't want to dive into the code right now to find out true runtime dependencies.


jaredpalmer:
@sw-yx is this okay?

thatsmiddi:
I don't think all of those are development dependencies. Since this package is an abstraction over other build tools these would need to be installed on the consumers device, too.

I just did a quick test by creating a package, installing it in another project and run the `tsdx create lib` command.

During installation I am getting:
```
npm WARN rollup-plugin-babel@4.3.3 requires a peer of @babel/core@7 || ^7.0.0-rc.2 but none is installed. You must install peer dependencies yourself.
npm WARN rollup-plugin-typescript2@0.25.2 requires a peer of typescript@>=2.4.0 but none is installed. You must install peer dependencies yourself.
```

Running `tsdx create lib` produces:
```
➜ node_modules/.bin/tsdx create lib
internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'jest'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/Users/hmittrop/dev/awesome-library/node_modules/tsdx/dist/index.js:12:40)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

FredyC:
Oh, I did not even know that TSDX allows running ESLint and Jest. I never felt the need to use those. In that case, I guess it makes sense to keep all those deps there.

PR #327, opened 2019-11-14
(fix): respect tsconfig esModuleInterop flag

### LINKED ISSUE #165: Default export not output as expected
### Current Behavior

TSDX doesn't seem to have the expected output when building a package with a default export. The type definitions are output as expected, but the resulting `.js` file actually exports as `{ default: /* value */ }`. I've created a minimal, reproducible example at [rtbenfield/tsdx-default-export](https://github.com/rtbenfield/tsdx-default-export) bootstrapped from `tsdx create`. This is my first time using TSDX and it is possible I am just doing something incorrectly.

In the repo mentioned above, I changed the `sum` function to be a default export, then added `output.test.ts` which imports directly from `dist` instead of `src`. `output.test.ts` logs the value of `import sum from "../dist"` and is showing `sum` to be `{ default: ... }` instead of the actual function.

### Expected behavior

The `.js` output files in `dist` can be consumed as default exports (ex. `import sum from "my-module"`) if the source file utilized default exports.

### Suggested solution(s)

Unsure, but willing to investigate if this is confirmed to be an issue. I know Webpack, but not Rollup, so it could be a good excuse for me to get some experience with it.

### Additional context

Nothing relevant that I know of.

### Your environment

| TSDX           | 0.7.2
| TypeScript       | 3.5.2
| Browser          | N/A
| npm/Yarn         | 6.9.0 / 1.16.0
| Operating System | Windows 10 Pro 1903

agilgur5:
- the previous assumption that ESM users will always import the ESM
  build doesn't actually hold up in practice due to various limitations
  in Node and testing and the prevalence of transpiling in general
  - even if one uses mostly ESM, many tools will require and use
    the CJS build, and this option would break compatibility

- without esModule set, CJS users would be unable to use the default
  export of any library built with tsdx
  - and this was not documented, meaning it would cause unintended
    breaking changes in libraries that use default exports and try to
    adopt tsdx
  - it would also break compatibility with certain tooling

- so, instead of always setting rollup's esModule to false and
  potentially causing unintended consequences, respect the user's
  esModuleInterop config as set in their tsconfig, but default to
  false

<hr>

Fixes #165 according to my suggestion in https://github.com/jaredpalmer/tsdx/issues/165#issuecomment-514924048 . Mine and other folk's comments in that issue give more details about the compatibility and adoption issues that drive this PR. Ultimately, I believe this is a bugfix as the current behavior causes unintended consequences and because default exports are _already_ output, just not with `__esModule` in CJS builds.

Have tested this successfully locally. All it does is, if `esModuleInterop` is set to `true` in `tsconfig.json`, will add a line of code to the CJS output as such:

```javascript
Object.defineProperty(exports, '__esModule', { value: true });
```

As I mentioned in https://github.com/jaredpalmer/tsdx/issues/165#issuecomment-511997338, everything else is the same as before, as ESM builds _already_ output `export default` and CJS builds _already_ output `exports.default`. `__esModule` is just a hint for transpilers to use `.default` when transpiling default _imports_ to CJS.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -1,4 +1,9 @@
-import { safeVariableName, safePackageName, external } from './utils';
+import {
+  safeVariableName,
+  safePackageName,
+  external,
+  resolveApp,
+} from './utils';
 import { paths } from './constants';
 import { terser } from 'rollup-plugin-terser';
 import { DEFAULT_EXTENSIONS } from '@babel/core';
@@ -12,6 +17,7 @@ import typescript from 'rollup-plugin-typescript2';
 import { extractErrors } from './errors/extractErrors';
 import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
+import * as fs from 'fs-extra';
 
 const errorCodeOpts = {
   errorMapFilePath: paths.appErrorsJson,
@@ -39,6 +45,11 @@ export function createRollupConfig(opts: TsdxOptions) {
     .filter(Boolean)
     .join('.');
 
+  let tsconfigJSON;
+  try {
+    tsconfigJSON = fs.readJSONSync(resolveApp('tsconfig.json'));
+  } catch (e) {}
+
   return {
     // Tell Rollup the entry point to the package
     input: opts.input,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
@benlesh does this solve your issue?

agilgur5:
@jaredpalmer it's been almost a month since I made this PR to fix the well-upvoted default export bug (#165), what can we do to get this merged?

I'm currently running a fork of `tsdx` in [a](https://github.com/agilgur5/mst-persist) [few](https://github.com/agilgur5/jest-without-globals) of my libraries in order to get this bugfix (as well as my other improvement at #329 ) and would prefer not have to keep using/maintaining a fork for too long, especially when there don't seem to be any issues with merging it.

PR #329, opened 2019-11-14
(change): use node_modules/.cache/... as cacheRoot

### LINKED ISSUE #328: Why not use the default cache directory (`node_modules/.cache/rollup-plugin-typescript2`)?
### Current Behavior

Currently `tsdx` outputs to `./.rts2_cache_*` and requires you to have a `gitignore` for these directories.

### Desired Behavior

Use [`rollup-plugin-typescript2`'s default](https://github.com/ezolenko/rollup-plugin-typescript2#plugin-options), which is `node_modules/.cache/rollup-plugin-typescript2`. This way there's no need to gitignore anything extra (as `node_modules` is usually gitignored already) and it doesn't clutter up the app root with more gitignored directories. [`babel-loader` has a similar default directory](https://github.com/babel/babel-loader#options) in `node_modules/.cache/babel-loader`, but it doesn't let you change the directory.

### Suggested Solution

Remove the `cacheRoot` configuration of `rollup-plugin-typescript2` on https://github.com/jaredpalmer/tsdx/blob/6252b705eb0fa29ac89a81d47ea676c9d5cda791/src/createRollupConfig.ts#L131. Alternatively, use `node_modules/.cache/rollup-plugin-typescript2_${opts.format}` so it retains the cache suffix, but isn't located externally.

This would also mean the default gitignores no longer need these directories.

**EDIT**: see #329 for an implementation of this suggested solution.

All users, but it shifts the cache into entirely internal.

Added an alternative in suggested solutions.

### Additional context

See `rollup-plugin-typescript2` and `babel-loader` links above.

I'm not sure what the original rationale / explanation for having the cache be outside of `node_modules/.cache/` and different from the default `cacheRoot` was. I couldn't find any rationale in any comments, issues, PRs, or in the commit description of [the commit](https://github.com/jaredpalmer/tsdx/commit/6bbf07793424511cc245f62f52694925fe452342) that added this config, so I don't know what the reasoning behind that was. It also didn't seem like that option was carried over in #208 (it doesn't seem like `rollup-plugin-ts` has that option though), but it was brought back in the rollback #287.

I think it would be useful to at least elaborate on that here as that was a somewhat consequential decision.

agilgur5:
- instead of outputting these to app root and requiring users to
  gitignore these directories, just output to node_modules/.cache/
  - rollup-plugin-typescript2 is the default cache directory [1], so this
    initially appended _${opts.format} to it
    - upon code review, changed it to tsdx/${opts.format}/
    - babel-loader similarly uses node_modules/.cache/babel-loader [2]
      as its default cache directory and does not allow users to
      change it
      - parcel uses node_modules/.cache/parcel too [3]
      - fuse-box uses node_modules/.fusebox [4]

[1]: https://github.com/ezolenko/rollup-plugin-typescript2#plugin-options > `cacheRoot`
[2]: https://github.com/babel/babel-loader#options > `cacheDirectory`
[3]: https://github.com/parcel-bundler/parcel#caching
[4]: https://github.com/fuse-box/fuse-box/blob/master/docs/getting-started/get-started.md#caching
<hr>

Fixes #328 . Not sure if this change makes sense as, per the issue, I don't know / couldn't find the rationale for using the app root `.rts2_cache_*` directories. I'm also not sure if just removing the `cacheRoot` option and using the default may cause issues -- I'm assuming the `${opts.format}` is there because it does, so I left it in while moving the cache into `node_modules`.

Feel free to close if this doesn't make sense, #328 is more of a question, but I figure I might as well make the small changes I suggested for it in case it does.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -128,7 +128,7 @@ export function createRollupConfig(opts: TsdxOptions) {
       },
       typescript({
         typescript: require('typescript'),
-        cacheRoot: `./.rts2_cache_${opts.format}`,
+        cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
           compilerOptions: {
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 *.log
 .DS_Store
 node_modules
-.rts2_cache_cjs
-.rts2_cache_esm
-.rts2_cache_umd
-.rts2_cache_system
 dist
 tester
 tester-react
--- a/templates/basic/gitignore
+++ b/templates/basic/gitignore
@@ -1,8 +1,4 @@
 *.log
 .DS_Store
 node_modules
-.rts2_cache_cjs
-.rts2_cache_esm
-.rts2_cache_umd
-.rts2_cache_system
 dist
--- a/templates/react/gitignore
+++ b/templates/react/gitignore
@@ -2,8 +2,4 @@
 .DS_Store
 node_modules
 .cache
-.rts2_cache_cjs
-.rts2_cache_esm
-.rts2_cache_umd
-.rts2_cache_system
 dist


src/createRollupConfig.ts
```diff
@@ -128,7 +128,7 @@ export function createRollupConfig(opts: TsdxOptions) {
       },
       typescript({
         typescript: require('typescript'),
-        cacheRoot: `./.rts2_cache_${opts.format}`,
+        cacheRoot: `./node_modules/.cache/rollup-plugin-typescript2_${opts.format}`,
```
jaredpalmer:
```suggestion
        cacheRoot: `./node_modules/.cache/tsdx_${opts.format}`,
```

agilgur5: Sure, if you want to use `tsdx` instead this time around. Amended that to the commit but went a step further as `tsdx/${opts.format}/` so it's all under the `tsdx` cache directory.

agilgur5:
Tested this [with my fork](https://github.com/agilgur5/tsdx/tree/dist) and it creates the new directories successfully and starts caching there with no problems (as far as I can tell at least) 👍

PR #335, opened 2019-11-15
feat: --noClean option for build

hardfist:
I want to build multi file for my library, unfortunately tsdx will clean previous build assets, which makes it's hard to implement, It will be conveniently to let user decide whether to clean dist file
image


sw-yx:
thanks for your PR - please open issues for discussion first when you want to contribute features! especially when adding a new flag, since we want to make sure we dont add flag bloat. thank you.

PR #361, opened 2019-12-04
Remove unnecessary yarn install command in GH action

sw-yx:
this command is breaking the windows build for our CI and causing a lot of greenkeeper false positives. and based on [this](https://github.com/actions/setup-node/commit/57adacb752d56f434eed924bad6ef995a163861b) doesnt seem to be necessary

making this a PR in case this works for github


sw-yx:
aww yissss 
image

Issue #348, opened 2019-11-22
Bump Typescript to `typescript@^3.7`


NoahZinsmeister:
I'd love to use the latest [Typescript release](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) with tsdx, and unless I'm mistaken tsdx always uses its own hard-coded version when building, which can't be overridden by a local version.

tlaak:
I would also be interested in using the latest version as it supports optional chaining. Is this something that can be achieved by using a configuration?

sw-yx:
it could but we should also update our TS version. i'll have a look

jeanbenitez:
https://github.com/jaredpalmer/tsdx/pull/363 closes this. it isn't?

NoahZinsmeister:
yeah the github auto-close issue thing must have been broken

Issue #341, opened 2019-11-20
deprecated rollup-plugin-json & rollup-plugin-replace


isubasti:
rollup-plugin-json & rollup-plugin-replace packages was moved to a new repo and the old ones are no longer being maintained

```
warning tsdx > rollup-plugin-json@4.0.0: This module has been deprecated and is no longer maintained. Please use @rollup/plugin-json.
warning tsdx > rollup-plugin-replace@2.2.0: This module has moved and is now available at @rollup/plugin-replace. Please update your dependencies. This version is no longer maintained.
```

sw-yx:
thanks, will roll this up in to a dep bump PR

PR #362, opened 2019-12-04
update all deps including typescript 3.7

sw-yx:
update deps and run tests all at once since greenkeeper was kinda broken

also closes https://github.com/jaredpalmer/tsdx/issues/341

and closes https://github.com/jaredpalmer/tsdx/issues/348


sw-yx:
ugh why is this `Error: Cannot find module '../package.json'` on windows

PR #364, opened 2019-12-04
update rollup deps and plugins

sw-yx:
continuing incremental dep upgrades since https://github.com/jaredpalmer/tsdx/pull/362 failed

closes https://github.com/jaredpalmer/tsdx/issues/341


sw-yx:
weird, this passes too...

PR #368, opened 2019-12-12
fix(chore): fixed minor typo

jamesgeorge007:
#291 replaced all instances of `sync` versions of some methods with their corresponding `async` versions. Updated the comment accordingly.


sw-yx:
thanks!!

PR #370, opened 2019-12-14
Add optional chaining and nullish coalescing operators support

### LINKED ISSUE #369: Support for the experimental syntax 'optionalChaining'
### Current Behavior

Could not use optional chaining with tsdx@0.11.0 and typescript@3.7.3:
```
(babel plugin) SyntaxError: /some-path/Foo.ts: Support for the experimental syntax 'optionalChaining' isn't currently enabled (1:19):
> 1 | console.log(foo.bar?.foo)
```

### Desired Behavior

Optional chaining is enabled

### Suggested Solution

- Issue has been solved in https://github.com/facebook/create-react-app/pull/7438/files with 2 new babel plugins:
```
// Adds syntax support for optional chaining (.?)
{ name: '@babel/plugin-proposal-optional-chaining' },
// Adds syntax support for default value using ?? operator
{ name: '@babel/plugin-proposal-nullish-coalescing-operator' },
```

LoicMahieu:
Closes #369

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "mkdirp": "^0.5.1",
     "ora": "^3.4.0",
     "pascal-case": "^2.0.1",
-    "prettier": "^1.18.2",
+    "prettier": "^1.19.1",
     "progress-estimator": "^0.2.2",
     "rimraf": "^3.0.0",
     "rollup": "^1.27.8",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "@babel/core": "^7.4.4",
     "@babel/helper-module-imports": "^7.0.0",
     "@babel/plugin-proposal-class-properties": "^7.4.4",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.7.4",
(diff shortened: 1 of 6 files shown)


test/fixtures/build-default/src/foo.ts
```diff
@@ -1 +1 @@
-export const foo = () => 'bar';
+export const foo = (foo?: { bar: string }) => foo?.bar || 'bar';
```
agilgur5: If a test is going to be added for this, I think it should be a separate file named appropriately, instead of changing an existing test.

Also this only tests optional chaining, but nullish coalescing was also added here. If there will be a test for this, then it should test both (also PR name should probably be changed to note it also supports nullish coalescing usage)

LoicMahieu: I added a `syntax` directory which contains a specific file for each syntax features. Thanks

sw-yx:
thanks guys!

sw-yx:
thanks for the great PR! @all-contributors please add @LoicMahieu  for code, tests

Issue #171, opened 2019-07-30
Tree-shaking not working as expected


wKovacs64:
### Current Behavior

When building a project with `tsdx` and exporting multiple things (all named exports), the dependencies (imports) of unused modules (that is, modules the consuming project does not import) all end up in the final consuming bundle. So, importing `Thing1` will also import dependencies of `Thing2` (even though `Thing2` itself is tree-shaken away). I'm probably just doing something wrong or misunderstanding what's possible, though.

### Expected behavior

I was hoping that if I only import `Thing1`, `Thing2`'s dependencies (imports) would not be included in the final consuming bundle.

### Suggested solution(s)

Unknown (so far), might not even be a bug.

### Additional context

Reproduction (includes build output and webpack-bundle-analyzer):

https://github.com/wKovacs64/tsdx-treeshaking-test

I've gotten around this in the past with a Rollup config that takes all the source files as an array of inputs instead of a single entry point and outputs them all individually to a directory (Rollup `output.dir`) rather than a single output file (Rollup `output.file` like `tsdx` uses), but that may not be the right:tm: solution nor does it look possible at the moment as `tsdx` does not currently support a customized Rollup configuration.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.7.2
| TypeScript       | 3.5.3
| Browser          | N/A
| npm/Yarn         | Yarn 1.17.3
| Operating System | Windows 10, macOS

TrySound:
Could you commit please gatsby output?

wKovacs64:
Done ([`consumer/public`](https://github.com/wKovacs64/tsdx-treeshaking-test/tree/master/consumer/public)), thanks for looking at it!

TrySound:
Could you point me where you found both functions? I'm not experienced with gatsby.

wKovacs64:
To be clear, I'm not seeing both functions (`Thing1` and `Thing2`), I'm seeing `Thing1` (good) and Emotion (bad). I'm not sure exactly where in the build output it lives, I just observed it in webpack-bundle-analyzer and the overall size of the final output directory. I tried it with Parcel as well (not included in the reproduction repo), and I suspect (untested) Create React App will include it too.

wKovacs64:
I added 2 more example consumer projects, one for Create React App and one for Parcel (essentially the tsdx example project), if that helps.

[...]

PR #371, opened 2019-12-14
(types/fix): explicit Rollup typing, fix treeshake location

agilgur5:
- treeshake was in output, but it's not a config of output :/ :/ :/
  - once explicit types were added, this was a big red underline :/ :/

Yeaaaaaaaaa so that's kind of a problem 😕 😕 😕 Related to #365 and potentially related to #171. This bug has been around for a long enough time that it's fix may cause some breakage, idk.

Related: Does the `eslint` config error on getters and setters? That wouldn't help with imported code, but would at least reduce unexpected errors in one's own codebase.

Also should we add some regression tests for treeshaking? It's typed correctly _now_, but might make sense to have a test for something like this that could be considered critical and hard to catch, idk.

<hr>

I was going around adding types to builder code during #367 because type-checking would add a lot more confidence and discovered this along the way -- figured it should be extracted out as its own commit.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -5,6 +5,7 @@ import {
   resolveApp,
 } from './utils';
 import { paths } from './constants';
+import { RollupOptions } from 'rollup';
 import { terser } from 'rollup-plugin-terser';
 import { DEFAULT_EXTENSIONS } from '@babel/core';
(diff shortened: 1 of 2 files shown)


sw-yx:
nice catch - who woulda thought that typechecking is useful for this stuff 😂 

wanna make sure jared sees this so not clicking merge

PR #372, opened 2019-12-14
(fix/format): formatting of #366 didn't pass lint

agilgur5:
It doesn't seem like GH Actions / CI are running for each PR, as no status checks appear for them --    probably why #366 slipped through. I'm not sure why they're not running, looks normal to me though I've never used GH Actions before for CI.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -577,9 +577,9 @@ prog
         ...jestConfig,
       })
     );
- 
+
     if (!process.env.CI) {
-      argv.push('--watch') // run jest in watch mode unless in CI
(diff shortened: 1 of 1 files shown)


sw-yx:
thanks my bad.

not sure what happened to the CI too, i even spent some time fixing it

sw-yx:
btw  i can see its running fine here.. just not showing up in the UI
image

agilgur5:
> not sure what happened to the CI too, i even spent some time fixing it

Followed up on this at #373 .

Issue #336, opened 2019-11-15
Imports of the entry module are not bundled


jlkiri:
### Current Behavior

I have an `src` folder which has `index.ts` in it and two subfolders `plugins` and `utils`, each with more TS files. `index.ts` imports most of these files from subfolders, but they are neither bundled nor independently compiled. Only their types are correctly output, however.

### Expected behavior

1. Output a compiled JS file for each TS file imported by `index.ts` so it can be imported by the compiled `index.js`
OR
2. Bundle each originally imported file inside the `index.js` output file

### Suggested solution(s)

Currently no idea.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.11.0
| TypeScript       | 3.7.2
| Browser          | 
| npm/Yarn         | 6.11.3 / 1.19.1
| Node             | 10.16.3
| Operating System | MacOS

jlkiri:
I figured out that this is only the case if imports within source files are made relative to `paths` option of `tsconfig.json`, such as `src/*`. If imports are relative to the importing file itself everything is alright.

sw-yx:
this is something we need to figure out. i suspect its a rollup fix we need to do. if you get time, please be my guest to investigate

ambroseus:
@sw-yx  @jlkiri 
found solution: https://github.com/zerkalica/zerollup/tree/master/packages/ts-transform-paths#setup-for-rollup-plugin-typescript2
dirty fix: 
![Screenshot from 2019-12-15 12-55-28](https://user-images.githubusercontent.com/380645/70861673-3cfd0300-1f3a-11ea-874b-7d6c469eca60.png)

ambroseus:
forked, fixed, checked build (using @ambroseus/tsdx) : https://github.com/DCKit/dckit/blob/master/packages/%40dckit/ui/tsconfig.json

ambroseus:
@jlkiri  while PR is under review, you can test fixed version `yarn add @ambroseus/tsdx`

ambroseus:
@jlkiri  solution: https://github.com/jaredpalmer/tsdx/issues/379#issuecomment-568202209

jlkiri:
@ambroseus Thanks, seems to be working now. I'm closing the issue.

agilgur5:
Related to / duplicate of #91

Issue #294, opened 2019-10-30
Support objectHashIgnoreUnknownHack flag from rpt2


arthurdenner:
### Current Behavior
I'm using tsdx to bundle [react-semantic-ui-datepickers](https://github.com/arthurdenner/react-semantic-ui-datepickers) and I can't [upgrade `rollup-plugin-copy`](https://github.com/arthurdenner/react-semantic-ui-datepickers/pull/64) to v3 because the build fails with the error `Unknown object type "asyncfunction"`. This is because we are not passing the flag `objectHashIgnoreUnknownHack` to `rpt2` which would solve the issue.

Can we add support for this flag on the `build` command or is that kind of risky? The `rpt2` documentation states: `Setting this option to true will make object-hash ignore unknowns, at the cost of not invalidating the cache if ignored elements are changed.`

### Desired Behavior
Build can be successful if the plugin gets the flag as `true`.

### Suggested Solution
Add an option to the `build` command - `--ignore-unknown-hack`? - that results in the `objectHashIgnoreUnknownHack` being true while creating the `rollup` config.

### Who does this impact? Who is this for?
`tsdx` users with custom `rollup` configs using plugins that use async/await, for example.

### Describe alternatives you've considered
N/A.

### Additional context
N/A.

sw-yx:
i dont 100% understand. what is rpt2? why are we adding a flag just to pass a flag to a specific dependency others may not have? i definitely dont like the suggested solution but think there is a bigger solve here. please explain your rpt2 usage.

arthurdenner:
Hi @sw-yx! Sorry for not being so clear at first. 

So, `rpt2` stands for `rollup-plugin-typescript2`, that `tsdx` is using again now (see #287). That's its internal name.

The plugin has a property called `objectHashIgnoreUnknownHack` which should be used when you want to compile a plugin that uses some features, e.g. async/await.

I'm extending `tsdx`'s default rollup config to use `rollup-plugin-copy`, but its latest version uses async/await, therefore, my building is failing because `rpt2` can't compile it.

I'm suggesting adding a flag to `tsdx build` that will be passed to `rpt2` as `objectHashIgnoreUnknownHack: true` or adding this flag as true by default on `tsdx`, but the second option might not be so good because `rpt2` advises you not to use the flag unless you really need - which is my use case :sweat_smile:.

Does this clarify the problem?

sw-yx:
[...]

PR #374, opened 2019-12-15
(fix): added rollup plugin @zerollup/ts-transform-paths, injected as …

ambroseus:
…transformer in rollup-plugin-typescript2 config

closes https://github.com/jaredpalmer/tsdx/issues/336

---

_Edit/Summary From @jaredpalmer:_

When using tsdx with larger packages, it is useful to use typescript path aliasing to avoid relative imports. This change allows tsdx (rollup and typescript) to respect such options.

**Before**
```
import Button from 'components/Button'
// becomes
const Button = require('components/button') // omitting interop stuff
```

**After**

```
import Button from 'components/Button'
// becomes
const Button = require('./components/button') // omitting interop stuff
```

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -14,6 +14,7 @@ import replace from '@rollup/plugin-replace';
 import resolve from 'rollup-plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import typescript from 'rollup-plugin-typescript2';
+import transformPaths from '@zerollup/ts-transform-paths';
 import { extractErrors } from './errors/extractErrors';
 import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
@@ -49,6 +50,9 @@ export async function createRollupConfig(opts: TsdxOptions) {
   try {
     tsconfigJSON = fs.readJSONSync(resolveApp('tsconfig.json'));
   } catch (e) {}
+
+  const pathTransformer = (service: any): any =>
+    transformPaths(service.getProgram());
 
   return {
     // Tell Rollup the entry point to the package
@@ -153,6 +157,7 @@ export async function createRollupConfig(opts: TsdxOptions) {
             target: 'esnext',
           },
         },
+        transformers: [pathTransformer],
       }),
       babelPluginTsdx({
         exclude: 'node_modules/**',
--- a/package.json
+++ b/package.json
@@ -1,5 +1,8 @@
 {
-  "name": "tsdx",
+  "name": "@ambroseus/tsdx",
+  "publishConfig": {
+    "access": "public"
+  },
   "version": "0.11.0",
   "author": "Jared Palmer <jared@palmer.net>",
   "description": "Zero-config TypeScript package development",
@@ -48,6 +51,7 @@
     "@types/shelljs": "^0.8.5",
     "@typescript-eslint/eslint-plugin": "^2.12.0",
     "@typescript-eslint/parser": "^2.12.0",
+    "@zerollup/ts-transform-paths": "^1.7.8",
     "ansi-escapes": "^4.2.1",
     "asyncro": "^3.0.0",
     "babel-eslint": "^10.0.3",
--- a/package.json
+++ b/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "@ambroseus/tsdx",
-  "publishConfig": {
-    "access": "public"
-  },
+  "name": "tsdx",
   "version": "0.11.0",
   "author": "Jared Palmer <jared@palmer.net>",
   "description": "Zero-config TypeScript package development",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1169,6 +1169,20 @@
     semver "^6.3.0"
     tsutils "^3.17.1"
 
+"@zerollup/ts-helpers@^1.7.4":
+  version "1.7.4"
+  resolved "https://registry.yarnpkg.com/@zerollup/ts-helpers/-/ts-helpers-1.7.4.tgz#59ac8c643b8cc6def5449691630091226991da36"
+  integrity sha512-8JLREeiO6MtmJhU8kw/eNwi+HKlh/K3cMlpH6HVbW53bqhAgTvt8KnIqFh9FMNbCL3QDht35g8VJHVWYkMEt2A==
+  dependencies:
+    resolve "^1.12.0"
+
+"@zerollup/ts-transform-paths@^1.7.8":
+  version "1.7.8"
+  resolved "https://registry.yarnpkg.com/@zerollup/ts-transform-paths/-/ts-transform-paths-1.7.8.tgz#e9de4ed23e9b6f584b9a625141326456f71d0a8a"
+  integrity sha512-xh6KZRF3cD/pYeQypADTh7U9zAU8vgG7PvUT9m20mW8rsiUF69OKDM+9CsR483ceXVh5PC5AHUA/GAvsJ1iVbw==
+  dependencies:
+    "@zerollup/ts-helpers" "^1.7.4"
+
 abab@^2.0.0:
   version "2.0.2"
   resolved "https://registry.yarnpkg.com/abab/-/abab-2.0.2.tgz#a2fba1b122c69a85caa02d10f9270c7219709a9d"


ambroseus:
tsdx tests: 
Screenshot from 2019-12-15 14-05-33

my monorepo build with fixed version:
https://github.com/DCKit/dckit/blob/master/packages/%40dckit/ui/tsconfig.json
Screenshot from 2019-12-15 14-39-22
Screenshot from 2019-12-15 14-37-18

sw-yx:
ok nice work, think this one needs @jaredpalmer to look at it before we merge

ambroseus:
> ok nice work, think this one needs @jaredpalmer to look at it before we merge

yep)

meanwhile I've checked official `@reduxjs/toolkit`  build with fixed version

Screenshot from 2019-12-15 18-00-22

src/createRollupConfig.ts
```diff
@@ -50,6 +51,9 @@ export async function createRollupConfig(opts: TsdxOptions) {
     tsconfigJSON = fs.readJSONSync(resolveApp('tsconfig.json'));
   } catch (e) {}
 
+  const pathTransformer = (service: any): any =>
+    transformPaths(service.getProgram());
```
agilgur5: Can these be more specific than `any`? `transformPaths` is written in TS, so I think it can be.

Also do you know if the second arg isn't necessary? It has an `exclude` option, not sure if we should pass `tsconfigJSON.exclude` into there or if it's read automatically (and the `exclude` option is just an override maybe?? idk the docs aren't comprehensive)

ambroseus: oh, yea) `transformPaths` has types LanguageService and CustomTransformer - need to use,
to be fixed

jaredpalmer:
sorry did not mean to close

jaredpalmer:
Questions:

- AFAIK TS transforms are not that popular? Should/can we does this with babel? am i talking out of ass? idk. 
- How/does this impact monorepo's with symlinks and ts paths? For example, you can setup lerna and ts to understand symlinked local packages using tsconfig paths. The result is that that you can have  `@project/a` resolve to `packages/a` and  `@project/b` resolve to `packages/b` inside of `packages/c`. Thoughts?

ambroseus:
> * AFAIK TS transforms are not that popular? Should/can we do this with babel? am i talking out of ass? idk.

mm.. yeah, I'll check possible solution with babel

> * How/does this impact monorepo's with symlinks and ts paths? For example, you can setup lerna and ts to understand symlinked local packages using tsconfig paths

I already use lerna monorepo + tsdx (my patched version with ts-paths plugin) in each package + test CRA app  + `lerna bootstrap --hoist`
https://github.com/DCKit/dckit/blob/master/packages/%40dckit/ui/tsconfig.json
it works)

ambroseus:
@jaredpalmer  @agilgur5  we can avoid use of transformers in rpts2 and use `babel-plugin-module-resolver` with paths from `tsconfig.json`.  just compare (checked with 'basic' tsdx template)

Screenshot from 2019-12-19 23-24-46

ambroseus:
OMG.. https://github.com/tleunen/babel-plugin-module-resolver/issues/336#issuecomment-458906808

ambroseus:
Useful info: https://mitchellsimoens.com/2019/08/07/why-typescript-paths-failed-me/

ambroseus:
@agilgur5 @jaredpalmer  proof with 'storybook' template & `ts-transform-paths` :)

Screenshot from 2019-12-20 17-15-45

src/createRollupConfig.ts
ambroseus: https://github.com/zerkalica/zerollup/issues/22

jaredpalmer:
So my question is if this should be part of the tsdx out of the box. Is this really that common of a thing? Seems like it is niche? Someone could easily just publish a “plugin” that adds this and we could link to it

ambroseus:
> So my question is if this should be part of the tsdx out of the box. Is this really that common of a thing? Seems like it is niche? Someone could easily just publish a “plugin” that adds this and we could link to it

@jaredpalmer  I know many developers who use aliases plugins to deal with "../../../" hell.. But also agree that adding yet another plugin for plugin for plugin in the core is not excellent idea..

so.. I reject my PR and add comment to HOWTOs how to setup aliasing (@sw-yx it would be good to pin HOWTOs issue.. :)

agilgur5:
So just to summarize:

> OMG.. [tleunen/babel-plugin-module-resolver#336 (comment)](https://github.com/tleunen/babel-plugin-module-resolver/issues/336#issuecomment-458906808)
> solution with `babel-plugin-module-transformer` is not so simple like with `ts-transform-paths` transformer...

1. `babel-plugin-module-resolver` is a lot more complex than `ts-transform-paths`

> Useful info: https://mitchellsimoens.com/2019/08/07/why-typescript-paths-failed-me/

2.
    1. TS paths are not meant to be used like this per https://github.com/Microsoft/TypeScript/issues/10866#issuecomment-246789510 .
    1. Adding a transformer like `ts-transform-paths` can cause even more issues listed in that blog post, summarized at the end as: "I’m not saying paths are bad just if you’re gonna use them in a way they are not intended to be used, expect a bumpy road".
    1. [`tscpaths`](https://github.com/joonhocho/tscpaths) and [`ts-transform-import-path-rewrite`](https://github.com/dropbox/ts-transform-import-path-rewrite) are some alternatives (there are more), but seem more immature than `ts-transform-paths`, at least afaict from quick glance.

So `babel-plugin-module-resolver` actually seems to be a more appropriate use than a TS transform, but neither should really change the usage of `paths` as they shouldn't be used like that.

In either case, it seems like trying to resolve paths is more of a hassle than worthwhile, and potentially a very fragile feature, so I think I'd agree that it should be left to user-land as a result. That being said, as it is a common use case, I do think there should be an easier way to configure it.

As `objectHashIgnoreUnknownHack` of #294 et al is also a common use, it probably makes sense more generally to expose the `rpts2` config somehow so it can be modified without having to straight copy the source code (which is brittle as that will change). Wrapping `rpts2` in a plugin like I mentioned in https://github.com/jaredpalmer/tsdx/issues/389#issuecomment-568866380 might make overriding it a bit easier, but maybe there's another method that makes sense for this as well.

0vidiu:
Copy/pasta snippet incoming for people in need:

```js
// babel.config.js
module.exports = {
  // ...
  plugins: [
    // ... other plugins
    [
      'babel-plugin-module-resolver',
      {
        root: './',
        alias: require('./tsconfig.json').compilerOptions.paths,
      },
    ],
  ],
};
```

This works wonderfully in combination with `TsconfigPathsPlugin` (Webpack plugin). I declare my aliases in `tsconfig.json` and they work in Webpack, Storybook, Babel (for tests), while it also keeps VSCode happy. Hope this helps someone.

ambroseus:
thanks @0vidiu, added to HOWTOs https://github.com/jaredpalmer/tsdx/issues/379#issuecomment-568202209

slorber:
thanks @0vidiu @ambroseus 

This workaround works fine for the JS output, but the typedefs are emitted in different files, and they still have the absolute links, making the typedef files unable to resolve each others. Any idea how to solve it?

jensbodal:
The https://github.com/zerkalica/zerollup/tree/master/packages/ts-transform-paths transformer should handle type files properly from what their docs say

sujingclg:
> Copy/pasta snippet incoming for people in need:
> 
> ```js
> // babel.config.js
> module.exports = {
>   // ...
>   plugins: [
>     // ... other plugins
>     [
>       'babel-plugin-module-resolver',
>       {
>         root: './',
>         alias: require('./tsconfig.json').compilerOptions.paths,
>       },
>     ],
>   ],
> };
> ```
> 
> This works wonderfully in combination with `TsconfigPathsPlugin` (Webpack plugin). I declare my aliases in `tsconfig.json` and they work in Webpack, Storybook, Babel (for tests), while it also keeps VSCode happy. Hope this helps someone.

It seems not working at "babel-plugin-module-resolver": "^4.1.0",  and can be fixed by change babel.config.js as follow.
```js
module.exports = {
  plugins: [
    [
      'module-resolver',
      {
        alias: {
          "@": "./src"
        }
      },
    ],
  ],
}

```

sujingclg:
> > Copy/pasta snippet incoming for people in need:
> > ```js
> > // babel.config.js
> > module.exports = {
> >   // ...
> >   plugins: [
> >     // ... other plugins
> >     [
> >       'babel-plugin-module-resolver',
> >       {
> >         root: './',
> >         alias: require('./tsconfig.json').compilerOptions.paths,
> >       },
> >     ],
> >   ],
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This works wonderfully in combination with `TsconfigPathsPlugin` (Webpack plugin). I declare my aliases in `tsconfig.json` and they work in Webpack, Storybook, Babel (for tests), while it also keeps VSCode happy. Hope this helps someone.
> 
> It seems not working at "babel-plugin-module-resolver": "^4.1.0", and can be fixed by changing babel.config.js as follow.
> 
> ```js
> module.exports = {
>   plugins: [
>     [
>       'module-resolver',
>       {
>         alias: {
>           "@": "./src"
>         }
>       },
>     ],
>   ],
> }
> ```

the key of it is change alias value from array to stirng.

sujingclg:
> > > Copy/pasta snippet incoming for people in need:
> > > ```js
> > > // babel.config.js
> > > module.exports = {
> > >   // ...
> > >   plugins: [
> > >     // ... other plugins
> > >     [
> > >       'babel-plugin-module-resolver',
> > >       {
> > >         root: './',
> > >         alias: require('./tsconfig.json').compilerOptions.paths,
> > >       },
> > >     ],
> > >   ],
> > > };
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This works wonderfully in combination with `TsconfigPathsPlugin` (Webpack plugin). I declare my aliases in `tsconfig.json` and they work in Webpack, Storybook, Babel (for tests), while it also keeps VSCode happy. Hope this helps someone.
> > 
> > 
> > It seems not working at "babel-plugin-module-resolver": "^4.1.0", and can be fixed by changing babel.config.js as follow.
> > ```js
> > module.exports = {
> >   plugins: [
> >     [
> >       'module-resolver',
> >       {
> >         alias: {
> >           "@": "./src"
> >         }
> >       },
> >     ],
> >   ],
> > }
> > ```
> 
> the key of resolving this problem is changing alias value from array to string.

PR #377, opened 2019-12-18
(deps/lint): upgrade @typescript-eslint to support ?. and ??

agilgur5:
- the babel proposal plugins were added, along with tests, but those
  tests didn't pass lint because of missing optional chaining and
  nulllish coalescing support in the linter/parser

Follow-up for #370


jaredpalmer:
Time for a new version?

sw-yx:
probably worth a minor

PR #378, opened 2019-12-18
(ci): add a lint job so PRs will require passing lint

agilgur5:
- a few recent PRs got in without passing lint, this breaks all future
  PRs/branches/commits because pre-commit will fail when it runs lint

Turns out `lint` was never running in CI! Which explains why #370 and #366 passed (and my confusion in #373 )
This should make formatting fixes like #372 and #377 unnecessary in the future if only PRs that pass lint are merged.

I limited this to run on Node v10 and `ubuntu-latest` -- not sure if a matrix is necessary for self-linting. The existing test suite already has tests for `tsdx lint` and that runs over the matrix ofc.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -3,6 +3,24 @@ name: Node CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    name: Lint on node 10.x and ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 10.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+
+      - name: Install deps and build
+        run: yarn install --frozen-lockfile
+
+      - name: Lint codebase
+        run: yarn lint
+
   test:
     runs-on: ${{ matrix.os }}
 


agilgur5:
Ok, this is a bit weird that it passes lint prior to #377 being merged...

The error I was getting was:
```
yarn run v1.19.1
$ yarn build && yarn lint:post-build
$ tsc -p tsconfig.json
$ node dist/index.js lint src test --ignore-pattern 'test/tests/lint'
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.7.0

YOUR TYPESCRIPT VERSION: 3.7.3

Please only submit bug reports when using the officially supported version.

=============

~/Desktop/GitHub/tsdx/test/fixtures/build-default/src/syntax/nullish-coalescing.ts
  5:24  error  Parsing error: Expression expected  prettier/prettier

~/Desktop/GitHub/tsdx/test/fixtures/build-default/src/syntax/optional-chaining.ts
  3:52  error  Parsing error: Expression expected  prettier/prettier

✖ 2 problems (2 errors, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

sw-yx:
great calls on every respect.

agilgur5:
Ah, I see, #370 added a commit later which bumped `prettier` which bumped its dependency on `@typescript-eslint/typescript-estree`, so it then started to pass lint when deps were upgraded. 

#377 was still needed to get rid of the big warning and to support optional chaining / nullish coalescing in `eslint` proper (not just in `prettier`), but not to fix the specific `prettier` errors I wrote of above.

PR #385, opened 2019-12-20
[docs] howto turn off watch mode

lookfirst:
Taken from here: https://github.com/jaredpalmer/tsdx/issues/319#issuecomment-566659469

Issue #319, opened 2019-11-10
test command not running in watch mode


tylerthehaas:
### Current Behavior

`yarn test` isn't running in interactive mode

### Expected behavior

to have tests run in interactive/watch mode.

### Suggested solution(s)

run jest in watch mode when tsdx test is run.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           |  0.11.0
| TypeScript       | 3.7.2
| Browser          | N/A
| npm/Yarn         | Yarn
| Node             | 12.8.1
| Operating System | MacOS Mojave version 10.14.4

sw-yx:
i think this is good to have by default. would take a pr for that

tylerthehaas:
I wonder if we can use bash's `$@` as mentioned in this post:

https://stackoverflow.com/a/3816747/6107169

sw-yx:
eh i think we have better ways to do that programmatically. but feel free to open up a pr and do what works!

tylerthehaas:
what do you suggest?

sw-yx:
not sure but check this out: https://github.com/jaredpalmer/tsdx/blob/c5f65bcdee60377886d721bc642cdb22c9f73db0/src/index.ts#L559-L581 and see how you'd change

arthurdenner:
@tylerthehaas, sorry if I misunderstood, but isn't `yarn test --watch` what you need?

jaredpalmer:
Yeah all we should do is key off of whether CI=true IMHO

sw-yx:
that works. nice 1-2 liner.

MartijnHols:
Previously I could use `yarn test` for a single run and `yarn test --watch` for watching. Is it no longer possible to make it run just once outside of CI now?

ambroseus:
@MartijnHols  I use env var CI=true
`"test": "CI=true tsdx test --color"`
 `"test:coverage": "CI=true tsdx test --color --coverage"`

sw-yx:
or ctrl + C to kill :)

sw-yx:
@MartijnHols @ambroseus we are reverting the "watch mode by default" behavior we added a month ago - sorry for the flipflop. more info https://github.com/jaredpalmer/tsdx/issues/408

davidstellini:
Hi @sw-yx Sorry if this comes off as bad, but - why?

[...]

PR #395, opened 2019-12-27
(docs): improve test watch mode docs

agilgur5:
- it was not clearly stated that CI=true was what needed to be added,
  so explicitly state that
- similarly, --color was an unnecessary option to pass and may have
  caused confusion, potentially making users believe that it was
  necessary in order to disable watch mode
  - also, --color does not seem to be a valid option of Jest, but
    [--colors](https://jestjs.io/docs/en/cli#--colors) (note the s) is used to force highlighting in non-TTY

- also, properly format package.json example as JSON, including the
  "scripts" section
- and add a test:watch script for context, making more explicit the
  difference between non-watch and watch mode

Follow-up to #385 , related to #319, #366 . Though this documentation does make me wonder whether it makes sense to have `--watch` as a default. It's much more intuitive to pass `--watch` than to prefix with `CI=true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,11 +450,16 @@ Examples
 
 This runs Jest v24.x in watch mode. See [https://jestjs.io](https://jestjs.io) for options. If you are using the React template, jest uses the flag `--env=jsdom` by default.
 
-If you would like to disable watch mode, this is one way to do it in your `package.json`:
+If you would like to disable watch mode, you can set the environment variable `CI=true`. For instance, you could set up your `package.json` `scripts` like:
 
-```
-"test": "CI=true tsdx test --color"
-"test:coverage": "CI=true tsdx test --color --coverage"
+```json
+{
+  "scripts": {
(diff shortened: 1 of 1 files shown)


sw-yx:
feel free to propose `--watch` in a separate issue. altho i think if u followed those issues, jared had a strong opinion that watch should be default

agilgur5:
> altho i think if u followed those issues, jared had a strong opinion that watch should be default

Unless I'm missing something, I think this is the only comment Jared made on it: https://github.com/jaredpalmer/tsdx/issues/319#issuecomment-564143894 , which didn't seem strong?

But yea would be a separate issue, perhaps that should be voted on by users

sw-yx:
its not really up to a vote, this is Jared's project. but do what you think is right :)

sw-yx:
heads up - we have decided to undo this move. will not run in watch mode by default. sorry for the flipflop. https://github.com/jaredpalmer/tsdx/issues/408

PR #397, opened 2019-12-27
(docs/fix): fix ToC changes / bugs introduced in #307

agilgur5:
- `__DEV__` link stopped working
- Contributors link didn't work

- backtick formatting was removed

- this was all re-generated by using `npx doctoc README.md`
  - doctoc is specified to be used in the ToC comments

Noted in https://github.com/jaredpalmer/tsdx/pull/307#discussion_r342945347 and https://github.com/jaredpalmer/tsdx/pull/307#discussion_r359907387


agilgur5:
No clue why windows tests seemed to randomly fail on this PR, especially when I made like 5 PRs in a row, and this is the only one that has failures (in code that wasn't even changed)

sw-yx:
not sure why this happens either. here's the failing test
image

so its some sort of filesystem permission issue that is also a race condition? need to investigate and do some trial and error. probably an easy fix.

PR #398, opened 2019-12-27
(format): alphabetize package.json deps

agilgur5:
- this is done automatically when adding a dependency via Yarn / NPM
  - (but not when doing so manually)
  - format as separate commit so it won't need to be repeated or
    randomly introduced in others


agilgur5:
Seems like this came from https://github.com/jaredpalmer/tsdx/pull/130 . Maybe should run `yarn` or something on pre-commit to automatically alphabetize as well? That could cause some confusion if deps are updated during pre-commit though...

sw-yx:
imo its not a big deal, but thanks for doing it

PR #399, opened 2019-12-27
(docs): run doctoc on pre-commit

agilgur5:
- and install it as a devDep

- doctoc is used in the README ToC and listed in comment, but wasn't
  a devDep or used in any internal scripts

Follow-up to #397 . Please merge after that one and after #398 , which I split out from this PR.

The big thing that's not so helpful about running `doctoc` on pre-commit as that it won't abort the commit if changes were made. It just adds new changes, but those haven't been staged yet so won't be included in the commit (unless you amend them in). Not sure what a better method may be that isn't overly complex 😕

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "@types/ps-tree": "^1.1.0",
     "@types/rollup-plugin-json": "^3.0.2",
     "@types/rollup-plugin-sourcemaps": "^0.4.2",
+    "doctoc": "^1.4.0",
     "husky": "^3.0.9",
     "pretty-quick": "^2.0.0",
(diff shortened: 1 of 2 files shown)


sw-yx:
how bout doing this via github action?

agilgur5:
ah so looks like this is a known issue of `doctoc`: https://github.com/thlorenz/doctoc/pull/169, https://github.com/thlorenz/doctoc/issues/161 . Could use that PR branch and run it with the code `--stdout` code listed which would be ideal

But, in general, `doctoc` isn't really maintained, so it might just make sense to look for an alternative instead of hacking around with it.

Issue #389, opened 2019-12-21
Improve internal configs generation 


ambroseus:
### Current Behavior

As mentioned in README: _TSDX is a zero-config CLI that helps you develop, test, and publish modern TypeScript packages with ease_

but there are a lot of issues that need to fix / patch / extend configs for ts / babel / jest / rollup / rollup plugins / plugins for rollup plugins etc. see https://github.com/jaredpalmer/tsdx/issues/379

### Desired Behavior

ability to easy deal with all kind of configs

### Suggested Solution

need to discuss

### Describe alternatives you've considered

possible simplest solution (as CRA): `tsdx eject`

sw-yx:
i'm open to this idea (i prefer shadowing rather than ejecting) but also the user can modify the rollup config via tsdx.config.js. What do you think is the tipping point between modifying via config and total ejection?

ambroseus:
@sw-yx  I also prefer to deal with merge approach instead of full ejecting configs but..
 periodically during issue investigation I want to see full configs to understand what is passing to exact tool (tsc, babel, jest, rollup, rpt2 etc.)

it will be good to extend `tsdx.config.js` to deal with other configs

sw-yx:
alright. im not as strongly opinionated on seeing full configs but this is really something Jared will have to decide.

jaredpalmer:
So ejection is not gonna happen. My goal with TSDX is to eviscerate boilerplate from typescript by providing a single abstraction with solid defaults that’s infinite customizable.

To make this strategy a reality, we always need to ensure that that the correct escape hatches exist. That is mission critical to the project’s continued success. I believe we should address each specific use case and decide to adapt or document a solution. 

Lastly, and maybe we should also document this, I have found that patch-package is a wayyyyy better approach to modifying create-react-app instead of ejecting. Most of the time I only need to tweak tweak 10-15 lines of webpack, and am not interested in foregoing the rest of the updates. If tsdx.config.js isn’t sufficient, patch-package is my suggested final escape hatch.

As for understanding the various configs, I agree it could perhaps be restructured internally to use deep merge instead of just functions. However, I’m not 100% sure if it’s going to make that big of difference. Since we are outputting up to 3 formats for n entries, we are going to have functions regardless. We could technically augment tsdx config to have a deep merge mode like storybook does. Personally, I don’t care for this, because I like being able to log everything. It also gets weird with dev/prod and different formats....something that storybook doesn’t need to think about. 

[...]

Issue #379, opened 2019-12-18
HOWTOs


ambroseus:
# Collection of useful HOWTOs

ambroseus:
### How to run tests in stream (non-watch) mode
`package.json`
```json
{
  "scripts":
    {
       "test": "CI=true tsdx test --color"
    }
}
```

ambroseus:
### How to collect test coverage
`package.json`
```json
{
  "scripts":
    {
       "test:coverage": "CI=true tsdx test --color --coverage"
    },
    "jest":
    {
        "collectCoverageFrom": [
            "<rootDir>/src/**/*.{ts,tsx}"
         ],
         "coverageReporters": [
            "lcov",
            "text-summary"
         ]
    }
}
```

ambroseus:
### How to correctly transpile spread on iterables (Set, Map) https://github.com/jaredpalmer/tsdx/issues/376
`tsconfig.json`
```json
  "downlevelIteration": true
```
`package.json`
```json
{
  "scripts":
    {
       "build": "tsdx build --target node"
    }
}
```

ambroseus:
### How to cange output folder name from "dist" to "build" https://github.com/jaredpalmer/tsdx/issues/351
`tsdx.config.js`
```js
module.exports = {
  rollup(config, options) {
    config.output.file = config.output.file.replace('dist', 'build');

    return config;
  },
};
```

ambroseus:
### How to 'play nicely' with tsdx + VS Code + eslint https://github.com/jaredpalmer/tsdx/issues/354
https://github.com/jaredpalmer/tsdx#npm-run-lint-or-yarn-lint
run `npm run lint --write-file` or `yarn lint --write-file` to generate `.eslintrc.js` file

sw-yx:
nice. worth adding to docs when you think its ready.

ambroseus:
@sw-yx  yep) can you plz add label `documentation` for this issue?

ambroseus:
### How to override `rollup-plugin-typescript2` configuration https://github.com/jaredpalmer/tsdx/issues/294
`tsdx.config.js`
```js
import rpts2 from "rollup-plugin-typescript2";

module.exports = {
  rollup(config, options) {
    config.plugins = config.plugins.map(plugin => {
      if (plugin && plugin.name === "rpt2") {
        return rpts2({
          // properties that I added for demonstration purposes
          clean: true,
          objectHashIgnoreUnknownHack: true,
[...]

PR #400, opened 2019-12-28
(docs): add warning to tsdx.config.js usage

agilgur5:
- similar to the ones that exist in [react-app-rewired](https://github.com/timarney/react-app-rewired/#rewire-your-app-) and [customize-cra](https://github.com/arackaf/customize-cra#-warning)
  as, like CRA, changing TSDX internals is a fragile path

Related to https://github.com/jaredpalmer/tsdx/issues/389#issuecomment-568866380 , #379 . This escape hatch is being used quite a bit by TSDX users as seen in various issues, but I'm not sure if all are aware of the risks that inherently entails (as I detailed in https://github.com/jaredpalmer/tsdx/issues/389#issuecomment-568866380). Those risks should be more explicit, as this may catch users off-guard, especially during updates, as even a patch update could break usage inside `tsdx.config.js`.

Plugins could at least pin a peer dependency and update when their own tests pass for newer versions, but ad-hoc code is more brittle than that. Notably, [Razzle does support plugins](https://github.com/jaredpalmer/razzle#plugins), and those are listed as the first option before modifying `webpack` config directly (`rollup` for TSDX)

Probably makes sense to leave this to @jaredpalmer as [Razzle doesn't quite have the same language](https://github.com/jaredpalmer/razzle#extending-webpack).

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,9 @@ _TODO: Simple guide to host error codes to be completed_
 
 ### Rollup
 
+> **❗⚠️❗ Warning**: <br>
+> These modifications will override the default behavior and configuration of TSDX. As such they can invalidate internal guarantees and assumptions. These types of changes can break internal behavior and can be very fragile against updates. Use with discretion!
+
 TSDX uses Rollup under the hood. The defaults are solid for most packages (Formik uses the defaults!). However, if you do wish to alter the rollup configuration, you can do so by creating a file called `tsdx.config.js` at the root of your project like so:
 
 ```js


agilgur5:
Oook, here's another example of tests randomly failing on existing code in a PR that doesn't change any source code (only docs). Seems related to errors I was getting in https://github.com/jaredpalmer/tsdx/pull/397#issuecomment-569366418 and  https://github.com/jaredpalmer/tsdx/pull/367#issuecomment-565290969 . It didn't seem like they were race conditions, but now I'm not sure if this is just filesystem unreliability either given how frequently it's occurred. Not sure what to do about it 😕

sw-yx:
here's the CI failure in case people cant see 
image

README.md
```diff
@@ -298,6 +298,9 @@ _TODO: Simple guide to host error codes to be completed_
 
 ### Rollup
 
+> **❗⚠️❗ Warning**: <br>
+> These modifications will override the default behavior and configuration of `tsdx`. As such they can invalidate internal guarantees and assumptions. These types of changes can break internal behavior and can be very fragile against updates. Use with discretion!
```
jaredpalmer: TSDX should be capitalized. We should be consistent.

agilgur5: Ah good catch, I mix these up often. Fixed

PR #401, opened 2019-12-28
(types/refactor): be more specific than 'any' in build code

agilgur5:
- add types for all options and use those types throughout the
  build/watch code
- fix some incorrect types in a few places that became more obvious
  or errors once types were added
- fix default target from 'web' (not an option) to 'browser'

- extract createBuildConfigs into a separate file as it's fairly long
  and has some relatively complex code
  - and refactor it a bit to make it a bit easier to read as well as
    easier to type (and type cast)

Follow-up to #371 , which was extracted out of this before I finished. I originally started this after #367 as the added types would give a lot of extra confidence around options, which were previously `any` everywhere. This is also going to merge conflict quite hard against #367 .

## Review Notes

There are two type casts I couldn't seem to get around here (guess it hits the limits of TypeScript's inference?), but this is still much better than all the `any`s, and now new options must be typed and then used properly.

## Future Work

`createAllFormats` (and possibly some of the code around it) is one of those things that could probably be split into a separate rollup plugin.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -25,7 +25,7 @@ import shell from 'shelljs';
 import ora from 'ora';
 import { paths } from './constants';
 import * as Messages from './messages';
-import { createRollupConfig } from './createRollupConfig';
+import { createBuildConfigs } from './createBuildConfigs';
 import { createJestConfig } from './createJestConfig';
 import { createEslintConfig } from './createEslintConfig';
 import { resolveApp, safePackageName, clearConsole } from './utils';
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -502,7 +439,7 @@ async function normalizeOpts(opts: any) {
         return 'esm';
       }
       return format;
-    }),
+    }) as [ModuleFormat, ...ModuleFormat[]],
```
sw-yx: what is this?!

agilgur5: haha, I had to [look up how to do this](https://stackoverflow.com/q/56006111/3431180).

This type means that there is "at least one" `ModuleFormat` in the array vs. plain `ModuleFormat[]` could mean no elements

sw-yx: loll interesting.

jaredpalmer:
Resolve conflicts and then lgtm

jaredpalmer:
will merge when the tests pass

PR #402, opened 2019-12-28
(refactor): use path constants instead of custom resolveApp

agilgur5:
- and use ensureDistFolder since that exists now
- and use async readJSON since most were converted before

Merges stuff from #130, #327 , #291 , #142 as they were all merged together at different times but changed some conventions in the codebase or DRY'd things up.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -44,7 +44,7 @@ const prog = sade('tsdx');
 let appPackageJson: PackageJson;
 
 try {
-  appPackageJson = fs.readJSONSync(resolveApp('package.json'));
+  appPackageJson = fs.readJSONSync(paths.appPackageJson);
 } catch (e) {}
 
 // check for custom tsdx.config.js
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -464,7 +464,7 @@ prog
     const logger = await createProgressEstimator();
     if (opts.format.includes('cjs')) {
       try {
-        await util.promisify(mkdirp)(resolveApp('./dist'));
+        await ensureDistFolder();
```
agilgur5: I just realized there's an `ensureDistFolder()` call literally 4 lines before this... So this call is actually redundant (and therefore the whole try-catch too); not sure why it was needed for #130 .

Also, using `outputFile` instead of `writeFile` in `writeCjsEntryFile` (like I did in #367 ) can replace the need for `ensureDistFolder`

agilgur5: Followed up with this change in #406

PR #406, opened 2019-12-31
(refactor): replace ensureDistFolder with fs.outputFile

agilgur5:
- fs.outputFile will create any directories that don't already exist

- also remove extraneous ensureDistFolder w/ try-catch in build command

Follow-up to https://github.com/jaredpalmer/tsdx/pull/402#discussion_r361822746 . Was going to add this to that PR but it got merged in before I could


sw-yx:
thanks for the great PR! @all-contributors please add @agilgur5 for maintenance, docs, code and tests

Issue #345, opened 2019-11-22
Problem with target = CommonJS in tsconfig


wiput1999:
### Current Behavior

I'm about to use tsdx to build library for Node.js but it still need to add .default after require. So I'm tried to change to `export = ` and tsconfig target to CommonJS then 

```
Module resolves to 'CommonJS'. This is incompatible with rollup, please use 'module: "ES2015"' or 'module: "ESNext"'.
```

is it possible to solve this issue?

### Who does this impact? Who is this for?

Who using `export = ` and target as commonjs

sw-yx:
not sure. if you can investigate and propose a fix, please let me know

wiput1999:
In my opinion but not sure I think it’s problem from rollup. Maybe next week if I’ve time I’ll try looking about this

hardfist:
I think the problem is that tsdx use 'named' exports rather than  'auto' exports(which is rollup's default behavior ), which I think auto is more appropriate for the ugly `export default` things,
the difference between auto mode and named mode is different when treating `export default`
```
export default 42;
```
auto mode generates
```
'use strict';

var main = 10;

module.exports = main;
```
named mode generates
```
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

var main = 10;

exports.default = main;
```
and auto mode works perfectly with node && browser
```
const lib = require('lib') // ok
```
```
import lib from 'lib' // ok when set esmoduleInterop: true
```
@sw-yx @jaredpalmer

SokratisVidros:
I also came across the same issue while building a Node library. Is there a suggested `tsconfig.json` setup that eliminates the need of `.default` in CommonJS imports?

PR #117, opened 2019-05-29
Better compilation via Babel

jaredpalmer:
Use babel to do better compilation #102 . Fix #109 

We can/should discuss overrides and customization for both babel and rollup and stuff in another issue.

PR #415, opened 2020-01-04
(fix): revert #130's breaking change of tsconfig options

### LINKED ISSUE #413: Breaking change with tsdx 0.12.1 with Tailwind babel macro
### Current Behavior

I'm using `tailwind.macro` in a React library with Storybook built with tsdx. On tsdx 0.12.0, everything was working as expected as babel macros are supported out of the box.
However, when updating to 0.12.1, my builds stopped working properly with tailwind.macro.

I created a repository with the two versions and a simple component including styled-components and tailwind.macro: https://github.com/fknop/tsdx-tailwind.macro-bug.
I also included the two dist folders from the two versions:
0.12.0: https://github.com/fknop/tsdx-tailwind.macro-bug/blob/master/tsdx-0.12.0/library/dist/library.esm.js
0.12.1: https://github.com/fknop/tsdx-tailwind.macro-bug/blob/master/tsdx-0.12.1/library/dist/library.esm.js

As we can see, `tw` is not imported anywhere in the 0.12.1 version. This makes any code importing this build failing with the function `tw` being undefined.

Tested on both MacOS and Windows 10.

agilgur5:
- not sure why this change was made in that PR as it was unrelated to
  the rest, but it broke some things related to Babel usage and maybe
  other things
  - e.g. babel macros were not included per #413
  - and it used `tslib` instead of adding babel helper functions

- add a comment of why it seems like the target is overridden to esnext
  - though I am not sure of the original reasoning, as it has existed
    since very early versions of TSDX, see #130 comments

Fixes #413 

Whether TS should transpile to target or Babel should is a good question to raise, but for now this just reverts the breaking change.
Should probably document in the README that `tsconfig` target is ignored and that users should use Babel's `preset-env` to configure the target instead.
Should also probably add more tests around this as bugs / breaking changes like this in TSDX's build output can have ripple-effects on libraries using it (vs. the other commands don't affect output, so the impact is limited to just library authors as opposed to library users).

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -144,7 +144,12 @@ export async function createRollupConfig(
             sourceMap: true,
             declaration: true,
             jsx: 'react',
-            target: 'es5',
+          },
+        },
+        tsconfigOverride: {
+          compilerOptions: {
+            // TS -> esnext, then leave the rest to babel-preset-env
+            target: 'esnext',
           },
         },
         check: opts.transpileOnly === false,


agilgur5:
Another example of `.d.ts` missing in Ubuntu CI tests here -- see growing list in #403

jaredpalmer:
I forgot why, but this is v important.

agilgur5:
Found #117, #102, and #111 that are all related to this.
Also possibly #345, but not sure about that one.

Per, #117, sounds like the comment I left in the code, `// TS -> esnext, then leave the rest to babel-preset-env`, describes the original rationale accurately.

PR #416, opened 2020-01-05
Fix onFailure example

karlhorky:
Hey, thanks for `tsdx` and all of the latest changes! Spotted what seems to be a typo in one of the newer examples. Here's a quick PR to fix.


sw-yx:
thanks for the great PR! @all-contributors please add @karlhorky for docs

karlhorky:
Thanks @sw-yx !

PR #417, opened 2020-01-06
[Docs]: Added a reference to contributing guidelines as part of the docs

jamesgeorge007:
It is a good idea to have a reference to the `contributing guidelines` as part of the docs which the users may miss out if otherwise.


sw-yx:
thanks for the great PR! @all-contributors please add @jamesgeorge007 for docs

Issue #408, opened 2020-01-01
Expose a way to run test in non-interactive mode without defining CI


etienne-dldc:
### Current Behavior

- `yarn test` start jest in interactive mode
- Setting `CI=true` start jest in non-interactive mode but without colors because most tools only check wether or not the `CI` environment variables exist an not it's value:
https://github.com/watson/ci-info/blob/master/index.js#L53
Because of this you loose some functionalities, like colors when using tools like `np` or `lerna`.

### Desired Behavior

Being able to run `tsdx test` in non-interactive way without setting the `CI` variable.

### Suggested Solution

Using a different environment variable name to disable "auto watch mode":

```
TSDX_AUTO_WATCH=false tsdx test
```

### Who does this impact? Who is this for?

This impact everyone who was using `tsdx test` in non-interactive mode outside of a CI.
It would make the migration from `0.11` to `0.12` easier.

sw-yx:
ohh.. good point 😬

@jaredpalmer we may have to revert this and just use a watch flag.

jaredpalmer:
Okay let’s revert then

PR #421, opened 2020-01-09
Revert "default jest watch unless in CI" PR #366

sw-yx:
Revert "default jest watch unless in CI" PR #366

we decided to revert this behavior in https://github.com/jaredpalmer/tsdx/issues/408

closes https://github.com/jaredpalmer/tsdx/issues/408

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,16 +452,14 @@ Examples
 
 ### `tsdx test`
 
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Why are tests failing?

sw-yx:
we've been having problems with nondeterministic test failures the past few weeks. like this one https://github.com/jaredpalmer/tsdx/pull/403. i'm not sure what the cause is and havent had time to investigate

Issue #424, opened 2020-01-10
Deprecate Node 8 support?


sw-yx:
### Current Behavior

Node 8 is EOL. things are starting to have hard Node 10 dependencies: https://github.com/jaredpalmer/tsdx/pull/422

Should we move our test matrix up and document that Node 8 is not supported?

### Desired Behavior

as above

### Suggested Solution

move our test matrix up and document that Node 8 is not supported

jaredpalmer:
Yep 👍🏻

PR #426, opened 2020-01-11
Deprecate Node 8 support

sw-yx:
closes https://github.com/jaredpalmer/tsdx/issues/424#issuecomment-573202098

also begins Node 13 in preparation for for Node 14 release in April (goes LTS in Oct this year)

recheck https://github.com/jaredpalmer/tsdx/pull/422 once merged


sw-yx:
ubuntu is failing again. i dont have any ideas why at all; this is looking rather persistent.

PR #429, opened 2020-01-13
Run tests in series with --runInBand

jaredpalmer:
Adding `--runInBand` which runs the test in a series instead of in parallel IIRC. This is what I believe was causing the race conditions

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,6 +1,8 @@
 name: Node CI
 
-on: [push, pull_request]
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


agilgur5:
Related to #403 which was a similar attempt.

Root cause of most CI issues (missing `*.d.ts` type declaration files) was fixed in #504 so this is _probably_ not necessary anymore unless the Windows issue mentioned in #403 (`EPERM` on creating directories) was fixed by this.
I'm not sure as the errors were different but the missing declaration one happened _very_ frequently and `EPERM` happened occasionally. `EPERM` hasn't been seen in a while.

Issue #432, opened 2020-01-16
Document Node version support / add a warning to tsdx create


lpolito:
### Current Behavior

If you're on Node 8 you get errors when trying to run `tsdx create`, but the reason for the failures are unclear.

```
nvm use 8.11.3

npx tsdx create bar

Failed to install dependencies
Error: Command failed with exit code 1: yarn add @types/jest @types/react @types/react-dom husky react react-dom tsdx tslib typescript --dev
```

### Desired Behavior

When I run `tsdx create` it should tell me that my current Node version is not compatible.

### Suggested Solution

Add a deprecation warning for unsupported Node versions when running `tsdx create`, as well as add Node versions supported in the `README.md`, and probably add a `engines` attribute to the package.json describing the same limitations.

### Who does this impact? Who is this for?

All users.

### Additional context

I only knew this by seeing the recent releases including this PR https://github.com/jaredpalmer/tsdx/pull/426

sw-yx:
yup, sure thing. we can add an engines field https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json and use something like `check-version.js` in the SO reply or something like https://www.npmjs.com/package/check-node-version

lpolito:
Attempting to resolve this myself ✋

agilgur5:
Mentioned this in https://github.com/formium/tsdx/pull/507#issuecomment-683453786 :

> [this is exclusively] an issue with `tsdx create` because it doesn't pin anything (including `husky`) and so it gets the most up to date versions of things, which can be breaking or have incompatible versions with TSDX core

TSDX itself does not yet require Node 10+, I've specifically batched [various dependencies's upgrades to Node 10](https://github.com/formium/tsdx/labels/Node%2010%2B) together to go into [v0.14.0](https://github.com/formium/tsdx/milestones/2), which is a breaking change

sw-yx:
hmmm. whats the logic for still supporting node <10?

agilgur5:
@sw-yx it should be formally dropped in a breaking change (same as all the deps that dropped it did), which is what v0.14.0 is. It was not officially dropped before and is actually still supported.

sw-yx:
ahh, makes perfect sense 👍🏽

PR #433, opened 2020-01-16
(fix) Enforce Node version for tsdx create, update documentation to express requirement.

lpolito:
Fixes https://github.com/jaredpalmer/tsdx/issues/432

There was a few ways I could have approached this, so let me know if there's something more ideal.

To test:
1. Build. 
2. Change Node version to < 10 - `nvm use 8.11.3`
3. Run `node /dist/index.js create foo`, select any template – observe error
```
Unsupported Node version! Your current Node version (v8.11.3) does not satisfy the requirement of Node >=10.
```
4. Change Node version to >=10 - `nvm use 10.17.0`
5. Run `node /dist/index.js create bar`, select any template – observe success.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,9 @@ cd mylib
 yarn start
 ```
 
+_Requires Node `>= 10`._
+
+
 That's it. You don't need to worry about setting up Typescript or Rollup or Jest or other plumbing. Just start editing `src/index.ts` and go!
 
 Below is a list of commands you will probably find useful:
--- a/package.json
+++ b/package.json
@@ -116,11 +116,13 @@
     "@types/react": "^16.9.11",
     "@types/rollup-plugin-json": "^3.0.2",
     "@types/rollup-plugin-sourcemaps": "^0.4.2",
+    "@types/semver": "^6.2.0",
     "doctoc": "^1.4.0",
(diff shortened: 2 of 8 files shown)


lpolito:
Turns out there's a way to [specify a version of node when running npx](https://www.npmjs.com/package/npx#specify-a-node-version-to-run-npm-scripts-or-anything-else)

So the command would be `npx -p node@10 npx tsdx create {name}`.
It's a little weird but it seems to work.

Maybe that should just go in the docs?

sw-yx:
TIL about `npx -p node@10`... but how does that even work if you dont have that version installed!?

lpolito:
> TIL about `npx -p node@10`... but how does that even work if you dont have that version installed!?

Wasn't sure if this PR was waiting on the answer to this 😶 

It works the same way as `npx` normally does (magic), except the package you're specifying is the particular node version. Then you just give it the command you want to run while running that node version. 🤷‍♂

sw-yx:
haha no i was just leaving room for the other maintainers to comment if needed. i'll go ahead n merge

sw-yx:
thanks for the great PR! @all-contributors please add @lpolito for code and docs

PR #436, opened 2020-01-20
Respect custom tsconfig path for esModuleInterop

tricoder42:
When custom `tsconfig` is used (e.g. `tsdx build --tsconfig tsconfig.build.json`), it is ignored in `createRollupConfig`.

For example, this causes bug in reach-ui build https://github.com/reach/reach-ui/issues/434. They use `tsconfig.build.json`, which is being ignore and the build files are missing `Object.defineProperty(exports, '__esModule', { value: true });`.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -45,7 +45,7 @@ export async function createRollupConfig(
 
   let tsconfigJSON;
   try {
-    tsconfigJSON = await fs.readJSON(paths.tsconfigJson);
+    tsconfigJSON = await fs.readJSON(opts.tsconfig || paths.tsconfigJson);
   } catch (e) {}
 
   return {


tricoder42:
@agilgur5 Both commit and PR were renamed. Thanks for the review! 👍

chancestrickland:
Hey @sw-yx / @agilgur5 anything I can do to help get this merged, or perhaps maybe you know of a workaround in the mean time? A few of our packages in Reach UI use default exports and are broken as it stands. Many thanks for your work on this!

agilgur5:
@chancestrickland I'm an active contributor but don't have merge privileges -- would merge if I could. If @sw-yx didn't merge I assume he's waiting for Jared's approval? Jared also handles releases.

I think you should be able to workaround it by overriding the rollup config with `tsdx.config.js`. Would give an example myself but don't have time right now (traveling till next week), there are some in #379 and other issues.

chancestrickland:
@agilgur5 Excellent, thanks for the link and super-quick reply. I'll see if I can go that route for now. 👍

sw-yx:
yea usually for anything config-centric i want to make sure jared approves everything personally since thats the core of it. he also handles the deploy. so leaving to @jaredpalmer.

dont forget you can also do local forks while u wait for pr's to get merged, with http://npm.im/patch-package

chancestrickland:
I genuinely did not know about `patch-package`. This may be the best day of my life.

agilgur5:
> yea usually for anything config-centric i want to make sure jared approves everything personally since thats the core of tsdx. i dont want to make decisions that he then has to reverse.

For sure. I thought one-liner bugfixes like this might fall into a different category since they're not really changing any behavior tho (I feel like I've made a few PRs like that but idr)

jaredpalmer:
Sorry for delay. Will get this out now

agilgur5:
@all-contributors please add @tricoder42 for bug, code

agilgur5:
Hmmm, so after writing #526 , I wonder if this will work properly without using `resolveApp`. Will need to add more tests around this.

Issue #441, opened 2020-01-23
Custom templates


MerrickGit:
tsdx is't silver bullet, but it can be.

### Current Behavior
May be I just did't find this feature, but tsdx can't create custom templates.

### Suggested Solution
Create package based on remote template.

Example:
`tsdx create your_new_package --remote-template=https://github.com/user/prototype`

Step by step instruction:
1. tsdx download template
2. Find into repo config file.
3. add template to templates and config to configs.
4. do usual work

### Who does this impact? Who is this for?

People are using monorepo.

### Additional context
We create big monorepo(about 300 packages) and we use tsdx, but we spend a lot of time creating/copying the same type of files(configs/structure/libraries) instead of use template.

ambroseus:
+1
https://github.com/jaredpalmer/tsdx/issues/409#issuecomment-570242716

MerrickGit:
@jaredpalmer i see its so demanded features. What do you think about it?

jaredpalmer:
Yeah, that’s fine. Probs can copy logic from create-react-app

MerrickGit:
@jaredpalmer @sw-yx can you see my comment in PR? What do you think?

sw-yx:
i'm fine with it, gonna leave this one up to jared to see if he wants to merge first and iterate or require remote

benjlevesque:
I submitted another proposal for this, with support for remote templates (npm could be added easily I guess). 
It reuses the `--template` option instead of introducing a new one.

See #502.

ambroseus:
:thinking:  what about this one: https://github.com/jondot/hygen

PR #455, opened 2020-01-24
Custom templates

MerrickGit:
#441

Added option `--custom-template`.

usage `tsdx create app_name --custom-template ~/Desktop/templates/my_template`

One important thing is `template.json` inside the template folder.

If this method is satisfies the reviewers, then I will add documentation and checking for `template.json` file.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,6 +47,7 @@ const pkg = require('../package.json');
 const prog = sade('tsdx');
 
 let appPackageJson: PackageJson;
+const customTemplate = 'custom-template';
 
 try {
   appPackageJson = fs.readJSONSync(paths.appPackageJson);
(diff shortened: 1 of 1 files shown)


sw-yx:
hmm it might be more common to host templates on github rather than your filesystem - what do you think

jaredpalmer:
It should be remote, via GitHub

MerrickGit:
I think it will be next feature because we need a logic to work with templates in the file system. I can leave only remote or can add the following PR —remote-template option

MerrickGit:
> It should be remote, via GitHub

jaredpalmer:
Imho all templates should be remote

MerrickGit:
> Imho all templates should be remote

May be via npm, not all companies use github?

kdubb1337:
We're using TSDX pretty heavily on a monorepo, and it causes you to make the same series of changes each time. I think custom templates is a much better option than forking. Can we reopen this PR?

luistak:
@kdubb1337 same problem in here

PR #457, opened 2020-01-25
Add GH Action configs for each template

kylemh:
In action [here](https://github.com/OperationCode/components/pull/7)

Also, it would be nice if we could have a folder for like "goes in all templates"... I would hate to see one of the files get updated in one template, but not the others.


sw-yx:
yeah i hear you. but we dont need to DRY everything while at an early stage. let's figure out what actions people want, and how much they have in common vs how much they vary :)

PR #466, opened 2020-01-29
(remove): redundant/confusing tsconfig target in templates

agilgur5:
- tsconfig target is always overwritten to ESNext by TSDX, and then
  the rest of the transpilation is done by @babel/preset-env
  - having templates with target set gives the false impression that
    target can be freely configured

- also remove from tests as it's unused there

See #415 for more about why the target is overwritten by TSDX


agilgur5:
Oh, probably should remove from the tests' `tsconfig` as well

Issue #443, opened 2020-01-24
Using jsx in .jsx files not achievable


DanweDE:
### Current Behavior

With my `tsconfig.json`'s  `compilerOptions.jsx` set to `"react"`, when using `.jsx` files that contain any jsx markup, on `tsdx build` I get

```
(babel plugin) SyntaxError: /path/src/File.jsx: Unexpected token (3:11)
  1 | import React from 'react';
  2 | export function SomeComponent() {
> 3 |     return <div>Foo</div>;
    |            ^
  4 | }
```

Now, I'd expect this to work out of the box with `.jsx` files but I also tried to instead set the `tsconfig.json`'s  `compilerOptions.jsx: "preserve"` and use `@babel/preset-react` via my .babelrc
```
{
  "plugins": [
    ...
  ],
  "presets": ["@babel/preset-react"]
}
```

Still getting the same error. While the plugins I use seem to work, preset-react seems to not do the job for some reason.

### Expected behavior

`.jsx` files' jsx code should be transpiled according to the tsconfig's `jsx` option or there should be another way to achieve this. 

### Additional context
Using the `.tsx` file extension instead of `.jsx` for the same file, everything works as expected while using `jsx: "react"` but still can't get it to work with `jsx: "preserve"` plus Babel.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.12.3
| TypeScript       | 3.7.3
| Node             | 0.12 
| Operating System | Mac, Linux

DanweDE:
Regarding babel, tsdx's README.md seems highly contradictory to me, stating at one point _"While yes, TSDX does use Babel [...] it does not support custom babel configurations"_ and then _"You can add your own .babelrc to the root of your project and TSDX will merge it with its own babel transforms"_ :thinking:

sw-yx:
good points. we may need to revisit that readme. meanwhile we'll need help digging into why this bug is happening, i dont really understand it myself. maybe something in our rollup config

sw-yx:
https://github.com/jaredpalmer/tsdx/blob/1d8f12311f73548b5bec58d3d44873a91c6f389c/src/createRollupConfig.ts#L159

can you try locally modifying this to add .jsx and see if that fixes it

DanweDE:
Thank you for the quick response @sw-yx! According to [babel docs](https://babeljs.io/docs/en/babel-core#default_extensions):
[...]

PR #467, opened 2020-01-29
(docs/fix): remove outdated language about custom Babel configs

agilgur5:
- as of v0.9, custom Babel configs are merged with TSDX's transforms
- remove language about Babel entirely from comparison docs as Babel
  support does not entail JS support (and at least some JS is already
  supported anyway)

See https://github.com/jaredpalmer/tsdx/issues/443#issuecomment-578141998 and further comments that noticed this outdated language & were confused by it.


agilgur5:
Tests are failing due to https://github.com/jaredpalmer/tsdx/pull/462#issuecomment-579731658

Issue #465, opened 2020-01-28
Enable useTsconfigDeclarationDir when declarationDir


etienne-dldc:
### Current Behavior

The `useTsconfigDeclarationDir` in `typescript2` rollup plugin is set to false.

### Desired Behavior

When `declarationDir` is set in `tsconfig.json` set `useTsconfigDeclarationDir` to `true`.

### Suggested Solution

```js
useTsconfigDeclarationDir: tsconfigJSON && tsconfigJSON.declarationDir
```

### Who does this impact? Who is this for?

This would fix #135  and in general improve experience in Monorepos (making "Go To Definition" work).

### Describe alternatives you've considered

Using `tsdx.config.js` but the config is not easy to change and it's needed in every package of the monorepo.

agilgur5:
Not totally sure how this is related to #184, but in general, that solution makes sense to me as necessary to support `declarationDir` 👍 Can probably use optional chaining (`?.`) syntax now as well.

Want to add a PR + test for this?

etienne-dldc:
Whoops, wrong issue number: #135 
Yeah, I can create a PR 😃

etienne-dldc:
It just appear to me that my "Suggested Solution" is stupid because `rollup-plugin-typescript2` will ignore `useTsconfigDeclarationDir` if there are not `declarationDir` defined so my proposal is the same as setting this to `true` and there would be no way to have a `declarationDir` but still compile enerything in `dist`.

1. This is fine, we just set `useTsconfigDeclarationDir: true`
2. We expose a way to opt-in this option
  a. With an additional CLI argument `--useDeclarationDir`
  b. We read the `typings` field in `package.json` and set `useTsconfigDeclarationDir` if it point to the same folder as `declarationDir` we turn `useTsconfigDeclarationDir` on
  c. ???

Personally, I thing **2.a** is the best options because it's non breaking and avoid unexpected behavior for people who might have a `declarationDir` by accident (copy/paste, default, other build tool...).

agilgur5:
> `rollup-plugin-typescript2` will ignore `useTsconfigDeclarationDir` if there are not `declarationDir` defined so my proposal is the same as setting this to `true`

I'm not certain that this is true. rpts2 defaults `declarationDir` to `cwd`, but if `useTsconfigDeclarationDir` is set, it's `undefined`:
[...]

PR #468, opened 2020-01-29
Enable useTsconfigDeclarationDir if declarationDir

### LINKED ISSUE #135: Incorrect Declaration Map Source
### Current Behavior

I am currently using `tsdx` to run the core package within my monorepo. When I import something from the core package and click through in my editor I end up in the declaration file - not the source file as I would expect.

Source file (`packages/core/src/`):
```ts
import base from './palette.json';

const use = {
  background: {
    primary: base.aliceBlue,
    secondary: base.white,
    tertiary: base.mirage,
  },
  text: {
    primary: base.mineShaft,
  },
};

const colors = { base, use };

export default colors;
```

Declaration file (`packages/core/dist/config/colors.d.ts.map`):
```json
{"version":3,"file":"colors.d.ts","sourceRoot":"","sources":["colors.ts"],"names":[],"mappings":"AAiBA,QAAA,MAAM,MAAM;;;;;;;;;;;;;;;;;;;;;;;;CAAgB,CAAC;AAE7B,eAAe,MAAM,CAAC"}
```

### Expected behavior
When I click-through on a variable imported from a `tsdx` package, I am taken to the source file - rather than the definition file.

Declaration file (`packages/core/dist/config/colors.d.ts.map`):
```
{"version":3,"file":"colors.d.ts","sourceRoot":"","sources":["../../src/config/colors.ts"],"names":[],"mappings":"AAiBA,QAAA,MAAM,MAAM;;;;;;;;;;;;;;;;;;;;;;;;CAAgB,CAAC;AAE7B,eAAe,MAAM,CAAC"}
```

### Suggested solution(s)

I have this working using the following rollup config, not sure how well this transfers to tsdx:

```js
import camelCase from 'lodash.camelcase';
import commonjs from 'rollup-plugin-commonjs';
import json from 'rollup-plugin-json';
import resolve from 'rollup-plugin-node-resolve';
import sourceMaps from 'rollup-plugin-sourcemaps';
import typescript from 'rollup-plugin-typescript2';
import ts from 'typescript';

const pkg = require('./package.json');

const libraryName = 'vulk-core';

export default {
  input: `src/${libraryName}.ts`,
  output: [
    {
      file: pkg.main,
      name: camelCase(libraryName),
      format: 'umd',
      sourcemap: true,
    },
    { file: pkg.module, format: 'es', sourcemap: true },
  ],
  // Indicate here external modules you don't wanna include in your bundle (i.e.: 'lodash')
  external: [],
  watch: {
    include: 'src/**',
  },
  plugins: [
    // Allow json resolution
    json(),
    // Compile TypeScript files
    typescript({
      useTsconfigDeclarationDir: true,
      typescript: ts,
    }),
    // Allow bundling cjs modules (unlike webpack, rollup doesn't understand cjs)
    commonjs(),
    // Allow node_modules resolution, so you can use 'external' to control
    // which external modules to include in the bundle
    // https://github.com/rollup/rollup-plugin-node-resolve#usage
    resolve(),

    // Resolve source maps to the original source
    sourceMaps(),
  ],
};
```

### Your environment

| TSDX           | 0.5.11
| TypeScript       | 3.4.5
| Browser          | N/A
| npm/Yarn         | yarn (with workspaces)
| Operating System | MacOs

etienne-dldc:
Following #465,

This enable the `useTsconfigDeclarationDir` option of  `rollup-plugin-typescript2` when `declarationDir` is set in tsconfig.

Probably Fix #135 too


etienne-dldc:
CI fails because "Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`." should I run yarn and update the PR ?

agilgur5:
> CI fails because "Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`." should I run yarn and update the PR ?

If you rebase with master and force-push, it should be fixed. It was just fixed in #470

test/fixtures/build-declarationDir/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
agilgur5: `target` was just removed from all test fixtures in #466

justingrant:
@etienne-dldc @agilgur5 -  if `declarationDir` is omitted, then `useTsconfigDeclarationDir` will be false, which in turn will trigger #479.  This is different from tsc's behavior which is that `outDir` will be used to emit declaration files. I think it's probably a bad idea for default tsdx behavior to be different from default tsc behavior for the same config.

What do you think about making `useTsconfigDeclarationDir` true by default in rollup config, even if there's no `declarationDir`?  If yes, then this tsdx config setting could be an opt-out to handle cases where that default is not desired.

I read @agilgur5's [comment](https://github.com/jaredpalmer/tsdx/issues/465#issuecomment-579877403) in #465, but I just ran rollup with no `declarationDir` and also with `useTsconfigDeclarationDir: true`. Rolllup emitted all files in the right place *and* emitted correct sourcemap and declaration map files.  So I'm not sure what bad outcome would happen if `useTsconfigDeclarationDir: true` was default.

agilgur5:
> This is different from tsc's behavior which is that `outDir` will be used to emit declaration files.

@justingrant I don't think this specific PR is the right place to discuss `outDir` because TSDX doesn't really support `outDir` as I mentioned in https://github.com/jaredpalmer/tsdx/issues/479#issuecomment-581173998.
We can add support for that separately. This PR is specific to `declarationDir` support

> I read @agilgur5's [comment](https://github.com/jaredpalmer/tsdx/issues/465#issuecomment-579877403) in #465, but I just ran rollup with no `declarationDir` and also with `useTsconfigDeclarationDir: true`. Rolllup emitted all files in the right place _and_ emitted correct sourcemap and declaration map files. So I'm not sure what bad outcome would happen if `useTsconfigDeclarationDir: true` was default

I do think we'd want testing around this because rpts2 does behave differently when it is set vs. unset, as shown by the rpts2 code I pointed to in my comment.
Did that work with `outDir` set? Or with no `outDir` and no `declarationDir`?? You changed your repro and didn't provide a specific branch here, so the specifics of when it worked is opaque and confusing me

justingrant:
@agilgur5 - oops, I didn't notice that tsdx doesn't support `outDir`. Probably because my `outDir` was just `./dist` which I assume is the default.  I agree with you that `outDir`-related questions are out of scope for this PR. Good point.

That said, I still think my question is relevant: would it be good for `useTsconfigDeclarationDir` to be true by default, regardless of whether `declarationDir` is present in tsconfig?  What problems can happen if `useTsconfigDeclarationDir` is true but `declarationDir` is missing?

> Did that work with outDir set? Or with no outDir and no declarationDir?? You changed your repro and didn't provide a specific branch here, so the specifics of when it worked is opaque and confusing me

Sorry for confusion. I observed that if `useTsconfigDeclarationDir` is true (I ensured this by hacking tsdx's createRollupConfig.js) then it didn't matter whether `declarationDir` was `./dist`, `dist`, `dist/`, or was not present. In all those cases, tsdx correctly emitted `index.d.ts` into `dist` and `index.d.ts.map` was emitted with correct `sources` paths. While running these tests, `outDir` was set to `./dist`, so I'm not sure about behavior if there's no `outDir`.  You are right that tests are needed to cover all the cases. Let me know if there's more clarification I could provide here about the behavior I observed.

agilgur5:
> @agilgur5 - oops, I didn't notice that tsdx doesn't support `outDir`. Probably because my `outDir` was just `./dist` which I assume is the default. I agree with you that `outDir`-related questions are out of scope for this PR. Good point.

Yea `outDir` set to anything else won't actually make it output to a different directory and also multiple files will not be output correctly (iirc they'll overwrite each other since there's some name hard-coding). It is still passed to rpts2 which passes it to TS Language Service, so it does have _some_ effect as you've seen.
_(There's #321 specifically for `outDir` and then like 10 issues related to `outDir`, `preserveModules`, and multi-entry; pretty sure these are the most popular issue by a wide margin. My #367 goes a good amount of the way toward solving some, but not all of these problems. It's quite non-trivial as well, so digging into those weeds is a lot for this comparatively extremely small PR 😅 )_

> That said, I still think my question is relevant: would it be good for `useTsconfigDeclarationDir` to be true by default, regardless of whether `declarationDir` is present in tsconfig? What problems can happen if `useTsconfigDeclarationDir` is true but `declarationDir` is missing?

I'll refer to my proposed fix in https://github.com/jaredpalmer/tsdx/issues/479#issuecomment-581184234 for any other readers.
Idk what can happen, hence the anxiety around it 😅 The code paths _are_ different, so don't want to change that and potentially break people's builds without a fix or ample testing. The proposed fix I reference above would add a default `declarationDir`, so that would ensure the same code path regardless. **No changes required here in this PR.**
  
> I observed that if `useTsconfigDeclarationDir` is true (I ensured this by hacking tsdx's createRollupConfig.js) then it didn't matter whether `declarationDir` was `./dist`, `dist`, `dist/`, or was not present. In all those cases, tsdx correctly emitted `index.d.ts` into `dist` and `index.d.ts.map` was emitted with correct `sources` paths. While running these tests, `outDir` was set to `./dist`, so I'm not sure about behavior if there's no `outDir`. You are right that tests are needed to cover all the cases.

Ok, so they were both only ever set to `dist/` and `outDir` was always set. Thanks for the summary. That covers no `declarationDir` and yes `outDir` case. When neither is set but `useTsconfigDeclarationDir` is set is another that is potentially problematic. In any case, my proposed fix should workaround any potential problems.

test/fixtures/build-withTsconfig/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "module": "ESNext",
+    "lib": ["dom", "esnext"],
+    "declaration": true,
+    "declarationDir": "typings",
+    "declarationMap": true,
+    "sourceMap": true,
+    "rootDir": "./src",
```
agilgur5:
```suggestion
    "rootDir": "./",
```

For consistency with the other fixtures. `"./"` is actually incorrect, according to #488 , but I'd rather make all the changes to `"./src"` in one place.
This one change made me very confused as to why the declarationMap was output correctly here but not in the other fixtures when working on #488 (see https://github.com/jaredpalmer/tsdx/issues/479#issuecomment-581721483 and the comment before it)

test/fixtures/build-withTsconfig/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "strictFunctionTypes": true,
+    "strictPropertyInitialization": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "moduleResolution": "node",
+    "baseUrl": ".",
```
agilgur5:
```suggestion
    "baseUrl": "./",
    "paths": {
      "*": ["src/*", "node_modules/*"]
    },
```

Similarly for consistency with the other fixtures

agilgur5:
Welp this still hasn't been merged by Jared and #504 got merged in first (that one is high prio to be fair), so now `rootDir` needs to be set back to `'./src'` 

Annd there's a merge conflict in here from #476 which also got merged first before this 😭 

SORRY for the churn, I'd have merged this in a while ago if I could 😐

etienne-dldc:
I've messed up a little bit with git (sorry 😞  ) but should be good now...

Ready to merge again 😉

etienne-dldc:
👋 @jaredpalmer any update on this ?

jaredpalmer:
Sorry for churn! Thank you for the effort! Makes a lot of sense. Would you be able to summarize this in one shirt comment for the release notes for me?

etienne-dldc:
Sure  😃,

> Tsdx now uses `declarationDir` when found in `tsconfig.json`. This let you change where types declarations (and associated `.d.ts.map`) are emitted. It also allow you to get correct `.d.ts.map` paths by setting `declarationDir` to `dist`.
This PR fix #135 & #465 

If this is too long you can keen just the twi first sentences or even just the first one.

agilgur5:
Just one thing -- this doesn't completely fix #135 / #479. As I wrote in #479 and a bit here, this creates a _workaround_ for correct declaration map `sources`, but doesn't resolve the _default_ case. That is resolved by #488 and #504 which was split off it; though it's kind of an upstream problem in general: https://github.com/ezolenko/rollup-plugin-typescript2/issues/204

agilgur5:
I emailed Jared and have now been added as a collaborator so I've merged this myself now. I don't yet have permissions on NPM to publish though.

Thanks again for this change @etienne-dldc and sticking through with the delays and conflicts 🙏 💯  😄 

@all-contributors please add @etienne-dldc for bugs, code, tests

PR #472, opened 2020-01-30
Ensure co-located test files don't have definitions generated

### LINKED ISSUE #464: Typings for test files end up in dist folder
### Current Behavior

If you run 

```sh
npx tsdx create test-project # choose react
```
, create a test file (say `index.test.tsx`) with some content, and run `yarn build`, you'll end up with a file called `dist/index.test.d.ts`. 

### Expected behavior

TSDX shouldn't create this.

### Suggested solution(s)

Perhaps there's a filter you could apply? Idk 🤷‍♂ Haven't had time to research it yet.

### Your environment

| TSDX           | @latest
| TypeScript       | not installed
| Browser          | n/a
| npm/Yarn         | yarn 1x
| Node             |  lts
| Operating System | osx

selbekk:
This commit fixes an issue where type declaration files were created
for test files as well.

Fixes #464

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -145,6 +145,7 @@ export async function createRollupConfig(
             '**/*.test.ts',
             '**/*.spec.tsx',
             '**/*.test.tsx',
+            // TS defaults below
             'node_modules',
             'bower_components',
             'jspm_packages',
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -148,7 +148,7 @@ export async function createRollupConfig(
             'node_modules',
             'bower_components',
             'jspm_packages',
-            'dist',
+            paths.appDist,
           ],
           compilerOptions: {
             sourceMap: true,
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -141,6 +141,7 @@ export async function createRollupConfig(
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
           exclude: [
+            // all TS test files, regardless whether co-located or in test/ etc
             '**/*.spec.ts',
             '**/*.test.ts',
             '**/*.spec.tsx',
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -140,6 +140,16 @@ export async function createRollupConfig(
         cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
+          exclude: [
+            '**/*.spec.ts',
+            '**/*.test.ts',
+            '**/*.spec.tsx',
+            '**/*.test.tsx',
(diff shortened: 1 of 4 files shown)


selbekk:
Updated the list of excludes in all templates. I can centralise in the defaults if you want - especially since the list got a bit longer?

agilgur5:
I would think it's better to add these universally as it makes the excludes transparent for folks with project structures like yours, and makes no changes to those with `test/` directories either. In templates that means people with `test/` would have to delete the excludes (or leave them in redundantly); some might also not understand that they're not necessarily needed for all projects.

In any case, let's wait for some thoughts from the other maintainers as they might decline wholesale these changes, idk

agilgur5:
@sw-yx any thoughts on adding this to templates vs. adding it to `tsconfigDefaults` (of rpts2)?

sw-yx:
i dont have strong feelings about this, i can see arguments for both sides. happy to go with whatever you and @selbekk want to do.

selbekk:
If we add it to the tsconfigDefaults, we can fix this issue for both existing users of tsdx and new ones.

selbekk:
I'm not sure why this isn't working on the windows-latest build - should work as expected?

src/createRollupConfig.ts
```diff
@@ -140,6 +140,16 @@ export async function createRollupConfig(
...
         tsconfigDefaults: {
+          exclude: [
+            '**/*.spec.ts',
+            '**/*.test.ts',
+            '**/*.spec.tsx',
+            '**/*.test.tsx',
+            'node_modules',
+            'bower_components',
+            'jspm_packages',
+            'dist',
```
agilgur5:
```suggestion
            paths.appDist,
```

src/createRollupConfig.ts
```diff
@@ -140,6 +140,16 @@ export async function createRollupConfig(
         cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
+          exclude: [
+            '**/*.spec.ts',
+            '**/*.test.ts',
+            '**/*.spec.tsx',
+            '**/*.test.tsx',
```
agilgur5:
```suggestion
            '**/*.test.tsx',
            // TS defaults below
```

src/createRollupConfig.ts
```diff
@@ -140,6 +140,16 @@ export async function createRollupConfig(
         cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
+          exclude: [
```
agilgur5:
```suggestion
          exclude: [
          // all TS test files, regardless whether co-located or in test/ etc
```

sw-yx:
@all-contributors pls thank @selbekk for code

agilgur5:
@all-contributors please add @selbekk for code

PR #473, opened 2020-01-30
(fix): ensure Babel presets are merged

### LINKED ISSUE #443: Using jsx in .jsx files not achievable

agilgur5:
- previously, if preset-env were not found in the user's presets,
  their presets would be overwritten by a single preset-env entry
  - now it properly merges their presets with an entry for preset-env

Fixes #443 

I _think_ this is good, but still need to create a test set-up for custom Babel configs. This requires installing some dependencies for automated tests. The `withConfig` `tsdx.config.js` fixture also requires installing some deps for automated tests and is currently a manual test (which means it isn't tested frequently -- I haven't tested it once in all my PRs), so would like to get that to be auto too.

### DIFF
--- a/src/babelPluginTsdx.ts
+++ b/src/babelPluginTsdx.ts
@@ -104,11 +104,13 @@ export const babelPluginTsdx = babelPlugin.custom((babelCore: any) => ({
     );
 
     const babelOptions = config.options || {};
+    babelOptions.presets = babelOptions.presets || [];
 
-    const envIdx = (babelOptions.presets || []).findIndex((preset: any) =>
+    const envIdx = babelOptions.presets.findIndex((preset: any) =>
       preset.file.request.includes('@babel/preset-env')
     );
 
+    // if they use preset-env, merge their options with ours
     if (envIdx !== -1) {
       const preset = babelOptions.presets[envIdx];
       babelOptions.presets[envIdx] = createConfigItem(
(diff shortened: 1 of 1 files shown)


agilgur5:
#481 was created to follow-up on the lack of test set-up here.
#627 added integration tests with a Babel fixture and a regression test was added there to ensure that Babel presets are merged

PR #475, opened 2020-01-30
(types): enforce stricter typings

agilgur5:
- noUnusedParameters, noImplicitReturns, and noFallthroughCasesInSwitch
  - improve typings for 1 unused parameter and 1 implicit return

- also remove noImplicitAny as that's redundant with strict: true

This is the same level of strictness as in the templates and I believe the highest level of strictness (I use the same level in my own libraries too).

The description of `noImplicitReturns` is a bit different from the name, not sure if it would've helped with something like #371 or not.

### DIFF
--- a/src/createEslintConfig.ts
+++ b/src/createEslintConfig.ts
@@ -14,7 +14,7 @@ export async function createEslintConfig({
   pkg,
   rootDir,
   writeFile,
-}: CreateEslintConfigArgs): Promise<CLIEngine.Options['baseConfig']> {
+}: CreateEslintConfigArgs): Promise<CLIEngine.Options['baseConfig'] | void> {
   const isReactLibrary = Boolean(getReactVersion(pkg));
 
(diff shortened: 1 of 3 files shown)


src/createEslintConfig.ts
```diff
@@ -14,7 +14,7 @@ export async function createEslintConfig({
   pkg,
   rootDir,
   writeFile,
-}: CreateEslintConfigArgs): Promise<CLIEngine.Options['baseConfig']> {
+}: CreateEslintConfigArgs): Promise<CLIEngine.Options['baseConfig'] | void> {
```
agilgur5: I typed it as `void` as well because it seems intentional that when `--writeFile` succeeds, it doesn't `return config` (because it does `return config` when it errors/the `catch` case), but I'm not sure if that actually is intentional. Maybe it makes sense to `return config` in all cases?

PR #476, opened 2020-01-30
(types): add @types/sade, fix transpileOnly default

agilgur5:
- transpileOnly default being the wrong type was found as a result of
  the better typing

I originally made this as a follow-up to #407 , as `prog` is passed around as an argument there (and is `any` without these types), but that hasn't gotten a response in almost a month now 😐 😐
So figured I'd separate it out.

Just as a note note, the `@types/sade` package was reviewed by DT, but never got a response from the maintainer, https://github.com/lukeed/sade/issues/27#issuecomment-554268697

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "@types/react": "^16.9.11",
     "@types/rollup-plugin-json": "^3.0.2",
     "@types/rollup-plugin-sourcemaps": "^0.4.2",
+    "@types/sade": "^1.6.0",
     "@types/semver": "^7.1.0",
     "doctoc": "^1.4.0",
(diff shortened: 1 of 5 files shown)


agilgur5:
Huh, this test is failing locally too and I'm not totally sure why 🤔 Is `'false'` getting evaluated wrong?

agilgur5:
Yep, looks like `'false'` was kept as a string. The boolean worked before but didn't type-check and [the docs](https://github.com/lukeed/sade#value) say:
> You probably only want to define a default value if you're expecting a String or Number value type.

So changed it to not have a default and modified the conditional

Issue #351, opened 2019-11-26
Change output folder name from "dist" to "build" ?


extmchristensen:
### Current Behavior

I can not find any way to override tsdx output to use any other folder than "dist"

### Desired Behavior

Would like to be able to specify another output folder such as "build" instead of "dist"

### Suggested Solution

--distFolder option.

### Who does this impact? Who is this for?

Users that don't want the generated output in their dist folder.

### Describe alternatives you've considered

Manually rename/move generated files.

sw-yx:
i havent tried this out yet but have you tried setting it in tsconfig.json?

arthurdenner:
It seems like the only way possible to do it at the moment is through `tsdx.config.js`:

```javascript
module.exports = {
  rollup(config, options) {
    config.output.file = config.output.file.replace('dist', 'build');

    return config;
  },
};
```

Regarding how opinionated `tsdx` wants to be, would be better to have it configurable.

sw-yx:
yeah. my inclination is to leave this sort of thing inside rollup config. and document it :)

arthurdenner:
Would be nice to have some official position about these open requests - not from you specifically, I see that you're more like a community moderator - so that we can create documentation about it (I'd gladly help) but so far we have pending questions only. 😕

arvigeus:
Same as #321

agilgur5:
@arvigeus #321 is quite different, that's asking for a bulk compilation option similar to [Babel's `--out-dir`](https://babeljs.io/docs/en/babel-cli#compile-directories), not to configure the _name_ of the output directory, as this is. The issues are only similar in title.

It might be good to create a separate FAQ doc around common use-cases that can be implemented with `tsdx.config.js` and link to it from the README

agilgur5:
I'm considering allowing this to be configurable via `tsconfig.json`'s `outDir` or `outFile`. Right now both are ignored by TSDX and we should try to make all `tsconfig.json` work where possible for similar output to `tsc` and less unexpected unsupported options.

dkovacevic15:
u

> It seems like the only way possible to do it at the moment is through `tsdx.config.js`:
> 
> ```js
> module.exports = {
>   rollup(config, options) {
>     config.output.file = config.output.file.replace('dist', 'build');
> 
>     return config;
>   },
> };
> ```
> 
> Regarding how opinionated `tsdx` wants to be, would be better to have it configurable.

[...]

PR #488, opened 2020-02-04
(fix): declarationMaps (*.d.ts.map) should have correct `sources`

### LINKED ISSUE #479: `sources` paths are incorrect in declarationMap files (*.d.ts.map) emitted by tsdx
### Current Behavior

When using the `declarationMap` option in tsconfig.json, the `sources` array in declaration map files does not contain correct relative paths.  Repro steps: 
1. `git clone https://github.com/justingrant/ts-declaration-map-repro.git`
2. `cd ts-declaration-map-repro`
3. `yarn`
4. `yarn run with-tsc` - this will simply run `tsc` against this repo.
5. Open `./dist/index.d.ts.map`. It will begin with: 
`{"version":3,"file":"index.d.ts","sourceRoot":"","sources":["../src/index.tsx"]`.  Note the correct relative path to index.tsx
6. `yarn run with-tsdx` which uses `tdsx` instead of `tsc` but uses the same TS config.
7. Open `./dist/index.d.ts.map` again. Here's what you'll see: 
`{"version":3,"file":"index.d.ts","sourceRoot":"","sources":["src/index.tsx"]`.  Note that the path to index.tsx is incorrectly relative to the project root, not relative to `dist`.

### Expected behavior

Correct relative path to index.tsx (`../src/index.tsx`) in the map file's `sources` array.

### Suggested solution(s)

I'm assuming that there's a way to configure rollup-plugin-typescript2 so that declaration maps are emitted with the correct paths.

### Additional context

Note that only declaration map files are affected. Relative paths in regular sourcemap files works fine. 

I'm not sure if this is a problem with tsdx or rollup-plugin-typescript2. But I didn't see any mention of this problem in rollup-plugin-typescript2's issue list, so I figured that tsdx may be the culprit because it's the new kid on the block.  I'm assuming (perhaps naively) that if the problem was in rollup-plugin-typescript2 that it would have been reported in its GitHub issues already.

At first I thought that this problem was related to #465, but I think that issue is solely about how to handle the `declarationDir` config setting which I'm not using at all. Instead, I'm relying on the same `outDir` folder for both transpiled output and for declaration output.

This issue may be a dupe of #135, but I'm not sure, because #135 has a much more complex config. But the problem does look similar: in both issues the relative path to the source files is missing in declaration maps emitted by tsdx.

BTW, great library! Glad you're building this.

### Your environment

| TSDX           | tsdx@0.12.3
| TypeScript       | typescript@3.7.5
| Browser          | n/a (this is a build issue)
| npm/Yarn         | yarn@1.21.1
| Node             |  v12.13.1
| Operating System | MacOS Catalina 10.15.2

agilgur5:
- when declarationDir is set, `sources` gets set correctly, but
  otherwise its relative path will be incorrect
  - so always set declarationDir to paths.appDist

BREAKING:
- rootDir needed to be changed to ./src because the previous ./ caused
  type declarations to be generated in dist/src/ instead of just dist/
  - the moveTypes function handled moving the declarations back into
    dist/, but .d.ts.map files would be relative to their original
    output directory, and hence wrong
    - with this change, moveTypes is no longer needed and .d.ts.map
      files will have the correct relative path, but the rootDir change
      is breaking

Fixes #479 . This is built on top of #468 , so please merge that first.

I'm currently setting this as a Draft PR, to discuss options for the breaking change. The templates and hence a lot of code out there has `rootDir: './'` which is problematic/incorrect/buggy. But requiring users to change to `rootDir: './src'` would be breaking.

Two options around this in my head:
1. Leave all the `moveTypes` code in for now and mark it as deprecated. This would mean types will be put into `dist/` if you have the old `rootDir: './'` or the new `rootDir: './src'` (in the new case it's basically a no-op). But if using the old one, the types in `dist/` are going to have incorrect relative paths in their declaration maps if using that feature -- this means that #479 won't actually be solved for most users 😕 
2. Set a `tsconfigOverride` for `rootDir: './src'`. But now users can't set their own `rootDir` 😕 
To be fair, I'm not entirely sure why this is included in the templates however, because I don't believe TSDX fully supports custom `rootDir` (pretty sure some custom paths will have buggy output).

Both of these seem suboptimal to me 😕


agilgur5:
Ok, so the breaking changes were split out and set to deprecated with a warning in #504 . Rebased this and removed the removals so now it is _almost_ ready. **#488 needs to be merged first** and that now has a conflict from #476 and the `rootDir` needs to be set to `./src` there too because #504 got merged before it 😐

agilgur5:
Ok, so I merged #468 (as I'm now a collaborator) and have rebased this PR with master to resolve conflicts.

But now I'm not sure if this makes sense because it's more an upstream issue in https://github.com/ezolenko/rollup-plugin-typescript2/issues/204 . That being said, this does workaround that issue. It just potentially introduces other issues with other Rollup plugins and with customizations like changing the output directory in `tsdx.config.js` (i.e. #351 )

PR #489, opened 2020-02-04
(fix): parse tsconfig extends, trailing commas, and comments

### LINKED ISSUE #483: tsdx silently fails parsing of tsconfig files containing comments, trailing commas, etc. that TSC accepts
### Current Behavior

`tsc` supports tsconfig.json files that contain [comments](https://github.com/microsoft/TypeScript/pull/5450), trailing commas, and other features beyond standard JSON. 

Current tsdx, however, parses tsconfig.json using `fs.readJson` which ends up [calling JSON.parse](https://github.com/jprichardson/node-jsonfile/blob/7b53118f76792dd7eade80415fb84985aa6c2a4f/index.js#L34) which will throw an exception if faced with comments, trailing commas, and other non-standard JSON content. Making things worse, the exception is silently ignored, which makes it really hard to figure out why config settings are being ignored.  I had to step through tsdx code to figure out what was going on. 

https://github.com/jaredpalmer/tsdx/blob/4f6de1083393057903a1837fb6658f8058ee832f/src/createRollupConfig.ts#L46-L49

Luckily this JSON is only used once in tsdx (only in createRollupConfig.ts) and that JSON is only currently used for one setting: `esModuleInterop`. But #468 is about to use it too for processing `declarationDir`.

BTW, tsdx is in good company with this problem: Create React App [had the same problem](https://github.com/facebook/create-react-app/issues/6865).

There's also an [extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) feature in tsconfig which allows configuration inheritance across multiple tsconfig files. This isn't supported by tsdx either. See #484.

### Expected behavior

1) tsdx should accept the same tsconfig.json syntax that does. 
2) Syntax errors (or other unexpected errors) reading tsconfig.json should not be silently ignored. 

### Suggested solution(s)

For 1) Luckily, the nice people in TypeScript-land have exported their tsconfig parser.  From https://github.com/facebook/create-react-app/issues/6865#issuecomment-485962203:

> Unfortunately it looks like typescript uses a hand-written parser. I traced it down to parseJsonText which is defined in [src/compiler/parser.ts](https://github.com/Microsoft/TypeScript/blob/6487d1ffe0d903b1e55a187ddeb973fe0d445a2f/src/compiler/parser.ts#L717). Ahahaha.
> 
> If we're allowed to hook into the typescript package, it looks like they provide [public access to their parse function](https://github.com/microsoft/TypeScript/blob/760393f893358a462b00eef7e8fb92342b0f71ad/lib/typescript.d.ts#L3639-L3647).
> 

Note that there are a whole set of parse functions defined. Hopefully one of those would meet tsdx's needs. Ideally it'd also support the [`extends`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) keyword in tsconfig.json which is also currently unsupported by tsdx. 

for 2) I'd suggest white-listing ignorable exceptions (e.g. file not found?) and reporting the rest to the user. In particular, any syntax errors should be reported to the user and should fail the build. 

### Your environment

| TSDX           | tsdx@0.12.3
| TypeScript       | typescript@3.7.5
| Browser          | n/a (this is a build issue)
| npm/Yarn         | yarn@1.21.1
| Node             |  v12.13.1
| Operating System | MacOS Catalina 10.15.2

### LINKED ISSUE #484: tsdx ignores tsconfig.json "extends" keyword
### Current Behavior

There's an [extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) feature in tsconfig which allows configuration inheritance across multiple tsconfig files.  This however isn't recognized in tsdx, which parses tsconfig here: 

https://github.com/jaredpalmer/tsdx/blob/4f6de1083393057903a1837fb6658f8058ee832f/src/createRollupConfig.ts#L46-L49

The JSON object is used as-is, instead of drilling up to resolve parent config. 

### Expected behavior

tsdx should process `extends` just like tsc does.

### Suggested solution(s)

In a quick look through TypeScript's source code I couldn't find the place where `extends` is handled, but it's possible that TypeScript's exported tsconfig parsing functions might handle extends for us automatically. See https://github.com/microsoft/TypeScript/blob/760393f893358a462b00eef7e8fb92342b0f71ad/lib/typescript.d.ts#L3639-L3647 what what TS offers in its API.  The implementation is here: https://github.com/Microsoft/TypeScript/blob/6487d1ffe0d903b1e55a187ddeb973fe0d445a2f/src/compiler/parser.ts#L717

### Additional context

This problem is similar to #483 in that both this issue and that one are caused by parsing tsconfig differently than tsc does.

### Your environment

| TSDX           | tsdx@0.12.3
| TypeScript       | typescript@3.7.5
| Browser          | n/a (this is a build issue)
| npm/Yarn         | yarn@1.21.1
| Node             |  v12.13.1
| Operating System | MacOS Catalina 10.15.2

agilgur5:
- use ts.readConfigFile to properly parse a commas and comments from
  tsconfig instead of just trying to readJSON
- use ts.parseJsonConfigFileContent to properly parse `extends` and
  get recursive compilerOptions

- add tests for all of the above use cases
  - add them to build-withTsconfig which changes declarationDir;
    should only work if TSDX properly parses the tsconfig

NOTE: this is only necessary for internal, TSDX-specific parsing of
tsconfig.json. rollup-plugin-typescript2 already handles these options,
but internal options like esModuleInterop etc are also used & parsed
independently of rpts2

Fixes #483 ~with its own suggestion, eventually getting to https://github.com/facebook/create-react-app/pull/7248 as the solution~ **EDIT**: that only solved commas and comments, `extends` needed more work, see below comments.
Also fixes #484 as it's the same parser that does both. Still need to add tests for this case, thinking of just having all test fixtures extend a base fixture and then override it?

**This is built on top of #436 , please merge that first.**

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -10,10 +10,11 @@ import replace from '@rollup/plugin-replace';
 import resolve from '@rollup/plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import typescript from 'rollup-plugin-typescript2';
+import ts from 'typescript';
+
 import { extractErrors } from './errors/extractErrors';
 import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
-import * as fs from 'fs-extra';
 
 const errorCodeOpts = {
   errorMapFilePath: paths.appErrorsJson,
@@ -43,10 +44,15 @@ export async function createRollupConfig(
     .filter(Boolean)
     .join('.');
 
-  let tsconfigJSON;
-  try {
-    tsconfigJSON = await fs.readJSON(opts.tsconfig || paths.tsconfigJson);
-  } catch (e) {}
+  const tsconfigPath = opts.tsconfig || paths.tsconfigJson;
+  // borrowed from https://github.com/facebook/create-react-app/pull/7248
+  const tsconfigJSON = ts.readConfigFile(tsconfigPath, ts.sys.readFile).config;
+  // borrowed from https://github.com/ezolenko/rollup-plugin-typescript2/blob/42173460541b0c444326bf14f2c8c27269c4cb11/src/parse-tsconfig.ts#L48
+  const tsCompilerOptions = ts.parseJsonConfigFileContent(
+    tsconfigJSON,
+    ts.sys,
+    './'
+  ).options;
 
   return {
     // Tell Rollup the entry point to the package
@@ -89,7 +95,7 @@ export async function createRollupConfig(
       // (i.e. import * as namespaceImportObject from...) that are accessed dynamically.
       freeze: false,
       // Respect tsconfig esModuleInterop when setting __esModule.
-      esModule: tsconfigJSON ? tsconfigJSON.esModuleInterop : false,
+      esModule: tsCompilerOptions ? tsCompilerOptions.esModuleInterop : false,
       name: opts.name || safeVariableName(opts.name),
       sourcemap: true,
       globals: { react: 'React', 'react-native': 'ReactNative' },
@@ -136,7 +142,7 @@ export async function createRollupConfig(
         },
       },
       typescript({
-        typescript: require('typescript'),
+        typescript: ts,
         cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
@@ -165,9 +171,7 @@ export async function createRollupConfig(
           },
         },
         check: !opts.transpileOnly,
-        useTsconfigDeclarationDir: Boolean(
-          tsconfigJSON?.compilerOptions?.declarationDir
-        ),
+        useTsconfigDeclarationDir: Boolean(tsCompilerOptions?.declarationDir),
       }),
       babelPluginTsdx({
         exclude: 'node_modules/**',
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Fix conflicts

agilgur5:
Done.

agilgur5:
Welp, so since I merged #468 , I switched to use that fixture, which actually does something different than the default, i.e. sets `declarationDir`. So if `declarationDir` works properly, then it means this works... except it didn't 😕 

Turns out `ts.readConfigFile` will handle commas and comments and such, but not `extends`... 😕 😕 It just returns the JSON with an `extends`.

Seems like the same issue encountered in https://github.com/kulshekhar/ts-jest/issues/245 , the solution to which is non-trivial... welppppp. investigating alternatives right now

agilgur5:
Ok, some links I went through to figure this out:

- https://github.com/kulshekhar/ts-jest/issues/245#issuecomment-308545799
  - the PR for which replaced the older [TypeStrong/tsconfig](https://github.com/TypeStrong/tsconfig) 
     - which has this issue: https://github.com/TypeStrong/tsconfig/issues/21#issuecomment-560170429
       - which pointed to https://stackoverflow.com/a/53898219/1046439
- also looked at various things in CRA like https://github.com/facebook/create-react-app/pull/5537
- eventually searched in rpts2 and found [this line](https://github.com/ezolenko/rollup-plugin-typescript2/blob/42173460541b0c444326bf14f2c8c27269c4cb11/src/parse-tsconfig.ts#L48) which is much simpler than all of the above

So ended up using that last one

PR #491, opened 2020-02-05
(fix): jest shouldn't transform json files in node_modules

### LINKED ISSUE #161: jest occasionally tries to transform json files
### Current Behavior

Once in while I get the following warnings when I run 'tsdx test' command:
```
ts-jest[jest-transformer] (WARN) Got a unknown file type to compile (file: /Users/smashercosmo/Documents/projects/tmp-babel-plugin/node_modules/@babel/core/package.json). To fix this, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match this kind of files anymore.
ts-jest[jest-transformer] (WARN) Got a unknown file type to compile (file: /Users/smashercosmo/Documents/projects/tmp-babel-plugin/node_modules/resolve/lib/core.json). To fix this, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match this kind of files anymore.
ts-jest[jest-transformer] (WARN) Got a unknown file type to compile (file: /Users/smashercosmo/Documents/projects/tmp-babel-plugin/node_modules/globals/globals.json). To fix this, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match this kind of files anymore.
```

### Expected behavior

jest shouldn't try transforming json files

### Suggested solution(s)

Probably 'json' should be added to transformIgnorePatterns in jest config

### Your environment

| TSDX           | 0.7.2
| TypeScript       | 3.5.2
| npm         | 6.9.0
| Operating System | macOS Sierra 10.12.3

agilgur5:
- not sure why ts-jest was picking these up and giving warnings, but
  TSDX's custom transformIgnorePatterns is different from jest's
  default (everything in node_modules), so set it to ignore json too

Fixes #161

@jaredpalmer is there a reason TSDX has a custom `transformIgnorePattern`? It's been like that since [one of the first commits](https://github.com/jaredpalmer/tsdx/commit/6bbf07793424511cc245f62f52694925fe452342) so idk.
[Jest's default](https://jestjs.io/docs/en/configuration#testpathignorepatterns-arraystring) is just `/node_modules/`, so not sure why TSDX's is slightly different. To allow for transforming TS files in `node_modules`?

### DIFF
--- a/src/createJestConfig.ts
+++ b/src/createJestConfig.ts
@@ -6,7 +6,7 @@ export function createJestConfig(
     transform: {
       '.(ts|tsx)': require.resolve('ts-jest/dist'),
     },
-    transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$'],
+    transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(js|jsx|json)$'],
     moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
     collectCoverageFrom: ['src/**/*.{ts,tsx}'],
     testMatch: ['<rootDir>/**/*.(spec|test).{ts,tsx}'],


agilgur5:
#513 fixes the root cause of this

Issue #381, opened 2019-12-19
hipster: build tsdx with tsdx


jaredpalmer:
We could compile tsdx with tsdx by first running it through tsc and then running it on itself. 

We would need to change the repo's tsc outDir to something like `tmp` and then we would just run `node ./tmp/index.js build ./tmp/index.js --tsconfig=tsconfig.compile.json`. We would probs need to tweak tsconfig to make it work. 

Kind of cool though. Probably useless.

agilgur5:
Had been thinking of doing this too for dogfooding, but didn't see much of a point to it (would probably be slower to build too). 

Could also use `ts-node` to simplify the above. Is there a reason `ts-node` isn't used internally? E.g. tests depend on the built `dist/` files, which can be out-of-date (have tested the wrong version multiple times now); with `ts-node` could always use the current version. Can send a PR to improve the tests with it if there isn't anything against it

Issue #289, opened 2019-10-29
Improve internal package testing


jaredpalmer:
I stole the current testing strategy from nytimes/kyt several years ago and have been using it on almost all of my CLI tools. Nowadays there are probably better ways to test CLI tools (including using tools like Verdaccio to run end-to-end tests). 

I want to open up discussion for improving testing and stability of tsdx.

agilgur5:
Just copy/pasting my thoughts from #492 here so they're in the same place:

> Would like to be able to support `jest --watch` for internal tests, but I believe that would require programmatic access to the CLI engine, which we don't currently have. #407 actually goes a long way toward this; it wouldn't be much work on top of it to expose the CLI actions as functions and then import those for internal tests.
> Just importing functions and having them in-memory should be a perf improvement over the current re-running of CLI multiple times, and `jest --watch`s incremental testing should be a significant improvement.
> Programmatic access and imports should also make it easier to get code coverage for TSDX, which is certainly lacking in some areas.
> 
> Could then leave `shelljs` stuff as integration tests (which are slower) which don't need to run locally for every change, but should continue running on CI.
> 
> A separate test perf improvement I was thinking of was having the temporary `stage-build` (and `stage-lint`) directory use a `tmpfs` like a RAMDisk or something, which should speed up the disk I/O considerably as the files be in-memory. Not sure how cross-platform this might be for the test matrix (and Windows contributors), but still have to investigate libraries for that.

PR #492, opened 2020-02-05
(test): use ts-node for tests for better testing DX

agilgur5:
- personally, I often forget to yarn build before yarn test and that
  means I'm testing stale changes
  - ts-node is one of the easiest ways to ensure the code isn't stale
    - but it comes at a big performance penalty, tests are about 1.5x
      slower, which is much longer than the additional time for
      yarn build

So I think the perf penalty here makes this a no-go, but figured I'd PR for posterity. See #493 as a somewhat similar alternative.

This was something I mentioned in https://github.com/jaredpalmer/tsdx/issues/381#issuecomment-576622565 but never got a response on 😐 😐 . Figured I might as well try and see what happens.

<hr>

Related to #289.
Would like to be able to support `jest --watch` for internal tests, but I believe that would require programmatic access to the CLI engine, which we don't currently have. #407 actually goes a long way toward this; it wouldn't be much work on top of it to expose the CLI actions as functions and then import those for internal tests.
Just importing functions and having them in-memory should be a perf improvement over the current re-running of CLI multiple times, and `jest --watch`s incremental testing should be a significant improvement

Could then leave `shelljs` stuff as integration tests (which are slower) which don't need to run locally for every change, but should continue running on CI.

A separate test perf improvement I was thinking of was having the temporary `stage-build` (and `stage-lint`) directory use a `tmpfs` like a RAMDisk or something, which should speed up the disk I/O considerably as the files be in-memory. Not sure how cross-platform this might be for the test matrix (and Windows contributors), but still have to investigate libraries for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,8 @@
     "pretty-quick": "^2.0.0",
     "ps-tree": "^1.2.0",
     "react": "^16.8.6",
-    "semver": "^7.1.1"
+    "semver": "^7.1.1",
+    "ts-node": "^8.6.2"
   },
   "husky": {
     "hooks": {
--- a/test/tests/tsdx-build.test.js
+++ b/test/tests/tsdx-build.test.js
@@ -8,6 +8,8 @@ const util = require('../fixtures/util');
 shell.config.silent = false;
 
 const stageName = 'stage-build';
+const buildCommand =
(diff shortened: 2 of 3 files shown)


agilgur5:
Self-closing due to aforementioned perf issues. See #493 as a somewhat similar alternative. Test perf comments still relevant here.

PR #502, opened 2020-02-08
Add support for remote custom templates

benjlevesque:
Proposal for #441 

Add support for custom remote templates:
```
tsdx create template --template https://github.com/benjlevesque/tsdx-template-basic my-app
```
Note: the repository must follow a specific structure
- there MUST be a `template` directory, containing the template files.
- there MIGHT be a `template.json` file, with the following structure:
```json
{
  "name": "...",
  "dependencies": [  ],
  "packageJson": {  }
}
```

Template structure & approach inspired from [CRA](https://github.com/facebook/create-react-app/tree/master/packages/cra-template).

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "sade": "^1.4.2",
     "shelljs": "^0.8.3",
     "tiny-glob": "^0.2.6",
+    "tmp-promise": "^2.0.2",
     "ts-jest": "^24.0.2",
     "tslib": "^1.9.3",
     "typescript": "^3.7.3"
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,6 +21,7 @@ import execa from 'execa';
 import shell from 'shelljs';
 import ora from 'ora';
 import semver from 'semver';
+import tmp from 'tmp-promise';
 import { paths } from './constants';
 import * as Messages from './messages';
 import { createBuildConfigs } from './createBuildConfigs';
@@ -112,6 +113,7 @@ prog
     )}]`
   )
   .example('create --template react mypackage')
+  .example('create --template https://github.com/myorg/mytemplate mypackage')
   .action(async (pkg: string, opts: any) => {
     console.log(
       chalk.blue(`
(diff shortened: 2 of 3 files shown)


jaredpalmer:
This looks good. Let’s fix tests and merge

jaredpalmer:
Needs docs too

benjlevesque:
> Needs docs too

@jaredpalmer where do you want the docs? I didn't find docs about the existing `--template` option

imflavio:
This would be very much appreciated! @jaredpalmer ?

emersonlaurentino:
what is missing for this PR to be dipped? tsdx is really interesting, but this feature is interesting for my usage context.

agilgur5:
I haven't reviewed this myself as I'd need to take a good bit of time to understand the use-case of how this is beneficial beyond what you can do with template repos and haven't had that time.

Off the top of my head:
- Auto-installing unvetted/untracked dependencies seems like an avenue for some security issues (my current day job).
- Usage of synchronous `fs.existsSync` instead of async `fs.pathExists`, which I've spent a good bit of time replacing/removing everywhere in the codebase where possible previously and would not want to be re-introduced unnecessarily

emersonlaurentino:
At my company, we need a tool similar to tsdx and we are thinking about it, but we need to use a custom template. So this resource is essential for us. Currently, we using a Create React Library fork, but it's no legal to maintain.

agilgur5:
> At my company, we need a tool similar to tsdx and we are thinking about it, but we need to use a custom template. So this resource is essential for us.

Couldn't you do the same thing with a template repo? Either via GitHub's templates or via copy+paste. I do that myself with my repos, which are a good bit different from TSDX's templates.

agilgur5:
Had looked into this before, but did some more digging to try to see what CRA's rationale was for this feature. Unfortunately I couldn't see a clear one:

- Here's [a proposal](https://github.com/facebook/create-react-app/issues/2960) for it. https://github.com/facebook/create-react-app/issues/2960#issuecomment-356108258 asks the same question of how it's different from `git clone` + `npm install`. It was never really further responded to
- 3 different PRs were made to add varying degrees of support, but none really go over rationale 😕  : https://github.com/facebook/create-react-app/pull/3371, https://github.com/facebook/create-react-app/pull/6514, then https://github.com/facebook/create-react-app/pull/7716 which was merged

luistak:
Hello, I'm really needing this feature

Is there anything I may help with?

PR #504, opened 2020-02-11
(fix): set rootDir to './src', not './'. deprecate moveTypes

agilgur5:
- rootDir needed to be changed to ./src because the previous ./ caused
  type declarations to be generated in dist/src/ instead of just dist/
  - the moveTypes function handled moving the declarations back into
    dist/, but occassionally had errors moving .d.ts files
    - particularly in CI and for projects with many of them
    - declarationMap (*.d.ts.map) files would also have issues due to
      the hackiness of moveTypes, setting to rootDir to './src' is one
      of the necessary steps in fixing them

- deprecate moveTypes and add a warning with instructions if it is used
  when a rootDir of './' is detected
  - add notes about a deprecation window in the comments

<hr>

Split out from #488 and **built on top of #500 (please merge that first)** . 
- Per https://github.com/jaredpalmer/tsdx/pull/500#issuecomment-583643567, I suspect that `fs.copy` issues being called inside of `moveTypes` is the root cause of the _very frequent_ CI issues we've had with missing `.d.ts.` files, which I've occasionally replicated locally (I saw it locally first and then it started happening more in CI).
  - Will try to re-run CI a few times here to get some results
- #488 goes into more details about the breaking change of changing `rootDir`, so here I've deprecated with a warning + instructions instead of fully removing it. I've also moved it to a `deprecated.ts` file with a clear notice in a comment.
  - As this changes the templates, this should be released sooner rather than later to stop the bleeding, which is another reason to split it off of #488 (#488 depends on this fix, so this can go first but not vice-versa. and #488 is dependent on #468 which still hasn't been merged)

### DIFF
--- a/src/deprecated.ts
+++ b/src/deprecated.ts
@@ -20,7 +20,9 @@ export async function moveTypes() {
   console.warn(
     '[tsdx]: Your rootDir is currently set to "./". Please change your ' +
       'rootDir to "./src".\n' +
-      'TSDX has deprecated setting tsconfig.compilerOptions.rootDir to "./".'
+      'TSDX has deprecated setting tsconfig.compilerOptions.rootDir to ' +
+      '"./" as it caused buggy output for declarationMaps and occassionally ' +
+      'for type declarations themselves.'
   );
 
   try {
--- a/src/deprecated.ts
+++ b/src/deprecated.ts
@@ -1,0 +1,41 @@
+import * as fs from 'fs-extra';
+
+import { paths } from './constants';
+
+/*
+  This was originally needed because the default
+  tsconfig.compilerOptions.rootDir was set to './' instead of './src'.
+  Now that it's set to './src', this is now deprecated.
+  To ensure a stable upgrade path for users, leave the warning in for
+  6 months - 1 year, then change it to an error in a breaking bump and leave
+  that in for some time too.
+*/
+export async function moveTypes() {
+  const appDistSrc = paths.appDist + '/src';
+
+  const pathExists = await fs.pathExists(appDistSrc);
+  if (!pathExists) return;
+
+  // see note above about deprecation window
+  console.warn(
+    '[tsdx]: Your rootDir is currently set to "./". Please change your ' +
+      'rootDir to "./src".\n' +
+      'TSDX has deprecated setting tsconfig.compilerOptions.rootDir to "./".'
+  );
+
+  try {
+    // Move the typescript types to the base of the ./dist folder
+    await fs.copy(appDistSrc, paths.appDist, {
+      overwrite: true,
+    });
+  } catch (err) {
+    // ignore errors about the destination dir already existing or files not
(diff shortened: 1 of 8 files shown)


agilgur5:
@sw-yx I'm _pretttty_ sure now that `moveTypes` was the root cause of missing `.d.ts` issue. There are now **40+** full matrix CI runs on this PR: see the 40 green checks of 38 empty commits made just to run CI, plus 2 regular commits. Statistically speaking, it's extremely likely that this PR solves the issue.
_Especially since each runs 9 CI runs (3 Ubuntu and 3 macOS at least, where the issue has been replicated) and each of those has many tests for `.d.ts` -- a failure in any of which would've cancelled the remaining matrix._

The errors encountered in #500 during the `moveTypes` also made it pretty likely that this was the issue as well.
And it seems to make sense that if `fs.copy`/`moveTypes` were the issue, that it would be happening more frequently over time, because we've added more tests and more fixture directories over time. Notably #370 which added a `syntax` directory (at my request), adding more files and more directories to copy

I think this PR's merge (and #500 that this depends on) & release, should be expedited (cc: @jaredpalmer ) because:
1. This error happens locally and is not limited to CI, meaning it can affect users of TSDX and their libraries before they are published. ["[T]hat's a BAD bug if it's true. oof", per Jared](https://github.com/jaredpalmer/tsdx/pull/367)
2. This error breaks CI pretty frequently and annoys me and @sw-yx a lot, as well as confuses contributors. Though the test changes _can_ be split from the template and deprecation changes.
3. Stop the bleeding due to the templates (and no warning), as mentioned in my opening comment.

src/deprecated.ts
```diff
@@ -0,0 +1,45 @@
...
+  const pathExists = await fs.pathExists(appDistSrc);
+  if (!pathExists) return;
+
+  // see note above about deprecation window
+  console.warn(
+    '[tsdx]: Your rootDir is currently set to "./". Please change your ' +
+      'rootDir to "./src".\n' +
+      'TSDX has deprecated setting tsconfig.compilerOptions.rootDir to ' +
+      '"./" as it caused buggy output for declarationMaps and occassionally ' +
+      'for type declarations themselves.'
```
sw-yx: want to make sure @jaredpalmer is ok with this but otherwise LGTM

agilgur5:
See https://github.com/jaredpalmer/tsdx/pull/500#issuecomment-605530750 for why I think `fs.copy` was causing issues. The `rootDir` _should_ still be `src`, but I think `moveTypes` _might_ be fixable.

agilgur5:
See #691 for a fix to the root cause of the `moveTypes()` issue that gave us so many errors and CI issues.

There are still issues with `rootDir` `./` because it's still not accurate, so changing it to `./src` was still correct and does still solve issues (with `declarationMap`s in particular), but that CI issue now has a root cause fix for it too with #691 .

Issue #496, opened 2020-02-06
Explore rollup-plugin-visualizer


jaredpalmer:
Seems useful. Perhaps we add an `--inspect` flag?

arvinsim:
Were you guys able to make `rollup-plugin-visualizer` work with resorting to the `objectHashIgnoreUnknownHack` hack to circumvent the `asyncfunction` issue?

agilgur5:
@arvinsim I'm fairly certain your hunch is correct that this is blocked on https://github.com/puleos/object-hash/pull/90 .
In theory disabling `rpts2`'s cache _only_ during `--inspect` isn't that bad, but I'm not a big fan of adding a hack on top of a hack to the codebase as that can lead to unexpected issues, esp. when they involve caching.

agilgur5:
Ok so https://github.com/puleos/object-hash/pull/90 was just merged and released so now we're blocked on https://github.com/ezolenko/rollup-plugin-typescript2/pull/203 (PR from me).

Notably, `rollup-plugin-visualizer` was mentioned in #200 as an explicit example of one of the things that wasn't possible due to `rpts2`'s lack of async/await support.

I mentioned in https://github.com/puleos/object-hash/pull/90 that it'd be difficult to fork/patch a transitive dependency, but this actually goes a step further as `rpts2` doesn't have externals for its dependencies and actually _bundles_ them in its own release (https://github.com/ezolenko/rollup-plugin-typescript2/pull/80, https://github.com/ezolenko/rollup-plugin-typescript2/pull/70), so TSDX _can't_ override the version of `object-hash` that `rpts2` uses (on top of having duplicated dependencies included in it...) 😕 😕

Issue #358, opened 2019-12-03
Add closure compiler 


jaredpalmer:
Steal it from here : https://github.com/ReactTraining/history/blob/dev/rollup.config.js

ambroseus:
right link: https://github.com/ReactTraining/history/blob/dev/scripts/rollup.config.js

ambroseus:
1st impression: it works :)
just need to tune up typescript2 plugin: https://github.com/ezolenko/rollup-plugin-typescript2#plugins-using-asyncawait
result:
![Screenshot from 2019-12-16 22-37-28](https://user-images.githubusercontent.com/380645/70941252-b426ba00-2054-11ea-86c0-43a86b7a7cdc.png)

ambroseus:
oh, I see difference in bundle size:
minified with terser: ~ 13900 b
with closure compiler: ~ 12700 b
(compilation_level: 'ADVANCED_OPTIMIZATIONS')

sw-yx:
good! what are the tradeoffs of simple vs advanced optimization?

we should try running against formik to get a good gauge on % savings 

nice work so far!

ambroseus:
> good! what are the tradeoffs of simple vs advanced optimization?
SIMPLE_OPTIMIZATIONS do same things as Terser plugin (and in my case optimized bundle size was little bit larger than  terser's one)
https://developers.google.com/closure/compiler/docs/api-tutorial3

BUT. there was a problem with 'ADVANCED_OPTIMIZATIONS' level in my lib with regeneratorRuntime polyfill

ambroseus:
> just need to tune up configuration for typescript2 plugin:
> https://github.com/ezolenko/rollup-plugin-typescript2#plugins-using-asyncawait
> objectHashIgnoreUnknownHack: true

continue to play with closure compiler due to #294 fix

ambroseus:
can be implemented as `tsdx plugin`

DrSensor:
Dynamic import doesn't work on the closure compiler 🤔
https://github.com/google/closure-compiler/issues/2770

Issue #278, opened 2019-10-22
Components doesn't ship with images properly


arvinsim:
### Current Behavior

I have been trying to create a react component with images for the purpose of reusing them across different apps.

However, when I build the library and import the component, the component works fine except all the images are not referenced properly

I also tried running Storybook and I got the same issue.

### Expected behavior

Build process should ship images properly

### Additional context

1. I run `tsdx build` on the `prepare` script in the `package.json`
2. I tried using `babel-plugin-file-loader`, `rollup-plugin-url` and `rollup-plugin-image` but they didn't work.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.9.3
| TypeScript       | 3.6.2
| Browser          | Chrome Version 77.0.3865.120
| npm/Yarn         | Yarn 1.19.1
| Node             | 10.15.3
| Operating System | MacOS Mojave 10.14.6

sw-yx:
got a test repo?

arvinsim:
Okay, I created a [repo](https://github.com/arvinsim/tsdx-asset-import-bug) to reproduce the error. I commented out `rollup-plugin-url` and `rollup-plugin-image` in `tsdx.config.js`. When I try to use them, I get errors.

arvinsim:
Any updates on this?

cainaleaouk:
bump @sw-yx @arvinsim also interested in this...

sw-yx:
havent had time to look into this yet, sorry. if anyone wants to help investigate, or propose the solve, pls be my guest

asyarb:
Sharing my config workaround to get images to work with `tsdx`.

```js
// tsdx.config.js
const images = require('rollup-plugin-image-files')

module.exports = {
  rollup(config, options) {
    config.plugins = [
      images({ incude: ['**/*.png', '**/*.jpg'] }),
      ...config.plugins,
    ]

    return config
  },
}
```

It seems like your `rollup` plugin to handle images needs to come before `tsdx`'s plugins. (I'm guessing before `rollup-plugin-typescript2`?).

---

I tried to get this working with `@rollup/plugin-url`, but couldn't figure it out. If anyone is interested, my config with `@rollup/plugin-url` can be seen below.

```js
const rpts2 = require('rollup-plugin-typescript2')
const url = require('@rollup/plugin-url')

module.exports = {
  rollup(config, options) {
    config.plugins = [
      // Force the `url` plugin to emit files.
[...]

PR #506, opened 2020-02-12
(fix): upgrade rpts2 / object-hash to support async rollup plugins

### LINKED ISSUE #294: Support objectHashIgnoreUnknownHack flag from rpt2

agilgur5:
- object-hash v2.0.2 adds/fixes support for async functions
- rpts2 v0.26.0 upgrades object-hash to v2.0.2
  - rpts2 currently bundles several of its dependencies, so they can't be
    overridden manually and that's also why the object-hash change isn't
    visible in the yarn.lock file

- previously, TSDX users who wanted to use rollup plugins (via
  tsdx.config.js) that use async/await internally (e.g.
  rollup-plugin-copy, rollup-plugin-url, rollup-plugin-visualizer,
  rollup-plugin-smart-asset, etc, etc, etc), would run into
  'Unknown object type "asyncfunction"' from object-hash as rpts2
  hashes the rollup config itself for caching
  - the only fix was to hackily override the rpts2 config in
    tsdx.config.js (which requires brittle copy+paste as its config
    isn't exposed) and then use a config option of rpts2,
    objectHashIgnoreUnknownHack (which is hacky, per the name)
    - this could cause stale cache issues, or, if used with clean,
      would disable the cache entirely, both of which are suboptimal
      experiences
  - now this really common issue is solved and there's no hacking
    needed to use plugins that make use of async/await

NOTE: this does not affect source code that uses async/await, that
was already supported, this just affects plugins (in tsdx.config.js)
that use async/await

<hr>

Phew, finally got here after https://github.com/puleos/object-hash/pull/90 and https://github.com/ezolenko/rollup-plugin-typescript2/pull/203 got merged and released.

Fixes #294 (the underlying root cause to be specific) and should simplify a lot in #379 and #278 , as well as unblock #358 and #496 .

With regard to rpts2 bundling some of its own dependencies, see https://github.com/ezolenko/rollup-plugin-typescript2/pull/80, https://github.com/ezolenko/rollup-plugin-typescript2/pull/70 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "rollup-plugin-babel": "^4.3.2",
     "rollup-plugin-sourcemaps": "^0.4.2",
     "rollup-plugin-terser": "^5.1.2",
-    "rollup-plugin-typescript2": "^0.25.3",
+    "rollup-plugin-typescript2": "^0.26.0",
     "sade": "^1.4.2",
     "shelljs": "^0.8.3",
     "tiny-glob": "^0.2.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2925,10 +2925,10 @@ fill-range@^4.0.0:
     repeat-string "^1.6.1"
     to-regex-range "^2.1.0"
 
-find-cache-dir@^3.0.0:
-  version "3.1.0"
-  resolved "https://registry.yarnpkg.com/find-cache-dir/-/find-cache-dir-3.1.0.tgz#9935894999debef4cf9f677fdf646d002c4cdecb"
-  integrity sha512-zw+EFiNBNPgI2NTrKkDd1xd7q0cs6wr/iWnr/oUkI0yF9K9GqQ+riIt4aiyFaaqpaWbxPrJXHI+QvmNUQbX+0Q==
+find-cache-dir@^3.2.0:
+  version "3.2.0"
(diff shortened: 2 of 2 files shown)


jaredpalmer:
Awesome

PR #513, opened 2020-02-18
fix: anchor regexps in Jest transforms for proper JSON handling

kyle-johnson:
Master has a new regexp for Jest which breaks tests that reference JSON at some point in the dependency chain. This is a fix.

The regexp `.(js|jsx)` matches `.json` which means Jest attempts to load JSON files as JS modules. Since they aren't modules, Jest fails.

### DIFF
--- a/src/createJestConfig.ts
+++ b/src/createJestConfig.ts
@@ -4,8 +4,8 @@ export function createJestConfig(
 ) {
   const config = {
     transform: {
-      '.(ts|tsx)': require.resolve('ts-jest/dist'),
-      '.(js|jsx)': require.resolve('babel-jest'), // jest's default
+      '.(ts|tsx)$': require.resolve('ts-jest/dist'),
+      '.(js|jsx)$': require.resolve('babel-jest'), // jest's default
     },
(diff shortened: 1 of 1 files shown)


src/createJestConfig.ts
```diff
@@ -4,8 +4,8 @@ export function createJestConfig(
 ) {
   const config = {
     transform: {
-      '.(ts|tsx)': require.resolve('ts-jest/dist'),
-      '.(js|jsx)': require.resolve('babel-jest'), // jest's default
+      '.(ts|tsx)$': require.resolve('ts-jest/dist'),
```
kyle-johnson: Added the `$` here just to be consistent!

kyle-johnson:
This _may_ interact with #491

agilgur5:
Okkk just double-checked #161 and I believe this fixes it as well (and so #491 is not necessary).

The full absolute path there is `/Users/smashercosmo/Documents/projects/tmp-babel-plugin/node_modules/@babel/core/package.json` -- notably "projects/" has a "ts" in it

jaredpalmer:
Which should I merge?

agilgur5:
@all-contributors please add @kyle-johnson for bug, code

PR #524, opened 2020-02-28
Add JSX extension to @rollup/plugin-node-resolve options

### LINKED ISSUE #523: Could not resolve .jsx file when import from other .jsx
### Current Behavior

Encounter an error

```
Error: Could not resolve './ComponentC' from src/ComponentB.jsx
```

When import JSX file from JSX file with than 2 levels deep, eg

```sh
index.ts -> A.jsx -> B.jsx -> C.jsx
# Error: Could not resolve './C' from src/B.jsx

index.ts -> A.js -> B.jsx
# Error: Could not resolve './B' from src/A.js

index.js -> A.jsx
#  Error: Could not resolve './A' from src/index.js
```

Below is working fine

```
index.ts -> A.jsx -> B.jsx 

# or

index.ts -> A.jsx -> B.jsx -> C.jsx
index.ts -> B.jsx
```

Repo for reproducing
- https://github.com/n3tr/tsdx-import-jsx-test

### Expected behavior

The build command should be successfully

### Suggested solution(s)

I don't really know what is the root cause of this, but trying to set  `extensions: ['.mjs', '.js', '.jsx', '.json', '.node']` to `@rollup/plugin-node-resolve` seems to solve the issue.

https://github.com/jaredpalmer/tsdx/blob/3a6d42fb767150233c651056aabda9dd91bd048d/src/createRollupConfig.ts#L105-L111

### Your environment

| TSDX           | 0.12.3 |
| TypeScript       | 3.8.2 |
| Browser          |
| npm/Yarn         |
| Node             |  v12.4.0
| Operating System |

n3tr:
close #523 

Add JSX extension to rollup `@rollup/plugin-node-resolve` plugin 

[The default extensions value](https://github.com/rollup/plugins/blob/b6621262e284057e5740089388ffdcc6c391998d/packages/node-resolve/src/index.js#L25) doesn't include the `.jsx` and It causes an error when import `jsx` from `js` file.

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -108,6 +108,7 @@ export async function createRollupConfig(
           'main',
           opts.target !== 'node' ? 'browser' : undefined,
         ].filter(Boolean) as string[],
+        // defaults + .jsx
         extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
       }),
       opts.format === 'umd' &&
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -108,6 +108,7 @@ export async function createRollupConfig(
           'main',
           opts.target !== 'node' ? 'browser' : undefined,
         ].filter(Boolean) as string[],
+        extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
       }),
       opts.format === 'umd' &&
         commonjs({
--- a/test/fixtures/build-default/src/index.ts
+++ b/test/fixtures/build-default/src/index.ts
@@ -1,6 +1,8 @@
 import './syntax/nullish-coalescing';
 import './syntax/optional-chaining';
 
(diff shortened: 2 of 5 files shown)


src/createRollupConfig.ts
```diff
@@ -108,6 +108,7 @@ export async function createRollupConfig(
           'main',
           opts.target !== 'node' ? 'browser' : undefined,
         ].filter(Boolean) as string[],
+        extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
```
agilgur5: Can you add a comment that says like `// defaults + .jsx`?

Just to be explicit since there doesn't seem to be away to import the default extensions (as there is with Babel)

n3tr: Updated

n3tr:
@agilgur5 I added a comment and tests

agilgur5:
@allcontributors please add @n3tr for code, test, bugs

Issue #508, opened 2020-02-12
Configure bundled files -- disable dev builds


pedrouid:
I want to configure the name of the output files and which files are bundled.

By default, the JS that are output'd are an index file, cjs and esm files with both production and development versions and respective maps. But I want to build a single umd file named `index.js`.

So I  added the following flags to the build command:
```sh
tsdx build --format umd --name index
```

However this while output no index file and two umd files both production and development versions and respective maps. 

So I just added postbuild script to copy the contents of the production files and delete the other files.
```sh
# rename dist files
cp ./dist/index.umd.production.min.js ./dist/index.js 
cp ./dist/index.umd.production.min.js.map ./dist/index.js.map

# delete duplicates
rm -rf ./dist/index.umd.*
```

Ideally all I want to output is a single production minified file called `index.js` and respective map `index.js.map`.

How can I achieve this by configuring the `tsdx.config.js`??

agilgur5:
You can do:

```javascript
module.exports = {
  rollup (config, opts) {
    // redirect dev build to nowhere
    if (opts.environment === 'development') {
       config.output.file = '/dev/null'
    } else {
      // rename prod build to index.js
      config.output.file = './dist/index.js'
    }
  }
}
```

And I _think_ it should work (not 100% about sourcemaps, but think they should also be renamed fine), but as you'll notice it's kind of hacky with the `/dev/null`. There's no way to cancel a dev/prod build right now. So your current method is probably more optimal (and less subject to breakage)

EnKrypt:
This would really be good to have. I don't want to include development builds into the published tarball.

Right now I'm removing the development files and modifying index.js manually after publishing.

PR #525, opened 2020-02-28
(feat): closure compiler for minification

ambroseus:
Related to #358 
(for internal use only :)
Added ability to use `closure compiler` instead of `terser` plugin for code minification
```
"build": "tsdx build --closureCompiler"
```
to play with advanced options use `tsdx.config.js`
```js
const closureCompiler = require('@ampproject/rollup-plugin-closure-compiler');

const closureCompilerPlugin = closureCompiler({
  compilation_level: 'ADVANCED_OPTIMIZATIONS',
});

module.exports = {
  rollup(config) {
    config.plugins = config.plugins.map(plugin => {
      // override closure compiler plugin's default config
      if (plugin && plugin.name === closureCompilerPlugin.name) {
        return closureCompilerPlugin;
      }
      return plugin;
    });

    return config;
  },
};
```


src/createRollupConfig.ts
```diff
@@ -15,6 +16,15 @@ import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
 import * as fs from 'fs-extra';
 
+const USE_COMPILER = Boolean(process.env.CLOSURE_COMPILER);
+let compilerConfig: any;
+
+// check for closure-compiler.config.js
+if (USE_COMPILER && fs.existsSync(paths.compilerConfig)) {
```
agilgur5: use `await fs.pathExists`

src/createRollupConfig.ts
```diff
@@ -15,6 +16,15 @@ import { babelPluginTsdx } from './babelPluginTsdx';
 import { TsdxOptions } from './types';
 import * as fs from 'fs-extra';
 
+const USE_COMPILER = Boolean(process.env.CLOSURE_COMPILER);
```
agilgur5: why not use a CLI flag? microbundle has a similar discussion going in https://github.com/developit/microbundle/pull/569

src/createRollupConfig.ts
```diff
@@ -15,6 +16,15 @@ import { babelPluginTsdx } from './babelPluginTsdx';
...
+let compilerConfig: any;
+
+// check for closure-compiler.config.js
+if (USE_COMPILER && fs.existsSync(paths.compilerConfig)) {
+  // for ex. module.exports = { compilation_level: 'ADVANCED_OPTIMIZATIONS' }
+  compilerConfig = require(paths.compilerConfig);
```
agilgur5: why not just set defaults like with terser? adding a whole other config file seems suboptimal and you should be able to config via `tsdx.config.js` already

src/createRollupConfig.ts
ambroseus: fixed

src/createRollupConfig.ts
ambroseus: fixed

ambroseus:
@agilgur5  please review:
--closureCompiler CLI flag
closureComplierOptions section in tsdx.config.js

src/createBuildConfigs.ts
```diff
@@ -15,7 +15,8 @@ let tsdxConfig = {
 };
 
 if (fs.existsSync(paths.appConfig)) {
-  tsdxConfig = require(paths.appConfig);
+  const customConfig = require(paths.appConfig);
+  tsdxConfig = { ...tsdxConfig, ...customConfig };
```
agilgur5: why was this changed? it doesn't seem to have been in your original commit either 🤔

src/createRollupConfig.ts
```diff
@@ -186,18 +195,20 @@ export async function createRollupConfig(
...
-            keep_infinity: true,
-            pure_getters: true,
-            passes: 10,
-          },
-          ecma: 5,
-          toplevel: opts.format === 'cjs',
-          warnings: true,
-        }),
+        (opts.closureCompiler
+          ? compiler(closureCompilerOptions)
```
agilgur5: Sorry, maybe this wasn't clear in my last review -- let's just add default options in here, you had `ADVANCED_OPTIMIZATIONS` before. If a user wants to override this they can do that by overriding `config.plugins` in `tsdx.config.js` as with other plugins.

That's not to say that it's a bad idea to allow for easier configuration of plugins in `tsdx.config.js`, it's just that we haven't done it yet and I don't want to introduce a new API without thinking through it more, especially not for an experimental option.

I think a future API for easier configuration of plugins should probably reflect something like `customize-cra` -- I think I mentioned this in #389

test/fixtures/build-compiler/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
agilgur5: This is just used as a minifier and is just a flag, so I'm not sure a new fixture is necessary, it could be used with the default fixture. The other fixture directories are for different configurations.

If configuring Closure, could use the `build-withConfig` directory since it's for testing `tsdx.config.js` (though it's not currently set-up for automated tests -- that needs to be fixed)

src/createBuildConfigs.ts
ambroseus: without this patch tsdx will fall there https://github.com/jaredpalmer/tsdx/blob/216bb56392adf066ea959577701314d0ca899b16/src/createBuildConfigs.ts#L41 in case `tsdx.config.js` will not have { rollup } section

src/createRollupConfig.ts
ambroseus: The main goal is to add ability (not for all, but for core maintainers :)  to start playing with closure compiler as internal experimental feature and collect right options to become defaults.. (this is why I've used dedicated config file in initial commit).

agilgur5: Yea I know you mentioned for internal usage, I didn't know you meant for figuring out defaults.

I think the problem with libraries is that once you expose an API, you shouldn't break it too easily and should remain consistent. Hence my resistance to exposing one at all.

Maintainers / internal users would know how to adjust `createRollupConfig.ts` and `tsdx.config.js` so I don't think that's an issue.
It is marked as experimental, so I'm not worried about changing the defaults too much in that sense. We could _also_ say it's "UNSTABLE" to really get the point across 😅

agilgur5:
Yea, I think it's fine to leave tests out for now, don't want to add too much iteration for you since I'm not really sure myself what those should look like and `build-withConfig` isn't yet set-up for automated tests. Feel free to share any ideas on that front though

src/createBuildConfigs.ts
agilgur5: but it wasn't failing before??? I'm very confused, it started failing here right?

ambroseus: yes! there are no tests with fixture `build-withConfig`
proof:
Screenshot from 2020-03-11 09-31-04
Screenshot from 2020-03-11 09-29-51

ambroseus: https://github.com/jaredpalmer/tsdx/issues/572

agilgur5: > yes! there are no tests with fixture `build-withConfig`

Yes, I know that -- I mentioned it's not set-up for automated testing yet and that needs to be fixed at some point.

> #572

See my response there. This change makes sense when adding `closureCompilerOptions` to `tsdx.config.js`, but I don't think we should do that at this time.

ambroseus:
@agilgur5  so... may be my first approach with dedicated closureCompileOptions config file and env var instead of --closureCompiler flag is more applicable considering this is INTERNAL UNSTABLE VERY SECRET feature? ;)

jaredpalmer:
I think snapshots on code output make sense here for testing. It may be hard to understand prod output. While it’s not exactly an integration test, perhaps we could run prettier on the prod output of the test before the snapshot so it’s at least readable.

While this experimental, if it works, I can’t see why we wouldn’t use it in core. IIRC React core uses it. 

A plugin system is beyond the scope of this issue, but a good one. My 2cents on this, is that unlike Gatsby who’s core does almost nothing, TSDX’s core should be the happy path and cover 80-90% of use cases.

agilgur5:
> @agilgur5 so... may be my first approach with dedicated closureCompileOptions config file and env var instead of --closureCompiler flag is more applicable considering this is INTERNAL UNSTABLE VERY SECRET feature? ;)

Well I don't think it's "very secret" given this public PR haha 😝
But I think opening up two new, different API surfaces (a dedicated config file and an env var) is much more confusing than just adding a flag.
And I'm totally fine with getting feedback from the community on an "unstable experimental" feature, it doesn't have to be internal-only. This is the same way React and Rollup have incrementally released features too, though usually they'll initially prefix them as `unstable_` or `experimental` before.

It can be configured in `tsdx.config.js` with `config.plugins` already, so let's just use the same method for now.

> While this experimental, if it works, I can’t see why we wouldn’t use it in core.

Based on `microbundle`'s work in https://github.com/developit/microbundle/pull/569 and @ambroseus's testing in #358, it sounds like it's not _always_ better than Terser (but often is?). `ADVANCED_OPTIMIZATIONS` might also take longer or use more memory than Terser, but it wasn't clear to me if it necessarily did.

> A plugin system is beyond the scope of this issue, but a good one.

I still need to write up an RFC for this but I have it in my head. The plugin system would be more to make common use cases of `tsdx.config.js` easier. Because they often require additional dependencies or don't apply to everyone, it makes sense for it to be out of core, but perhaps still maintained internally for better maintainability against changes to the internal configs (and because they're common).
There's also a separate use of the word "plugin" in that a good bit of TSDX could probably be re-written as a Rollup plugin(s)

src/index.ts
```diff
@@ -386,6 +386,11 @@ prog
   .example('build --tsconfig ./tsconfig.foo.json')
   .option('--transpileOnly', 'Skip type checking')
   .example('build --transpileOnly')
+  .option(
+    '--closureCompiler',
+    'EXPERIMENTAL: Use closure compiler to minify production bundle'
```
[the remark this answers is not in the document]
ambroseus: fixed

test/tests/tsdx-build.test.js
```diff
@@ -102,6 +102,29 @@ describe('tsdx build', () => {
...
+    expect(shell.test('-f', 'dist/index.js')).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-compiler.cjs.development.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-compiler.cjs.production.min.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-compiler.esm.production.min.js')
+    ).toBeTruthy();
```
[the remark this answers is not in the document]
ambroseus: I've found difference))

test/tests/tsdx-build.test.js
```diff
@@ -102,6 +102,29 @@ describe('tsdx build', () => {
     expect(code).toBe(0);
   });
 
+  it('should compile files using closure compiler with options from tsdx.config.js', () => {
```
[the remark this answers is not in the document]
ambroseus: fixed

test/fixtures/build-compiler/src/syntax/nullish-coalescing.ts
```diff
@@ -0,0 +1,5 @@
+// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
```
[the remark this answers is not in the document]
ambroseus: fixed

src/createBuildConfigs.ts
ambroseus: please, check updated tests

test/fixtures/build-compiler/package.json
ambroseus: done

ambroseus:
@agilgur5  @jaredpalmer  please, review

there are some closure-compiler specific tests. only CC can minify this:
```js
import { split } from './foo';

export const sum = (a: number, b: number) => {
  return a + b;
};

const bar = split('bar');

export const signature = `${split('bar').join('')} ${sum(bar.length, -3)}`;
```
to this:
```js
var signature="bar 0";function sum(a,b){return a+b};export{signature,sum}
```

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,78 @@
...
+      shell.test('-f', 'dist/build-withconfig.esm.production.min.js')
+    ).toBeTruthy();
+
+    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+
+    // only closure compiler minifies bundle to `signature="bar 0"`
+    output = shell.grep('bar 0', [
+      'dist/build-withconfig.esm.production.min.js',
+    ]);
+    expect(/bar 0/.test(output.stdout)).toBeTruthy();
```
agilgur5: I think
```suggestion
    expect(output.code).toBe(0);
```
should be fine. It'll have an error code if it's not found

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,78 @@
...
+      shell.test('-f', 'dist/build-withconfig.cjs.production.min.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-withconfig.esm.production.min.js')
+    ).toBeTruthy();
+
+    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+
+    // only closure compiler minifies bundle to `signature="bar 0"`
+    output = shell.grep('bar 0', [
```
agilgur5:
```suggestion
    output = shell.grep(/bar 0/, [
```

src/createBuildConfigs.ts
ambroseus: removed

src/createRollupConfig.ts
```diff
@@ -2,6 +2,7 @@ import { safeVariableName, safePackageName, external } from './utils';
 import { paths } from './constants';
 import { RollupOptions } from 'rollup';
 import { terser } from 'rollup-plugin-terser';
+import compiler from '@ampproject/rollup-plugin-closure-compiler';
```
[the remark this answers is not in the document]
ambroseus: yep, done

src/createRollupConfig.ts
ambroseus: using
```js
closureCompiler({
   compilation_level: 'SIMPLE_OPTIMIZATIONS',
 })
```
 as defaults

test/fixtures/build-withConfig/src/index.ts
```diff
@@ -1,12 +1,9 @@
...
-import warning from 'tiny-warning';
-invariant(true, 'error occurred! o no');
-warning(false, 'warning - water is wet');
-export { foo } from './foo';
+import { split } from './foo';
 
 export const sum = (a: number, b: number) => {
-  if ('development' === process.env.NODE_ENV) {
-    console.log('fuck');
-  }
```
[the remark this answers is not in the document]
ambroseus: fixed

test/fixtures/build-withConfig/tsdx.config.closure-advanced.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  closureCompilerOptions: {
+    compilation_level: 'ADVANCED_OPTIMIZATIONS',
+  },
+};
```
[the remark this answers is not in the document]
ambroseus: fixed

test/fixtures/build-withConfig/tsdx.config.closure-simple.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  closureCompilerOptions: {
+    //SIMPLE_OPTIMIZATIONS level is default option
+    //compilation_level: 'SIMPLE_OPTIMIZATIONS',
+  },
+};
```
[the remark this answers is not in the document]
ambroseus: removed

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: fixed

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: yep, done

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,78 @@
...
+      'dist/build-withconfig.esm.production.min.js',
+    ]);
+    expect(/bar 0/.test(output.stdout)).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
ambroseus: done

ambroseus:
@agilgur5  next round), please review

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,90 @@
...
+ * @jest-environment node
+ */
+
+const shell = require('shelljs');
+const util = require('../fixtures/util');
+
+shell.config.silent = false;
+
+const grep = (pattern, file) => {
+  const output = shell.grep(pattern, file);
```
ambroseus: @agilgur5  unfortunately shelljs.grep output.code always 0 for matched/unmatched patterns, so need to test  output.stdout

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const lib = require(`../../${stageName}/dist`);
+    expect(lib.signature).toBe('bar 0');
+
+    expect(
+      grep(
+        'exports.signature=signature',
+        'dist/build-withconfig.cjs.production.min.js'
+      )
+    ).toBeTruthy();
```
ambroseus: pattern only for simple level (not with terser, not with advanced)

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const lib = require(`../../${stageName}/dist`);
+    expect(lib.signature).toBe('bar 0');
+
+    // closure compiler with advanced optimization level
+    // minifies bundle with `signature="bar 0"`
+    expect(
+      grep('bar 0', 'dist/build-withconfig.cjs.production.min.js')
+    ).toBeTruthy();
+  });
```
ambroseus: pattern only for advanced level (not with terser, not with simple)

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: https://github.com/shelljs/shelljs/blob/a8452b89f2b9dd372da81e973a16074a5628b0b6/src/grep.js#L73

agilgur5: Oh, welp, guess my tests in #627 don't completely work either then. Actual `grep` does have an error code when it doesn't match, that's a really confusing mismatch from `shelljs` 😕

Seems like that line doesn't throw because it's silent. `common.error` seems to throw when not silent.

Probably want to put this helper into `util` since it'll be used by other tests like in #627

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+const shell = require('shelljs');
+const util = require('../fixtures/util');
+
+shell.config.silent = false;
+
+const grep = (pattern, file) => {
+  const output = shell.grep(pattern, file);
+  const grepOutput = output.stdout.replace(`\n`, ``);
```
agilgur5: Should have a comment here on why this is necessary, e.g. `// shelljs.grep always has at least one newline`. But the `match` code you used before is probably better than this anyway.

test/tests/tsdx-build-closure-compiler.test.js
agilgur5: Should have a comment like there is in the advanced one.

Also should use `/pattern/` instead of `'pattern'` for explicitness because it gets evaluated like a regex and not a plain string. I tried using a plain string when I was writing some similar tests and it errored out because a regex would have required an escape there. Syntax highlighting would've shown that too.

test/tests/tsdx-build-closure-compiler.test.js
agilgur5: Why not test the whole expression like in the simple one? i.e. `exports.signature="bar 0"`

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,90 @@
...
+describe('tsdx build with closure compiler default options', () => {
+  beforeAll(() => {
+    util.teardownStage(stageName);
+  });
+
+  it('should minify bundle with default options', () => {
+    util.setupStageWithFixture(stageName, 'build-withConfig');
+
+    let output = shell.exec(
+      'node ../dist/index.js build --env production --closureCompiler'
```
agilgur5: `--env production` doesn't actually do anything (and isn't an official flag for build either). See also #508

src/createRollupConfig.ts
```diff
@@ -195,18 +196,22 @@ export async function createRollupConfig(
...
-            passes: 10,
-          },
-          ecma: 5,
-          toplevel: opts.format === 'cjs',
-          warnings: true,
-        }),
+        (opts.closureCompiler
+          ? closureCompiler({
+              compilation_level: 'SIMPLE_OPTIMIZATIONS',
+            })
```
agilgur5: this is the default, right? so can just be:
```suggestion
          ? closureCompiler()
```

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: sure, done

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: yep, fixed

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: yep, fixed

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: fully agree)

test/fixtures/build-withConfig/src/index.ts
```diff
@@ -1,12 +1,12 @@
...
 export const sum = (a: number, b: number) => {
   if ('development' === process.env.NODE_ENV) {
     console.log('fuck');
   }
   return a + b;
 };
+
+const bar = split('bar');
+
+export const signature = `${split('bar').join('')} ${sum(bar.length, -3)}`;
```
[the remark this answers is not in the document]
ambroseus: done

src/createRollupConfig.ts
ambroseus: yea, fixed

test/tests/tsdx-build-closure-compiler.test.js
ambroseus: fixed

test/fixtures/build-withConfig/src/index.ts
```diff
@@ -1,12 +1,12 @@
-import invariant from 'tiny-invariant';
-import warning from 'tiny-warning';
-invariant(true, 'error occurred! o no');
-warning(false, 'warning - water is wet');
-export { foo } from './foo';
+import { split } from './foo';
```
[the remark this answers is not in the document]
ambroseus: done

ambroseus:
@agilgur5 final (maybe:) round, please review

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+describe('tsdx build with closure compiler', () => {
+  beforeAll(() => {
+    util.teardownStage(stageName);
+  });
+
+  it('should minify bundle with default options', () => {
+    util.setupStageWithFixture(stageName, 'build-withConfig');
+
+    let output = shell.exec('node ../dist/index.js build --closureCompiler');
```
agilgur5: `const`

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,70 @@
...
+    // simplePattern 'exports.signature=signature'
+    expect(util.grep(simplePattern, cjsBundleProd)).toBeTruthy();
+    expect(util.grep(advancedPattern, cjsBundleProd)).toBeFalsy();
+  });
+
+  it('should minify bundle with advanced options', () => {
+    util.setupStageWithFixture(stageName, 'build-withConfig');
+    shell.mv('-f', 'tsdx.config.closure-advanced.js', 'tsdx.config.js');
+
+    let output = shell.exec('node ../dist/index.js build --closureCompiler');
```
agilgur5: `const`

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,70 @@
...
+    expect(shell.test('-f', cjsBundleProd)).toBeTruthy();
+    expect(shell.test('-f', esmBundle)).toBeTruthy();
+
+    const lib = require(`../../${stageName}/dist`);
+    expect(lib.signature).toBe('bar 0');
+
+    // with SIMPLE optimization level minified bundle should contain
+    // simplePattern 'exports.signature=signature'
+    expect(util.grep(simplePattern, cjsBundleProd)).toBeTruthy();
+    expect(util.grep(advancedPattern, cjsBundleProd)).toBeFalsy();
```
agilgur5: aren't these mutually exclusive anyway? like they're minifying the same export

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,70 @@
...
+const shell = require('shelljs');
+const util = require('../fixtures/util');
+
+shell.config.silent = false;
+
+const stageName = 'stage-build-closure-compiler';
+
+const cjsBundleDev = 'dist/build-withconfig.cjs.development.js';
+const cjsBundleProd = 'dist/build-withconfig.cjs.production.min.js';
+const esmBundle = 'dist/build-withconfig.esm.js';
```
agilgur5: think these should be left in the code as literals for now to be consistent with all the other tests

test/tests/tsdx-build-closure-compiler.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+const simplePattern = /exports\.signature=signature/;
+const advancedPattern = /exports\.a="bar 0"/;
```
agilgur5: huh... doesn't this minification break the usage?? since signature is no longer exported. is that a bug in closure or is there a way to disable exports minification?

ambroseus: @agilgur5  hmm.. this is not good.. need to investigate (a lot of ADVANCED options)

test/tests/tsdx-build-closure-compiler.test.js
agilgur5: I'm probably gonna make some test helpers to cut down on use of literals soon; I've been meaning to, and they'd make renames simpler, among other consistency

test/tests/tsdx-build-closure-compiler.test.js
agilgur5: Ah, I haven't looked the API at all. Keywords would likely be "exports" and "mangle" as I believe the technical term for changing variable names is "mangling"

agilgur5:
While this is waiting for more changes, figure I should credit you in `allcontributors` for your contributions especially since a piece of the code here went into #627 already.

agilgur5:
@allcontributors please add @ambroseus for test, examples, questions, ideas

ambroseus:
@agilgur5 thanks! proud to work with `tsdx` team and you personally :)

ambroseus:
@agilgur5  rebased & moved tests to integration folder. please check)

yarn.lock
```diff
@@ -2,152 +2,147 @@
...
+    uuid "3.4.0"
+
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.3":
   version "7.8.3"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
   integrity sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==
   dependencies:
     "@babel/highlight" "^7.8.3"
 
+"@babel/compat-data@^7.8.6", "@babel/compat-data@^7.9.0":
```
agilgur5, line 33: Something definitely went wrong with the merge conflict resolution here; [your initial commit](https://github.com/jaredpalmer/tsdx/pull/525/commits/a17e146f9f3ac637cb285bb029942d740153a75b) has a **10x** smaller `yarn.lock` diff.

When resolving conflicts with lockfiles like this, I'd recommend throwing out the changes you have in the lockfile and just re-generating the lockfile changes with `yarn`. So take what was added in the conflict and then run `yarn` on top of it to re-add your changes.

test/integration/tsdx-build-withClosure.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    ).toBeTruthy();
+    expect(shell.test('-f', 'dist/build-withclosure.esm.js')).toBeTruthy();
+
+    const lib = require(`../../${stageName}/dist`);
+    expect(lib.signature).toBe('bar 0');
+
+    // with ADVANCED optimization level minified bundle should contain
+    // advancedPattern 'exports.a="bar 0"'
+    expect(
+      grep(/exports\.a="bar 0"/, [
```
agilgur5, line 67: this one hasn't been fixed yet -- this is the blocker in this PR

ambroseus:
@agilgur5  ... time is going) I thought a lot about this feature and decide to close PR, because it's not very useful right now. We can implement it later as tsdx plugin for ex.

PR #526, opened 2020-02-28
(feat): support custom jest config paths via --config

### LINKED ISSUE #150: Cannot set custom jest.config.{js,ts} even with --config flag in test CLI command

agilgur5:
- --config will now be parsed shallow merged with the defaults,
  just like package.json.jest already is

- previously adding --config to tsdx test would result in jest
  outputting a usage prompt (I believe due to the second --config
  that's added internally) and then the somewhat cryptic
  "argv.config.match is not a function"
  - if --config is detected, it will be parsed, merged, and then
    deleted from argv so that this error doesn't occur anymore

Also dogfood `tsdx test` internally (this was the last thing needed for it to be possible) & improve Jest config typings

Fixes #150 , or rather what's left of it after #229 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "build": "tsc -p tsconfig.json",
     "lint": "yarn build && yarn lint:post-build",
     "lint:post-build": "node dist/index.js lint src test --ignore-pattern 'test/tests/lint'",
-    "test": "yarn build && jest --config ./test/jest.config.json",
+    "test": "yarn build && yarn test:post-build",
+    "test:post-build": "node dist/index.js test --config ./test/jest.config.json",
     "watch": "chokidar \"./package.json\" \"./src/**/*.ts\" \"node_modules\\@jaredpalmer\\rollup-plugin-preserve-shebang\\dist\\index.js\" -c \"yarn build && echo Success\"",
     "start": "tsc -p tsconfig.json --watch"
   },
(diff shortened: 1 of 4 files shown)


agilgur5 (commented):
Seems ok to me. Would be good if there were a simpler way to remove the config argument or to customize Jest but probably can't do much better as of right now due to [Jest's very bare API](https://github.com/facebook/jest/issues/5048).

Issue #299, opened 2019-11-04
Example with subdirectories


arvigeus:
I want to be able to do this:

```
import { foo } from "foo";
import { bar } from "foo/bar"
```

Is this a bad practice? If not, can we have an example of how to do it in the docs?

sw-yx:
its fine, not bad practice. i’m just not sure we support it tbh.

arvigeus:
According to [this](https://github.com/rollup/rollup/issues/2216#issuecomment-391881386) it is possible if you pass a json object as input. Will try it now.

arvigeus:
Doesn't seem so...

ryanditjia:
I used TSDX to make a lib for my commonly used hooks, mixins, React components, utils here:
https://github.com/ryanditjia/apostrof/tree/master/src

I’d like to be able to import from apostrof/hooks, apostrof/mixins, and so on. But it looks like I can’t use TSDX for this use case?

What other option do I have?

arvigeus:
Rollup support that. TSDX uses it under the hood.

ryanditjia:
When I run yarn build, it generates empty bundles. Am I missing something in package.json?

slikts:
This is the same issue as #276.

Bundling libraries is ultimately a leftover from before automated dependency management or having an automated build step was the norm. A significant downside is that bundled sub-dependencies don't get deduplicated. It's a fairly large oversight in tsdx.

arvigeus:
I will close it now. Moving discussion to #276

agilgur5:
Actually seems like more of a duplicate of #175 & #365 , which are, to an extent, fixed by #367

PR #535, opened 2020-03-04
(feat): add --preserveModules flag

### LINKED ISSUE #276: Support Rollup's preserveModules
### Current Behavior

Currently a single bundle is output by rollup and module structure is not preserved.

### Desired Behavior

There should be a way to do deep imports from the packaged library like so: `import FooModule from "mylib/FooModule"`.

### Suggested Solution

Rollup has a [`preserveModules`][pm] option that includes the module structure in the built files.

A caveat is that the library should be published from the build directory to avoid having to include the build directory in the import path (like `import FooModule from "mylib/dist/FooModule"`).

[pm]: https://rollupjs.org/guide/en/#preservemodules

Here's a simple benchmark comparing deep vs bare imports from a large library: https://github.com/slikts/deep-vs-bare-import-bench

The performance impact of having to parse unused code is non-negligible for larger libraries, and it's compounding. Tree shaking only mitigates this for the final consumer but not while developing. Larger component libraries like MUI specifically support deep imports to avoid this problem.

### LINKED ISSUE #382: Allow disable output dist/index file
### Current Behavior

I have customized the `tsdx.config.js` as such

```
// Not transpiled with TypeScript or Babel, so use plain Es6/Node.js!
module.exports = {
  // This function will run for each entry/format/env combination
  rollup(config, options) {
    config.preserveModules = true;
    config.output.file = undefined;
    config.output.dir = `${__dirname}/${config.output.format}`;
    console.log(config);
    return config; // always return a config.
  },
};
```

Hope it to generate files that match the structure of sources in `${format}` dir .

However, there always is a `dist/index.js` file, which is not desired.

### Desired Behavior

Only `esm` and `cjs` dirs are generated, no `dist` dir.

### Suggested Solution

Support some customization to allow the pattern of keeping the structure of source files.

?

As `cjs` is not for the future, I may disable generating cjs modules to workaround it for some time. But it is not perfect.

agilgur5:
- this works the same way as Rollup's preserveModules option, leaving
  the module structure and not bundling any files together
  - format, env, and minify are used as directories instead of filename
    suffixes since the filename can't be changed without configuring
    output.entryFileNames (possibly a future improvement?)

Fixes #276, #299, #321 .
Might fix #382 , but need to think about what to do with `index.js`.
Related to #367, #175 , #365 

Things to Review / TODOs:
- [ ] add docs
- [ ] declaration files are currently duplicated in each format directory. I'm not sure if this is a problem or not. Changing this requires something like setting `declarationDir` (conflicts/depends on #488) and then not producing types multiple times (conflicts/depends on #367)
- [ ] `dist/index.js` is still created. I'm not sure if it still makes sense to have (related to #382). If it does, its content needs to be modified (conflicts/depends on #367) and maybe its location should be changed to `dist/cjs/index.js`?


agilgur5:
I'm thinking that -- in the future as it would be very breaking -- it might make sense to have TSDX eventually use `outDir` and `outFile` to determine whether to use `preserveModules` or not. That way there's no flags to configure, it just follows whatever you have in your `tsconfig.json`

Issue #91, opened 2019-05-08
Modules resolved absolutely against `src` are treated as external


a-type:
Followup to #89 and #87.

### Current Behavior

Running `tsdx create` initializes a `tsconfig.json` with absolute module resolution relative to the `src` directory. I have VS Code configured to prefer absolute paths if possible when importing modules, so if I have a module like...

`src/foo.ts`
```ts
export const foo = () => 'bar';
```

And I type

```ts
foo()
```

and hit Tab, VS Code will add the following import:

```ts
import { foo } from 'foo';
```

This is valid and properly resolved by Typescript. However, if I compile my library, the current Rollup behavior assumes that `foo` is an external module and doesn't include it in the bundle.

### Expected behavior

I would expect my local modules to be included in my bundle regardless of whether I imported them using 'absolute' resolution as above,

**OR**

I would expect `tsdx` to not initialize `tsconfig.json` with resolution rules which would lead me to believe importing modules relative to `src` is a valid thing to do.

### Suggested solution(s)

Either fix the external modules rules to properly understand local `src`-based resolution, or remove the default `src` resolution rules from `tsconfig.json` to avoid confusing users.

### Additional context

I probably hit this more often than most because I have configured my editor to prefer absolute path resolutions if available.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX             | 0.5.7 
| TypeScript       | 3.4.3
| Browser          | N/A
| npm/Yarn         | NPM 6.2.0
| Operating System | Windows 10

sw-yx:
great issue. we should probably resolve this on a Rollup level. thanks for writing it up. i'll take a look at the rollup config  but am not sure how to solve it

sw-yx:
options:

- (first party) [rollup-plugin-alias](https://github.com/rollup/rollup-plugin-alias)
- [includePaths](https://github.com/dot-build/rollup-plugin-includepaths)
- handrolled https://github.com/rollup/rollup/issues/558
- [rollup-plugin-root-import](https://www.npmjs.com/package/rollup-plugin-root-import)

sw-yx:
probably want to add it [here](https://github.com/palmerhq/tsdx/blob/c2b9421e15c528c3b8aa835549c290693bdc9130/src/createRollupConfig.ts#L73). want to try it in a local version and PR?

jaredpalmer:
I defer to you @sw-yx

a-type:
[...]

Issue #512, opened 2020-02-18
More frequent releases please!


stevesims:
### Current Behavior

As of writing, the current release of tsdx is 0.12.3, which came out over 5 weeks ago, and there has since been 34 new commits, half of which have brought new or revised functionality.  This new functionality, as it has not been released, is not readily accessible to users of tsdx.

### Desired Behavior

More frequent updates please!

### Suggested Solution

Adopt a continuous release strategy.  When code is merged into master (and all tests are passing) a new release should be made.

If current tests are insufficient to ensure that the latest master branch is safe for release, then new tests should be added. 😀 

Some discussion on automated releases can be found in #254 

### Who does this impact? Who is this for?

All users of tsdx.

### Describe alternatives you've considered

I have tried to run tsdx directly from the github repo using `npx github:jaredpalmer/tsdx create package-name` however as of the time of writing this failed with the following error:
```
npm ERR! prepareGitDep src/index.ts(349,11): error TS2367: This condition will always return 'false' since the types '"START" | "BUNDLE_START" | "BUNDLE_END" | "END" | "ERROR"' and '"FATAL"' have no overlap.
npm ERR! prepareGitDep src/index.ts(351,24): error TS2339: Property 'error' does not exist on type 'RollupWatcherEvent'.
npm ERR! prepareGitDep   Property 'error' does not exist on type '{ code: "START"; }'.
```

### Additional context

n/a

agilgur5:
Seems like TSDX isn't alone here: https://github.com/jaredpalmer/razzle/issues/1159 😕 

I've emailed Jared to ask to make me a collaborator here and on NPM or move TSDX to an org. We'll see if there's any progress there.

agilgur5:
So I was added as a collaborator here and on my first day yesterday I closed 35+ issues, triaged, consolidated, labeled, etc, about 80% of open and _closed_ issues, and closed out ~15 PRs.

Did some more triage today and merging of PRs, but unfortunately I wasn't yet added as a collaborator on NPM, so I can't yet make releases.

That being said, I did merge #544 , so you should now be able to install from GitHub without the error listed above.

jaredpalmer:
Sorry I have been a bit MIA. Have been cranking  on Formik Inc. now full time and haven’t been on GitHub much. Similarly Shawn (@swyx)  is crazy busy as well. Giving contributor and publish rights to @agligur who has done an amazing job with this project over the last several months

MikelArnaiz:
[...]

PR #540, opened 2020-03-05
(fix): don't treat @/ ~/ alias as external module

rockmandash:
### Current Behavior

Treating none relative import as external.

### Desired Behavior

I know TSDX absolute import is under [discussion](https://github.com/jaredpalmer/tsdx/issues/91) right now.

But can we start supporting `@/` and `~/` as an alias of srcDir?

It's widely used in Nuxt community, see [documentation](https://nuxtjs.org/guide/directory-structure/#aliases)

### Suggested Solution

Don't treat import path that startsWith `@/` and `~/` as external.

So people can write their own absolute import plugin like this:

```js
function rollupPluginAbsoluteImport() {
  return {
    name: "rollup-plugin-absolute-import",
    resolveId(source: string) {
      if (source.startsWith("@/") || source.startsWith("~/")) {
        return path.resolve(paths.appSrc, "." + source.slice(1, source.length));
      }
      return null;
    }
  };
}
```

All users who can't wait to use absolute import path

There is no way to use the absolute import path for TSDX right now I think.

I tried a lot of rollup plugins...

### Additional context

Maybe release more often?

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -56,6 +56,12 @@ export async function createRollupConfig(
       if (id === 'babel-plugin-transform-async-to-promises/helpers') {
         return false;
       }
+      /**
+       * ~/ or @/ for srcDir
+       */
+      if (id.startsWith('@/') || id.startsWith('~/')) {
+        return false;
+      }
       return external(id);
     },
     // Rollup has treeshaking by default, but we can optimize it further...


agilgur5:
A few things:

1. This is a feature, not a bugfix
2. There is a way to do aliases with `babel-plugin-module-resolver`, see this HOWTO: https://github.com/jaredpalmer/tsdx/issues/379#issuecomment-568202209
3. I think I could get behind a default using `@`, though this might need support in Parcel and Storybook examples for consistency. `~` I think is very confusing since that refers to homedir in unix.
4. RE: frequency of releases: #512 . I think you already up-voted it

PR #546, opened 2020-03-07
(env/refactor): use eslintrc ignorePattern instead of CLI arg

agilgur5:
just an internal change


agilgur5:
Ah, I see the lint tests fail since they use the same `.eslintrc` 😅 😅 . Wonder if there's a way to do configure with `.eslintrc` without affecting the tests... 🤔

agilgur5:
If we add a blank `eslintrc` into the lint folder that _might_ fix it from looking up a directory

PR #555, opened 2020-03-10
(fix): correctly read tsconfig esModuleInterop

### LINKED ISSUE #469: Esmodule options in rollup output is always undefined
Hi,

While working on #468 I noticed [this line of code](https://github.com/jaredpalmer/tsdx/blob/master/src/createRollupConfig.ts#L92)

```js
esModule: tsconfigJSON ? tsconfigJSON.esModuleInterop : false,
```

But `esModuleInterop` is a compiler option so I guess it should be 

```js
esModule: Boolean(tsconfigJSON?.compilerOptions?.esModuleInterop)
```

I can create a PR to fix this I just want to be sure that I'm not missing something here...

agilgur5:
- it's a property of compilerOptions, not of overall tsconfig
  - i.e. tsconfig.compilerOptions.esModuleInterop

- because this was incorrectly read, Rollup's esModule would always
  be set to `undefined`, and Rollup would then default to `true`
  - so esModuleInterop still wasn't being respected properly, I just
    shifted the defaults when I incorrectly patched this :/

(test): add tests for the default tsconfig as well as for when it's
explicitly set to false

Fixes #469 . Actually this was unintentionally fixed by me a short while ago when I updated #489 , but this improves the syntax and adds tests

This is a follow-up on bugs in #327

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -95,7 +95,7 @@ export async function createRollupConfig(
       // (i.e. import * as namespaceImportObject from...) that are accessed dynamically.
       freeze: false,
       // Respect tsconfig esModuleInterop when setting __esModule.
-      esModule: tsCompilerOptions ? tsCompilerOptions.esModuleInterop : false,
+      esModule: Boolean(tsCompilerOptions?.esModuleInterop),
       name: opts.name || safeVariableName(opts.name),
       sourcemap: true,
       globals: { react: 'React', 'react-native': 'ReactNative' },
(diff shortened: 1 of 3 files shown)


agilgur5:
Ok, the test failures here may have revealed an issue with `dist/index.js` -- it itself doesn't have `__esModule` set 😕 

So I'm not totally sure if this works properly or not. The test output is very weird, because it even gets `true` when there's nothing set in either `dist/index.js` or in the actual CJS files...

But no one has complained about it being wrong since #367, nay, I even tested and use this in my own libraries... And I just confirmed the output is correct again in the default case...

This might just be an artifact with how Jest loads these things 😕 🤷‍♂

PR #582, opened 2020-03-14
Add publish npm script

jaredpalmer:
Add np to devDeps

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "lint:post-build": "node dist/index.js lint src test --ignore-pattern 'test/tests/lint'",
     "test": "yarn build && yarn test:post-build",
     "test:post-build": "node dist/index.js test --config ./test/jest.config.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,7 +27,8 @@
     "lint:post-build": "node dist/index.js lint src test --ignore-pattern 'test/tests/lint'",
     "test": "yarn build && yarn test:post-build",
     "test:post-build": "node dist/index.js test --config ./test/jest.config.json",
-    "start": "tsc -p tsconfig.json --watch"
+    "start": "tsc -p tsconfig.json --watch",
+    "publish": "np"
```
agilgur5: `publish` is a lifecycle hook that's [actually equivalent to `postpublish`](https://docs.npmjs.com/misc/scripts). `np`'s docs say to use something like [`release`](https://github.com/sindresorhus/np#release-script). I've used `pub` in the past too.

PR #618, opened 2020-03-17
(refactor): use cacheRoot path constant

agilgur5:
- and put progressEstimatorCache into the TSDX cache dir too

Noticed in #249 that this was used before my #329 , so should standardize usage of both


agilgur5:
Closing this since `cacheRoot` was removed/set to the default in #691

PR #621, opened 2020-03-17
(refactor): split build tests into separate files per fixture

agilgur5:
- better organized that way as the tests are somewhat independent of
  one another and previously even I was confused where to put different
  types of tests

- also refactor some test naming and some redundant usage of flags
  - especially for 0-config, where flags shouldn't be used

- and remove testEnvironment comment as it's already set in jest config (this is why #405 didn't make a big difference)

(optim): now that it's one fixture per file, we only need to copy the
fixture directory once per file, which should be a speed boost
- and sometimes may not need to re-run `tsdx build` when not necessary
- also the different test files get parallelized, which is also faster

- before: 53s avg on my machine, after: 37s avg on my machine
  - avg for running all tests
  - pretty BIG difference

Been meaning to do this for a while (as well as some other test optimizations still in-progress), and now that there's more and more fixtures (e.g. #468, #525) and more and more tests, I think now is a good time to do this


test/tests/tsdx-build-default.test.js
```diff
@@ -0,0 +1,75 @@
...
+      shell.test('-f', 'dist/build-default.cjs.production.min.js')
+    ).toBeTruthy();
+    expect(shell.test('-f', 'dist/build-default.esm.js')).toBeTruthy();
+
+    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+
+    expect(output.code).toBe(0);
+  });
+
+  it('should create the library correctly', () => {
```
agilgur5: ok, so skipping the `shell.exec` build step is actually problematic -- it means that you can't comment out some of the tests or use `it.only` because order matters now... right, I kind of missed that.

So a better option would be to create a sort of caching function that checks if the same build were just run and skip it if so, and if not, run it. So if things get commented out or `it.only` is used everything still runs properly and order only affects the speed/cache, not the actual output.

agilgur5: Now that `util.setupStageWithFixture` is in a `beforeAll` we could move the first `shell.exec` into there too or put all tests that re-use the same `shell.exec` into one nested `describe` with a `beforeAll`, which is probably more correct.

Ideally it should remake the whole fixture and exec per test, but that makes tests much slower, which has been hampering me in dev-test cycles a lot 😕

agilgur5: decided to create a naive `execWithCache` function as mentioned above. main problem with `beforeAll` is that `output` must then be a `describe`-scoped variable, which is sub-optimal testing practice

PR #625, opened 2020-03-19
(ci/optim): add caching for yarn install

agilgur5:
- via [bahmutov/npm-install](https://github.com/bahmutov/npm-install)
- it handles all the cache steps, yarn vs. npm, OS differences, etc

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -17,8 +17,8 @@ jobs:
         with:
           node-version: 10.x
 
-      - name: Install deps and build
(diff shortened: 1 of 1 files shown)


agilgur5:
So it's at least twice as fast on the Ubuntu runs, which is great!

But it fails on Windows and macOS and is probably a bit slower as a result. Filed https://github.com/bahmutov/npm-install/issues/24

agilgur5:
Merging as some caching is still better than no caching. And any bugfixes will probably apply without needing to make any changes. If changes need to be made, can improve on this at a later date.

Issue #543, opened 2020-03-06
Doesn't play well with babel-plugin-styled-components -- plugin order matters


thany:
### Maintainer Update

For any new readers of this issue, per https://github.com/formium/tsdx/issues/543#issuecomment-694561305, you can use `styled-components/macro` instead of `babel-plugin-styled-components`, which will work.

<hr>

### Current Behavior

I'm using `babel-plugin-styled-components` to annotate styled-components with CSS classes to more easily debug them in the browser's inspector.

### Expected behavior

I'd expect this plugin to do something. Namely, do what it says on the tin. *And it does* when I run storybook for our project. But it doesn't when I'm done and want to build my project through tsdx.

### Suggested solution(s)

Perhaps babel plugins ought to be merged differently. I suspect the plugins from `.babelrc` are executed too early or too late. I find it massively complicated to figure out what exactly is happening under the hood, and therefor I have no way of knowing how and when a babel plugin is executed, or supposed to be executed. But since the plugin itself does work, there must be an issue with how or when it's being called.

### Additional context

Our project is based on React + styled-components in Storybook. In Storybook, babel-plugin-styled-components does it job beautifully. But in the tsdx build, it does not, and indeed, I cannot find any mention of the classnames that are supposed to be generated. Having or not having this plugin, literally makes not a byte of difference in the dist files that tsdx generates.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.12.3
| TypeScript       | 3.8.3
| Browser          | Firefox 73
| npm/Yarn         | npm 6.9.0
| Node             | 10.16.3
| Operating System | Windows 10

agilgur5:
The Babel plugins that are used internally are here: https://github.com/jaredpalmer/tsdx/blob/3a6d42fb767150233c651056aabda9dd91bd048d/src/babelPluginTsdx.ts#L64-L102

It might be that the Rollup plugins that are run prior change some things too. Are you using a custom `tsdx.config.js`? There wasn't a repro provided.

<hr>

Also please ignore the reference above, that was an accident as I was trying to reference two PRs I made back-to-back but this issue was created in-between

thany:
Sorry, no there's no custom `tsdx.config.js`. Only a `.babelrc` that literally only contains the plugin I mentioned:

```
{
  "plugins": [
    "babel-plugin-styled-components"
  ]
}
```
[...]

PR #627, opened 2020-03-19
Add Integration Tests

### LINKED ISSUE #481: create test setup for custom Babel configs
### Current Behavior

this is a followup to https://github.com/jaredpalmer/tsdx/pull/473 - we want to support this so they should be part of our tests.

agilgur5:
This is quite a big PR as it refactors a bunch of the test set-up to handle two different types of tests, as well as adds integration tests for `tiny-invariant`, `rollup-plugin-postcss`, `styled-components`, and `regeneratorRuntime`. Fixes #481 

I initially tried to have the integration tests have their own `package.json` and `node_modules`, as the `build-withConfig` folder had before (and with a new CI workflow with `working-directory` set to the integration test dir), but this proved extremely confusing and _very_ time-consuming to deal with.
It basically requires that scripts, like tests, run from the integration test dir, as if they run from root, they won't read the `node_modules` (node scans upward, so from the integration test dir, it'll get the integration deps and the TSDX deps, but the opposite doesn't work). And trying to run scripts from there causes lots of inconsistencies all over the place, so I just scrapped that idea.
Open to ideas on better integration testing structures, but the other option proved completely not worth the time, so don't want to spend a ton more time trying to get something else to work either.

Also found some integration failures all over the place while adding this, welp 😕 These include:
- `--extractErrors` removes `invariant` but doesn't replace it with `ErrorProd` or `ErrorDev`. And it leaves the import in, just importing nothing 😕 
  - `warning` isn't changed what-so-ever
  - I still think this should be split out as its own Rollup or Babel plugin, it's fairly independent of TSDX, doesn't seem to be used much by users, and can benefit more than just TSDX
- `rollup-plugin-postcss` any PostCSS plugins like `autoprefixer` and `cssnano` work, but the name of the extracted CSS file is the same as the output, i.e. `package-name.cjs.development.css`
- `babel-plugin-styled-components` doesn't work, plugin order matters for it. See #543


agilgur5:
Ok fixing the conflicts with #582 in `yarn.lock` was not easy 😕 . But done now, so this is ready to merge. Should go in before #525 , so need to review this soon

test/integration/fixtures/build-withBabel/src/styled.tsx
```diff
@@ -0,0 +1,12 @@
...
+import styled from 'styled-components';
+
+const Title = styled.h1`
+  /* this comment should be removed */
+  font-size: 1.5em;
+  text-align: center;
+  color: palevioletred;
+`;
+
+ReactDOM.render(<Title />, document.getElementById('test'));
```
agilgur5: Think I might just be able to `export const Title` and get rid of the `@types/react-dom` dependency (though not `react-dom` itself since it's a dependency of `styled-components)

test/integration/tsdx-build-withBabel.test.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should convert styled-components template tags', () => {
+    let output = shell.exec('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    // from styled.h1` to styled.h1(
+    output = shell.grep(/styled.h1\(/, ['dist/build-withbabel.*.js']);
+    expect(output.code).toBe(0);
+  });
```
agilgur5: well this isn't exactly a babel integration test, only the `babel-plugin-styled-components` one is, but I guess it's also a pre-req... maybe I should merge this with the test below?

agilgur5: Or, I guess template tags are babel since it's a newer syntax. Will leave it as is, nbd either way.

test/integration/fixtures/build-withBabel/src/styled.tsx
agilgur5: yeppp that worked, also had to `export { Title }` in `index.ts` as it doesn't have side-effects anymore without the render

test/util.js
```diff
@@ -0,0 +1,41 @@
...
+// @param {RegExp} pattern
+// @param {string} fileName
+// @returns {boolean} true if pattern has matches in file
+function grep(pattern, fileName) {
+  const output = shell.grep(pattern, fileName);
+  // output.code is always 0 regardless of matched/unmatched patterns
+  // so need to test output.stdout
+  // https://github.com/jaredpalmer/tsdx/pull/525#discussion_r395571779
+  return Boolean(output.stdout.match(pattern));
+}
```
agilgur5: ehhhh this should probably be in a separate file because all the functions here are solely fixture related. This file was originally in `fixtures/util.js` so maybe it should be in `utils/fixture.js` instead and this in a more generic `utils/misc.js` or something.

agilgur5: yep this was put into the same `utils/shell.js` from #621's recent changes
and `test/util.js` was moved to `utils/fixture.js`

agilgur5:
- Rebased with #621 and added `execWithCache` everywhere like it has. This actually took a bit of effort due to the number of commits and renames to rebase against.
- Moved `grep` helper to `utils/shell.js`
- Moved `util` to `utils/fixture.js` (also took some effort due to renames / rewriting imports)
- Removed unused `utils/psTree.js` and its dependencies
- Fixed a rebase issue I made where I accidentally merged two commits together that should be split
- Added a test that ensures Babel presets are merged, which was the original goal of #481, to add a regression test for the fix in #473

PR #644, opened 2020-03-27
(fix): change plugin order to make styled-components/macro work

agilgur5:
- if babel-plugin-macros is first, macros get applied first
  - babel-plugin-styled-components says it should be placed first, but
    unfortunately all user plugins are added after TSDX plugins
    - unknown how changing this ordering could impact lots of code out
      there, but could be very breaking.
      - moving to a preset instead would mean this is more in user
        control
  - but we can change it so macros are added first and so one can use
    styled-components/macro instead as a workaround

- based on babel-plugin-styled-component's code, docs, and output in
  detail, I think its ordering conflict may be with
  babel-plugin-annotate-pure-calls, which so happens to be the first
  plugin in babelPluginTsdx
  - maybe this should be last given that other plugins can change
    functions etc

(test): update styled-component template tag test to reflect the
slightly different tag due to the usage of the macro

(fix/test): comment removal should use toBeFalsy, not toBeTruthy,
since it's removed
- this was a bug I introduced when adding the grep helper; just added
  it the same everywhere but this was the one place that was testing
  to get an error code
  - fix the comment so it doesn't say error code anymore either
  - since this test was skipped, I didn't pick up that it was wrong
    until it ran now

Workaround for #543 

Should go out in a minor since it's possible changing the order of this one plugin could break some things, but highly unlikely

### DIFF
--- a/test/integration/tsdx-build-withBabel.test.ts
+++ b/test/integration/tsdx-build-withBabel.test.ts
@@ -19,19 +19,22 @@ describe('integration :: tsdx build :: .babelrc.js', () => {
     const output = execWithCache('node ../dist/index.js build');
     expect(output.code).toBe(0);
 
-    // from styled.h1` to styled.h1(
-    const matched = grep(/styled.h1\(/, ['dist/build-withbabel.*.js']);
+    // from styled.h1` to styled.h1.withConfig(
+    const matched = grep(/styled.h1.withConfig\(/, [
+      'dist/build-withbabel.*.js',
+    ]);
     expect(matched).toBeTruthy();
   });
 
-  // TODO: make this test work by allowing customization of plugin order
-  it.skip('should remove comments in the CSS', () => {
+  // TODO: make styled-components work with its Babel plugin and not just its
+  // macro by allowing customization of plugin order
+  it('should remove comments in the CSS', () => {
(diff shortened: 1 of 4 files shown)


zebapy:
Is this going to get merged? What's holding it back?

agilgur5:
@zebapy per the PR description, label, and milestone, it's a breaking change and will go out in v0.14.0

test/integration/tsdx-build-withBabel.test.ts
```diff
@@ -19,19 +19,22 @@ describe('integration :: tsdx build :: .babelrc.js', () => {
...
+    // from styled.h1` to styled.h1.withConfig(
+    const matched = grep(/styled.h1.withConfig\(/, [
+      'dist/build-withbabel.*.js',
+    ]);
     expect(matched).toBeTruthy();
   });
 
-  // TODO: make this test work by allowing customization of plugin order
-  it.skip('should remove comments in the CSS', () => {
+  // TODO: make this test work with the babel-plugin and not just the macro
```
agilgur5, line 29: Babel plugin for consistency

agilgur5 (commented):
One nit, otherwise seems good to go if all tests pass

PR #226, opened 2019-10-05
Add "test" to include key in template tsconfigs

third774:
Fixes https://github.com/jaredpalmer/tsdx/issues/225

Issue #225, opened 2019-10-05
tsconfig.json not applied to test files


third774:
### Current Behavior

Currently the `tsconfig.json` file only lists `["src", "types"]` in the `include` key, which leaves out the `"test"` folder. This means that VS Code doesn't use the `tsconfig.json` file for files in that directory. None of the `strict` rules are applied.

![image](https://user-images.githubusercontent.com/8732191/66255623-713e9100-e74b-11e9-85bb-a413878e63e4.png)

### Expected behavior

Expected the `tsconfig.json` rules to also apply to test files

### Suggested solution(s)

Add `"test"` to the `include` key

![image](https://user-images.githubusercontent.com/8732191/66255644-bf539480-e74b-11e9-8f2d-71359d3bcd46.png)

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | 0.9.3
| TypeScript       | 3.6.2
| Browser          | n/a
| npm/Yarn         | 1.12.3
| Operating System | macOS Mojave 10.14.2

agilgur5:
Turns out this was a duplicate of #84

Issue #84, opened 2019-05-05
vscode doesn't use tsconfig.json settings for files in test folder


xaviergonz:
### Current Behavior

When I set experimentalDecorators to true in tsconfig.json it works ok for files in "src", but it doesn't work for files in "test" (vscode marks decorators in red, but it works ok when running 'test')

### Desired Behavior

When experimentalDecorators is set to true in tsconfig.json it should also be for vscode test files

### Suggested Solution

Add a tsconfig.json file to the tests folder:
 ```
{
  "extends": "../tsconfig.json",
  "include": ["."]
}
```
with any other overrides that might be needed to make tests work and make jest use that one.

### Who does this impact? Who is this for?

Libs that use experimental decorators

### Describe alternatives you've considered

N/A

### Additional context

N/A

**Comment by xaviergonz** [comment]
Ok, aparently as of v0.5.7 decorators do work for tests and builds, but vscode complains when they are used inside tests

**Comment by xaviergonz** [comment]
One way I found to fix this was including the following tsconfig.json inside the tests folder
```
{
  "extends": "../tsconfig.json",
  "include": ["."]
}
```

agilgur5:
I think #225 ended up duplicating this and #226 fixed it in the templates. Though this is a great workaround

PR #646, opened 2020-03-27
(fix): remove faulty tsconfig.json include of test dir

### LINKED ISSUE #638: test/ dir is in `include`, non-spec TS files (e.g. test utils) break build with v0.13 rootDir: `./src`
### Current Behavior

Prior to v0.13 #504 , I could put a file into my test folder that contained utility code for tests and it would get compiled without a problem.

`test/TestUtils.tsx`

Now, when I `tsdx build`, I see this error:

`rpt2: options error TS6059: File '/mui-rff/test/TestUtils.tsx' is not under 'rootDir' '/mui-rff/src'. 'rootDir' is expected to contain all source files.`

### Expected behavior

I have a place to put utility code for tests.

### Suggested solution(s)

No idea, but you have a list of 3rd parties using this project and it would be nice if you would start to test future versions of tsdx against these projects to see if you break peoples stuff and can then use that to find out how to tell people how to fix their broken stuff.

### Your environment

| TSDX           | v0.13 |
| TypeScript       | 3.8.3 |
| npm/Yarn         |  6.13.7 / 1.22.4 | 
| Node             |  v13.11.0 |
| Operating System | OSX Catalina |

agilgur5:
- `include` means all files that should be compiled, and the test dir
  should definitely not be compiled
  - this was causing issues where declarations were being output into
    dist/ for all files in test/
    - that shouldn't happen and it means `build` was unnecessarily
      slowed down in order to process test/ files
    - default `exclude` that was added of *.spec/*.test files,
      originally for co-located tests, ended up making most test/ dir
      files excluded too, but not, say, test utils that don't have a
      .spec or .test suffix

- this was also causing errors with the `rootDir: './src'` change as
  the test/ dir is outside of the rootDir
  - problem wasn't the rootDir, but that the test/ dir shouldn't be
    compiled or processed in any way

(test): ensure test/*.test.ts and test/*.ts files are correctly
excluded and don't have declarations generated
- and that they don't error with rootDir: './src'

(test): ensure types/*.d.ts don't error with rootDir and don't have
declarations re-generated either
- n.b. tsc won't re-generate them either, they don't get copied to
  outDir, this isn't new or different behavior

(env/test): Jest shouldn't run / should ignore tests inside of fixtures

<hr>

Fixes #638 , the root cause of it, and adds tests for this.

Effectively reverts #226 . I haven't been able to reproduce #225 / #84 in my VS Code, but maybe that's because I have it configured a little differently?
Would like to add VS Code integration tests so there's an automated way to check that it works independently on CI without any configuration (and so I don't have to attempt and fail to test this on my own machine).
But `tsc --noEmit` at least, indeed doesn't output errors in `test/` anymore. `tsdx lint` still does.
Per my comments in #638 if this and the VS Code integration are problematic, we should move this to a `tsconfig.build.json` and then have a plain `tsconfig.json` that extends from it and changes `include` to `'./'` so everything is type-checked.

### DIFF
--- a/test/e2e/tsdx-build-default.test.js
+++ b/test/e2e/tsdx-build-default.test.js
@@ -31,6 +31,15 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(output.code).toBe(0);
   });
 
+  it("shouldn't compile files in test/ or types/", () => {
+    const output = execWithCache('node ../dist/index.js build');
+
+    expect(shell.test('-d', 'dist/test/')).toBeFalsy();
+    expect(shell.test('-d', 'dist/types/')).toBeFalsy();
+
+    expect(output.code).toBe(0);
+  });
+
   it('should create the library correctly', () => {
     const output = execWithCache('node ../dist/index.js build');
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,5 +4,7 @@ module.exports = {
   testPathIgnorePatterns: [
     '/node_modules/', // default
     '<rootDir>/templates/', // don't run tests in the templates
+    '<rootDir>/test/.*/fixtures/', // don't run tests in fixtures
+    '<rootDir>/stage-.*/', // don't run tests in auto-generated (and auto-removed) test dirs
   ],
 };
--- a/src/deprecated.ts
+++ b/src/deprecated.ts
@@ -22,7 +22,9 @@ export async function moveTypes() {
       'rootDir to "./src".\n' +
       'TSDX has deprecated setting tsconfig.compilerOptions.rootDir to ' +
       '"./" as it caused buggy output for declarationMaps and occassionally ' +
-      'for type declarations themselves.'
+      'for type declarations themselves.\n' +
+      'You may also need to change your include to remove "test", which also ' +
+      'caused declarations to be unnecessarily created for test files.'
   );
 
   try {
--- a/templates/basic/tsconfig.json
+++ b/templates/basic/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src", "types", "test"],
+  "include": ["src", "types"],
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
--- a/templates/react-with-storybook/tsconfig.json
+++ b/templates/react-with-storybook/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src", "types", "test"],
+  "include": ["src", "types"],
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
--- a/templates/react/tsconfig.json
+++ b/templates/react/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src", "types", "test"],
+  "include": ["src", "types"],
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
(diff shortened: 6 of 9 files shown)


test/e2e/tsdx-build-default.test.js
```diff
@@ -31,6 +31,15 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(output.code).toBe(0);
   });
 
+  it("shouldn't compile files in test/ or types/", () => {
+    const output = execWithCache('node ../dist/index.js build');
+
+    expect(shell.test('-f', 'dist/test/')).toBeFalsy();
+    expect(shell.test('-f', 'dist/types/')).toBeFalsy();
```
agilgur5: This should be [`-d` or `-e`](https://github.com/shelljs/shelljs#testexpression), not `-f`, which specifically checks for a file. I could test for the specific files too, but would prefer to keep it generic, as the directories shouldn't exist at all.

agilgur5: used `-d` and tested to make sure it does give a test error when `test/` is erroneously in `include`.

kylemh:
Does this mean we'd also wanna avoid compiling stories?

agilgur5:
@kylemh yes, your stories shouldn't be in your `include`, otherwise they'll be unnecessarily compiled (and output to `dist`/likely unnecessarily published).

The Storybook template doesn't have them in the `include` as you can see here. Are yours by chance co-located? We could add `**/*.stories.tsx` to the default `exclude` if it's a common enough pattern (I've seen it a few times).

The approach I listed in https://github.com/jaredpalmer/tsdx/issues/638#issuecomment-605728767, especially with `file: ['src/index.ts']` is optimal, but unfortunately seems to have issues in upstream rpts2 as a I mentioned there.

kylemh:
So I’m doing monorepo shit in a personal project first before I try to come up with a template to resolve #122

https://github.com/innocuous-tech/core

I’ve got it all flying for now, but I can’t see how to prevent the stories’ typings from getting compiled into the bundle.

I have it excluded... what do you think?

agilgur5:
Ah yours are like semi-colocated, in the `src` dir but as separate dirs. I'd leave them outside of `src` for a template fwiw.
Probably just need to mess with the globs a bit. If you moved the `exclude` into each package it would probably be easier to figure out. Path resolution is a bit weird with `extends` too but you seem to be using it correctly so I'm not sure.

kylemh:
Hmmmm I remember reading somewhere that @jaredpalmer prefers colocating tests with source code. Having the answer be “don’t do that” doesn’t seem ideal.

I’ll try manually listing the excludes in each package, but that wasn’t working either.

agilgur5:
Well none of the existing templates have co-located tests or stories, so it wouldn't make sense to add a monorepo template that is the odd one out has semi-colocated tests. Would be very inconsistent and confusing to users

kylemh:
So this is my individual use case. I can make the template match the rest. The point I’m making is that we should be able to have collocated tests and stories.

In my example, test files are ignored but not stories, so I was just tryna make sure it wasn’t something under the hood. Not extending and excluding per package seems to be working. I’ll have to figure out where I’m going wrong with `extends`

agilgur5:
The `include`/`exclude` handling should work the same as `tsc`, the only difference is that TSDX a default `exclude` for fully co-located tests (if you're adding your own it doesn't matter). Though it's possible there are bugs with rpts2 as that's one reason why I haven't moved templates to just a `files` set-up which wouldn't be impacted by location at all, only by imports.

PR #649, opened 2020-03-28
(refactor): migrate all tests and test helper code to TS

agilgur5:
- woot, all TS now!
  - get that big blue line for language TS on GitHub haha

- for the most part, not much more to do other than .js -> .ts
  - for the utils had to add some types around

- also use ESM instead of CJS in tests and test helpers
  - require -> import
  - module.exports -> export
    - the indentation change here with utils/fixture made git treat
      it as a delete + new instead of a rename

<hr>

This was, of course, done because #627 added lots of tests, and definitely not because TSDX became majority JavaScript after #627. This was definitely not because the majority JavaScript made TSDX classified as a "JavaScript" library 
Screen Shot 2020-03-27 at 2 10 25 AM

(it actually wasn't, I always wanted to convert the tests to TS, especially after the addition of the `grep` helper with JSDoc types in #525 (later moved to #627 which got merged first). but I can't say that didn't increase priority 😉 ).


agilgur5:
Screen Shot 2020-03-29 at 12 05 33 AM
That's better. Although I still don't think it's totally accurate

PR #666, opened 2020-04-07
(feat): exclude co-located stories by default from build

### LINKED ISSUE #663: tsconfig `exlude` not adhered too in subdirectories of a monorepo
### Current Behavior

When using a monorepo, I'd like to minify the amount `tsconfig` I am defining per package. To that end, the value of the `exclude` property at the root config is being ignored when building TSDX packages in their own subdirectories.

You can test for yourself thru [this example project](https://github.com/innocuous-tech/core) by doing:
```sh
yarn install
yarn workspace @innocuous/components add @innocuous/hooks
yarn workspace @innocuous/hooks run build
yarn workspace @innocuous/components run build
```

and noticing that storybook files are compiled, despite being ignored from the root config extended by the package's config.

It seems like running `tsc` with my tsconfigs works as expected, but the build step continues to compile things I am excluding.

### Expected behavior

In the example repo, stories are not compiled.

### Suggested solution(s)

Unsure... I'd love to resolve this myself. @agilgur5 has DMs disabled on Twitter, but I was curious if anybody would be down to pair for a short chat about how files/folders are chosen for the build step in `tsdx`.

### Additional context

Similar issue resolved: https://github.com/jaredpalmer/tsdx/issues/61

### Your environment

| TSDX           | latest
| TypeScript       | latest
| Browser          | Chrome latest
| npm/Yarn         | yarn latest
| Node             | node latest
| Operating System | MacOS latest

kylemh:
A brute attempt to resolve the issue I'm encountering in #663

Resolves #663

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -154,6 +154,8 @@ export async function createRollupConfig(
             '**/*.test.ts',
             '**/*.spec.tsx',
             '**/*.test.tsx',
+            // storybook files
+            '**/*.stories.tsx',
             // TS defaults below
             'node_modules',
             'bower_components',


src/createRollupConfig.ts
```diff
@@ -154,6 +154,8 @@ export async function createRollupConfig(
             '**/*.test.ts',
             '**/*.spec.tsx',
             '**/*.test.tsx',
+            // storybook files
+            '**/*.stories.tsx',
```
kylemh, line 158: I wonder if we need to do `...(!!opts.tsconfig.exclude ? [opts.tsconfig.exclude] : [])`

kylemh: I tested the [logic](https://github.com/ezolenko/rollup-plugin-typescript2/blob/master/src/parse-tsconfig.ts#L44) in rpts2:
Screen Shot 2020-04-07 at 1 53 08 AM

I don't see an issue... it should work fine.

Although strangely enough a different merge loses some options which I'm surprised get lost...
Screen Shot 2020-04-07 at 1 56 39 AM

kylemh: There exists a few lines after L44 in that rpts2 file I linked which I believe also do some merging.

agilgur5: That's not because of rpts2 -- that's how `_.merge` works with arrays. `exclude` is an array, so your custom exclude would just override the same indices.
In your first "test", `**/*.stories.tsx` also overrides the first entry, `**/*.spec.ts`

I think this jogged my brain a bit -- I'm pretty sure that line in rpts2 is around where the bug is.
TSDX has similar logic as rpts2 using TS's custom parsers to read a `tsconfig` and then get its `compilerOptions` through all `extends` chains. But that's only for `compilerOptions`, meaning `exclude` etc from an `extends` aren't actually parsed.

Problem there is that I don't know if there's a way to get the rest. This part of the TS API isn't documented but is used in lots of places in the wild for parsing, it took me a while to figure out how to do it myself, c.f. #489

agilgur5: Ah, the original config gets mutated with the new `include`/`exclude`, while `compilerOptions` is unchanged and returned per https://github.com/facebook/create-react-app/pull/5537

I'll have to look into rpts2 config and see if there's actually a bug or not, I'm not sure. I still haven't repro'd #663

kylemh:
Sounds like I should just wait for https://github.com/ezolenko/rollup-plugin-typescript2/issues/211

Thanks for diving deep! Love having you on this project. ❤️

agilgur5:
> Sounds like I should just wait for [ezolenko/rollup-plugin-typescript2#211](https://github.com/ezolenko/rollup-plugin-typescript2/issues/211)

I don't think that'll be fixed any time soon, especially not if it requires parsing declarations.

Hence why I'd be fine with this going in, as, for the time-being we have to use `include` instead of `file` (even if we changed to the latter, we'd still have older users of the former).

kylemh:
I see. My guess is a relevant test would be considered e2e? I can do a default build with a Storybook file. Make a uniquely named type, and assert that none of the file contents in `dist` include the type name?

agilgur5:
Yea you'd like create `e2e/fixtures/build-default/src/story-test.stories.tsx` that's empty other than a comment (see the `build-default/test/` for what that would look like) and then use the same style tests as in the `e2e/tsdx-build-default.test.ts` except you'd have a `toBeFalsy()` for this file instead of `toBeTruthy()`

PR #669, opened 2020-04-08
(fix): fix output file name after build

kotarella1110:
Fixed output file name after build to use the name of `package.json` instead of the passed name if the `--name` option is passed.

Currently, the output file name uses the value passed by the `--name` option, as follows:

1. Create a library using hyphen: `npx tsdx create my-lib`
2. Build with `--name` option: `yarn build --name MyLib --format esm,cjs,umd`
3. Check `dist` dir: `tree dist/`

```
dist/
├── index.d.ts
├── index.js
├── mylib.cjs.development.js
├── mylib.cjs.development.js.map
├── mylib.cjs.production.min.js
├── mylib.cjs.production.min.js.map
├── mylib.esm.js
├── mylib.esm.js.map
├── mylib.umd.development.js
├── mylib.umd.development.js.map
├── mylib.umd.production.min.js
└── mylib.umd.production.min.js.map
```

I think the output file name should use the name of `package.json` instead of the name passed with the `--name` option, as follows:

```
dist/
├── index.d.ts
├── index.js
├── my-lib.cjs.development.js
├── my-lib.cjs.development.js.map
├── my-lib.cjs.production.min.js
├── my-lib.cjs.production.min.js.map
├── my-lib.esm.js
├── my-lib.esm.js.map
├── my-lib.umd.development.js
├── my-lib.umd.development.js.map
├── my-lib.umd.production.min.js
└── my-lib.umd.production.min.js.map
```

In fact, in ReactDOM, it is `react-dom.development.js`, not` reactdom.development.js`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -290,10 +290,9 @@ prog
     if (!opts.noClean) {
       await cleanDistFolder();
     }
-    opts.name = opts.name || appPackageJson.name;
     opts.input = await getInputs(opts.entry, appPackageJson.source);
     if (opts.format.includes('cjs')) {
-      await writeCjsEntryFile(opts.name);
+      await writeCjsEntryFile(opts.outputName);
     }
 
     type Killer = execa.ExecaChildProcess | null;
@@ -393,7 +392,7 @@ prog
     await cleanDistFolder();
     const logger = await createProgressEstimator();
     if (opts.format.includes('cjs')) {
-      const promise = writeCjsEntryFile(opts.name).catch(logError);
+      const promise = writeCjsEntryFile(opts.outputName).catch(logError);
       logger(promise, 'Creating entry file');
     }
     try {
@@ -421,6 +420,7 @@ async function normalizeOpts(opts: WatchOpts): Promise<NormalizedOpts> {
   return {
     ...opts,
     name: opts.name || appPackageJson.name,
+    outputName: appPackageJson.name,
     input: await getInputs(opts.entry, appPackageJson.source),
     format: opts.format.split(',').map((format: string) => {
       if (format === 'es') {
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -35,7 +35,7 @@ export async function createRollupConfig(
     opts.minify !== undefined ? opts.minify : opts.env === 'production';
 
   const outputName = [
-    `${paths.appDist}/${safePackageName(opts.name)}`,
+    `${paths.appDist}/${safePackageName(opts.outputName)}`,
     opts.format,
     opts.env,
     shouldMinify ? 'min' : '',
--- a/src/types.ts
+++ b/src/types.ts
@@ -28,6 +28,7 @@ export interface WatchOpts extends BuildOpts {
 export interface NormalizedOpts
   extends Omit<WatchOpts, 'name' | 'input' | 'format'> {
   name: string;
+  outputName: string;
   input: string[];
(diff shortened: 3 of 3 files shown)


src/index.ts
```diff
@@ -290,10 +290,9 @@ prog
     if (!opts.noClean) {
       await cleanDistFolder();
     }
-    opts.name = opts.name || appPackageJson.name;
```
agilgur5, line 293: ahhhh good catch, this line and the line below it were redundantly added in #130 , c.f. https://github.com/jaredpalmer/tsdx/pull/130#issuecomment-610716225 .

I'll remove both of these in a separate PR

agilgur5:
This is somewhere in between a bug and a feature, because the docs and the help dialog say it's only for UMD names, but apparently it's used for the output filename in the source code too.

I'll have to look at the history and possibly `microbundle` history to see if this is an evolution of the `--name` use-case or if it's a bug that it's used this way in the code.

In any case, this is a breaking change as there might be folks who use this undocumented behavior.
There's another related request back in https://github.com/jaredpalmer/tsdx/pull/208#issuecomment-536625558 / #219 around changing the output name to always just be `dist/index.js` instead of this `safePackageName` stuff which I agree with, for consistency's sake. Changing that would fix this too, but that change is _extremely_ breaking and this one not so much.

src/index.ts
kotarella1110: @agilgur5 Should I revert this fix?

agilgur5: It might merge conflict with that PR, probably better to revert yea. I'm just about falling asleep so I won't make that PR today. Hopefully will get that in tomorrow and I'll make sure to give you credit there too (as `Co-Authored-By`) for the find

kotarella1110: OK! I will!

kotarella1110:
@agilgur5 Do you have any update on this?

agilgur5:
So I did some investigation and this indeed some likes an unintentional bug that `--name` changes the filename on top of the UMD name. The docs both here and [in `microbundle`](https://github.com/developit/microbundle/#all-cli-options) say it's just for UMD.
`microbundle` has a [very similar bug](https://github.com/developit/microbundle/blob/2dd4afec7088300320e432a399a2f8a7a48ccbad/src/index.js#L317) that will cause, in its case, `options.name` to be set to the `--name` flag, but it doesn't affect `microbundle` nearly as much because its filenames are set to [whatever you've specified in your `package.json` fields](https://github.com/developit/microbundle/blob/2dd4afec7088300320e432a399a2f8a7a48ccbad/src/index.js#L439-L442). That might be a good feature for us to have, which would make any filename changes a bit less breaking (though itself would similarly be breaking if a user did any renames on their own).

That being said, because this is unfortunately a breaking change, this won't be able to go out soon.
[v0.14](https://github.com/jaredpalmer/tsdx/milestone/2), the next minor (minors are breaking in v0.x), is already going to be a pretty big release with a handful of dependency-related potentially breaking changes, so this won't make it in there either 😕
Probably won't be till at least v0.15, so I wouldn't hold out on this getting merged in soon unfortunately.

In the meantime, if you need this, I'd recommend using `patch-package` or configuring `tsdx.config.js` to get this behavior, sorry about the unfortunate timing / breaking-ness.

<hr>

Also in the meantime, I'll still make some review comments and I'll give credit to you with `allcontributors` for this bug report and code, esp since some code (the removals) was already used in #671 .

agilgur5:
@allcontributors please add @kotarella1110 for bug, code

PR #146, opened 2019-06-17
strip package scope in safePackageName function

hyanmandian:
fix #144

PR #672, opened 2020-04-11
(fix): @types/jest should be a dep, not a devDep

### LINKED ISSUE #633: Question: Why not include @types/jest as TSDX's dependencies?
### Current Behavior

`@types/jest` is in newly generated package's devDependencies.

### Desired Behavior

Why not include `@types/jest` as TSDX's dependencies?

This way, the newly generated boilerplate package's devDependencis will be cleaner.

### Suggested Solution

Actually, if the newly generated boilerplate package's devDependencis includes only TSDX, that will be awesome, and I think it's doable.

All users

None.

agilgur5:
- per bug report, this caused an issue where the @types/jest that was
  installed via `create` would have a version mismatch with the jest
  and jest-related deps used by TSDX
- and ts-jest's installation instructions also say to install
  @types/jest, so it's kind of a peerDep to it

- also move @types/shelljs to devDeps as it's not needed for usage of
  the CLI, it's only used internally

<hr>

Fixes #633 .
`@types/jest` was added to the templates in one of the first commits in 9010ecbda5121.
Per the issue, `@types/jest` was added as a devDep in #146 when the first TS test was added (makes sense) and `@types/shelljs` was added as a dep in #244


agilgur5:
`ts-jest` recently added `@types/jest` as a dep: https://github.com/kulshekhar/ts-jest/issues/1859

PR #678, opened 2020-04-12
(deps): upgrade several deps' majors; require Node 10+

agilgur5:
- officially drop support for Node <10 since many deps don't support
  it anymore
  - it has been EoL since Dec 31, 2019 and many deps are using newer,
    ES2017 language features now too
  - previously we deprecated it and semi-officially dropped, this time
    is fully official with engines requirement
    - and will be released as a breaking change

- camelcase, chalk, execa, and fs-extra all dropped Node <10
  - so did prettier v2 and rollup v2 which we will update to in the
    near future

- ora and rollup-plugin-sourcemaps dropped Node <8 in breaking bumps
  - so could've upgraded before, but greenkeeper didn't bring it up,
    probably because of previous downgrades or CI errors
  - Jest 25, which we will update to in the same breaking release, also
    dropped support for Node <8

- rollup-plugin-sourcemaps v0.5 is now written in TS and ships its own
  types, so no need for its @types/ counterpart anymore

<hr>

Officially drop Node 10 after deprecation in #426 and semi-official drop in #433 .

Merges together a handful of Greenkeeper PRs that required Node 10+:
- Closes #665 
- Closes #660 
- Closes #629 

And now Prettier v2 (#632) and Rollup v2 (#545) upgrades at least won't have to worry about this breakage (those do still have worry about other breakage still though 😅 )


agilgur5:
Checked what all the deprecation warnings on a fresh install of `v0.13.3` are per https://github.com/formium/tsdx/pull/789#issuecomment-695854530, and saw that this should fix two of them since they're transitive dependencies of `rollup-plugin-sourcemaps` v0.4.2, which was upgraded here to v0.5:

```bash
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated // rollup-plugin-sourcemaps@0.4.2
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated // rollup-plugin-sourcemaps@0.4.2
```

PR #679, opened 2020-04-13
(deps): upgrade to Jest 25

### LINKED ISSUE #664: Use Jest 25
### Current Behavior

Jest 24.9

### Desired Behavior

Jest 25

### Suggested Solution

Upgrade to Jest 25

Everyone that uses Jest tests

Chai

### Additional context

Maybe there will be breaking changes, I'm not sure so will have to check the source code

agilgur5:
- upgrade jest, ts-jest, @types/jest, and jest-watch-typeahead
- Jest 25 drops support for Node <8 that doesn't impact us
  - we only support Node 10+ anyway
- Jest 25 upgrades JSDOM from v11 to v15, which is potentially breaking
  for any users that relied on some older behavior

- yarn.lock changed heavily
  - from a look-over, many are from babel-jest (which is included with
    Jest) upgrading its Babel deps, some from istanbul for coverage,
    and lots of other transitive deps I can't really tell from a glance

<hr>

Fixes #664 

Merges a few Greenkeeper PRs together so all of Jest 25 related deps are done in one step:
- Closes #471 
- Closes #497 
- Closes #658 

Not sure why Greenkeeper didn't make a PR to update Jest itself, it only did the related deps 🤔 

<hr>

**This is built on top of #678 , merge that first**


Bnaya:
Make it 26 now? 😅

agilgur5:
@Bnaya other than Jest 26 having more breaking changes, it also [requires TS 3.8](https://jestjs.io/blog/2020/05/05/jest-26). TS 3.8 private fields [break `rollup-plugin-typescript2`](https://github.com/ezolenko/rollup-plugin-typescript2/issues/217), which is more an upstream [Rollup issue](https://github.com/rollup/rollup/issues/3128), which is more of an upstream Acorn issue, which hasn't been maintained too well [so Rollup contributors forked it for class fields support](https://github.com/rollup/rollup/pull/3488) (but not private fields), and those would only really be fixed in Rollup v2, which is also breaking, and [breaks various plugins](https://github.com/jaredpalmer/tsdx/pull/693#issuecomment-619494004) TSDX users use.

Rest assured, I have my eyes on the deps and how they break TSDX's users' code.

nareshbhatia:
@agilgur5, I am porting over a significant monorepo to TSDX. Getting the packages to compile was fairly straight forward, however many of the tests are failing because TSDX is using Jest v24.9.0. My original monorepo is using Jest v26.1.0 and runs successfully, thanks to various bug fixes in Jest.

I read your comment about TSDX's dependency on rollup-plugin-typescript2 which is holding up the upgrade to Jest. What workaround would you recommend in my case?

TIA.

agilgur5:
@nareshbhatia if you need Jest 26, you don't need to wait for TSDX to upgrade to it. You don't need to use `tsdx test` and can swap it for `jest`.

nareshbhatia:
Turns out that even after upgrading to Jest 26 and running jest directly, those errors would not go away. Here are the specific errors:

- MutationObserver is not a constructor
- TypeError document.createRange is not a function

From everything I read, Jest 26 now includes jsdom 16 which fixes these errors. All I can think of is somehow the correct version of jsdom is not being picked up.

Finally I installed [jest-environment-jsdom-sixteen](https://github.com/SimenB/jest-environment-jsdom-sixteen) and forced it via `jest.config.js`:

```
testEnvironment: 'jest-environment-jsdom-sixteen'
```

This made all the tests pass. This works even with `tsdx test`. So I am switching back to it (removed jest 26 completely). Can't explain exactly why it didn't work with Jest 26, but hope this helps anyone else running into this issue.

agilgur5:
> Can't explain exactly why it didn't work with Jest 26

Yea that's weird. Maybe Jest's auto-import capabilities glitched out with two versions of Jest/JSDOM installed. I know there have been issues with it before but I believe Node's algorithm should choose the closest one by default (which should be the one the Jest version you use specifies).

> MutationObserver is not a constructor

Yea this one specifically was addressed in #724 as well.

Good to hear it got resolved one way or another though and that you could still stick to `tsdx test`.

agilgur5:
Checked what all the deprecation warnings on a fresh install of `v0.13.3` are per https://github.com/formium/tsdx/pull/789#issuecomment-695854530, and saw that this should fix several of them since they're all from transitive dependencies of JSDOM 11, which was upgraded to JSDOM 15 with Jest 25:

```bash
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2. // jest@24.9.0 ... jsdom@11.12.0
npm WARN deprecated left-pad@1.3.0: use String.prototype.padStart() // jest@24.9.0 ... jsdom@11.12.0
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142 // // jest@24.9.0 ... jsdom@11.12.0
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142 // jest@24.9.0 ... jsdom@11.12.0
npm WARN deprecated har-validator@5.1.5: this library is no longer supported // jest@24.9.0 ... jsdom@11.12.0 ... request@2.88.0
```

JBaczuk:
> Finally I installed jest-environment-jsdom-sixteen and forced it via jest.config.js:
>
> testEnvironment: 'jest-environment-jsdom-sixteen'

I can confirm this workaround from @nareshbhatia  fixes the issue for me as well. I was having the same problem with `TypeError document.createRange is not a function` preventing me from using `userEvent.type` from `testing-library/react` and the issue is the version of jest. I'm not sure why the test environment needs to be overridden, even without using tsdx and using `jest` directly, instead.

PR #683, opened 2020-04-14
deps: add deduplicate script and apply yarn-deduplicate

Bnaya:
Motivation:
https://github.com/atlassian/yarn-deduplicate#why-is-this-necessary

Benefits:
Smaller & less nested node_modules
less deps are loaded into node when running dev stuff

It's good to run this command after every deps changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "test:post-build": "node dist/index.js test",
     "start": "tsc -p tsconfig.json --watch",
     "release": "np",
-    "deduplicate": "npx yarn-deduplicate -s fewer yarn.lock"
+    "deduplicate": "yarn-deduplicate -s fewer yarn.lock",
+    "deduplicate:check": "yarn-deduplicate -s fewer yarn.lock --list --fail"
   },
   "files": [
     "dist",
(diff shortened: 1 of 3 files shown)


agilgur5:
Huh, was looking at some of my recent PRs to change deps and did see duplicates -- but Yarn suggests that can't happen -- so I was confused. Guess Yarn has some bugs around this. Thanks for the add!

package.json
```diff
@@ -28,7 +28,8 @@
     "test": "yarn build && yarn test:post-build",
     "test:post-build": "node dist/index.js test",
     "start": "tsc -p tsconfig.json --watch",
-    "release": "np"
+    "release": "np",
+    "deduplicate": "npx yarn-deduplicate -s fewer yarn.lock"
```
agilgur5, line 32: This should probably be a `postinstall` hook (and at the top of scripts)

Bnaya: After running that dedup, you need to run yarn install again to make the yarn.lock changes to actually effect the node_modules
And adding another yarn install as post install hook can make a loop and mess

My personal way of using this tool is mostly manually

Bnaya:
> Huh, was looking at some of my recent PRs to change deps and did see duplicates -- but Yarn suggests that can't happen -- so I was confused. Guess Yarn has some bugs around this. Thanks for the add!

That's behaviour of yarn is indeed cryptic, but not completely senseles.
There's more info and discussions around it on yarn issues and yarn-deduplicate docs

Bnaya:
Worth noting:
I choose to use the "fewer" strategy and not "newer"

package.json
agilgur5: > The `--list --fail` exits, but it wasn't documented

> I've made similar PR to webpack, this is how they eventually choose to impl it:
> https://github.com/webpack/webpack/blob/master/package.json#L157-L158
> `yarn-lint` runs as part of general lint
> `yarn-lint-fix` is manual

Ok will try this out, thanks for the notes. I was asking more of if that might cause a loop but guess not

> I personally don't really like pre commit hooks, (Not for failing or for mutating source code like prettier fix)
> I prefer CI checks and pre push

Yea I get that it can be a bit of a hassle when you've got a commit in your head or no relevant files have changed and then you have to wait for pre-commit to run. But with pre-push only and CI, you miss potential errors in between commits that are fixed before the final push/PR, meaning it allows for individual commits to be invalid so long as the whole is valid.

For the most part, PRs get squash-merged, but sometimes I rebase things that are meant to be distinct commits

agilgur5:
@all-contributors please add @Bnaya for bug, code

PR #691, opened 2020-04-20
(optim/fix): only emit type declarations once

agilgur5:
- every other emission is just a duplicate -- no need to spend compute
  time to make duplicates
  - even for the upcoming multi-entry, the same types are emitted for
    each entry as the compiler just emits types for everything in the
    `tsconfig` `include`

- fixes a long-standing bug with the deprecated moveTypes() function
  that would occasionally cause types to not be properly output
  - this was actually due to moveTypes() being run multiple times in
    parallel (as well as the types themselves being emitted multiple
    times in parallel)
    - hence the EEXIST and ENOENT filesystem errors being thrown, as
      that directory was being changed multiple times in parallel
      - race conditions, fun!
  - now they're only emitted once and only moved once, so no problems!

- also fixes a bug with an initial version of multi-entry where if an
  entry in a subdir of src/ were added, e.g. src/foo/bar, the entire
  tree of type declarations would get output into dist/foo/src/*.d.ts
  - alternatively, could call `moveTypes()` with an arg for each entry,
    but there's no need since all declarations get produced the first
    time around anyway
  - similar bug occurred with an initial version of `--preserveModules`
    that used separate format directories

Also threw in another TS optimization here with the `cacheRoot` change that's along the same lines but slightly different of a change.

<hr>

This is extracted out of #367 's 14b539763ef21, but is a better version in a few ways:
1. Doesn't have issues with `declarationMap`s as they've also been set to `false`. Otherwise would get errors / test failures that you can't have `declarationMap`s without `declaration`s
    - Better testing around this nowadays revealed this bug
1. `: undefined` still overrode some `declaration`/`declarationMap` settings, also causing errors / test failures, so changed this to not set either on the first emission
1. `moveTypes()` was moved into a `then` of the `asyncro()` so that it's still part of the progress estimation.

The bugs this fixed in #367 and by extension #535 should no longer be present anyway since #367 switched/improved to use Rollup's code-splitting instead of a separate build & bundle per entry and #535 will need to be updated to use the same `output.entryFileNames` as a later version of #367 used instead of using separate format dirs.

<hr>

I also realized that this might fix the bugs with `moveTypes()` that gave us so many CI issues prior to #504 very recently in https://github.com/jaredpalmer/tsdx/pull/500#issuecomment-605530750.
And indeed this does fix those bugs, hence the removals of the error throwaways and a change in the deprecation message by a bit (it's still a problematic option, with another caveat other than those listed that `moveTypes()` doesn't apply to custom `declarationDir`s, though that could be fixed). I tested that this was the case by changing `rootDir` to `./` in every test fixture and running tests a couple of times and no errors were thrown anymore!

<hr>

The `cacheRoot` optimization is also related to #328 / #329. And might make #618 obsolete.

### DIFF
--- a/src/createBuildConfigs.ts
+++ b/src/createBuildConfigs.ts
@@ -35,9 +35,9 @@ export async function createBuildConfigs(
   );
 
   return await Promise.all(
-    allInputs.map(async (options: TsdxOptions) => {
+    allInputs.map(async (options: TsdxOptions, index: number) => {
       // pass the full rollup config to tsdx.config.js override
-      const config = await createRollupConfig(options);
+      const config = await createRollupConfig(options, index);
       return tsdxConfig.rollup(config, options);
     })
   );
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -146,7 +146,6 @@ export async function createRollupConfig(
       },
       typescript({
         typescript: ts,
-        cacheRoot: `./node_modules/.cache/tsdx/${opts.format}/`,
         tsconfig: opts.tsconfig,
         tsconfigDefaults: {
           exclude: [
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -24,7 +24,8 @@ const errorCodeOpts = {
 let shebang: any = {};
 
 export async function createRollupConfig(
-  opts: TsdxOptions
+  opts: TsdxOptions,
+  outputNum: number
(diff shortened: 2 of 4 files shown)


agilgur5:
Ah, also should note that from my local test runs, the time it took to run all tests decreased by 10-15%, which is a substantial perf improvement!
If this tiny change was so substantial, switching to per-ouput plugins is _really_ going to be _big_ when building multiple output formats 😮

Issue #634, opened 2020-03-23
[RFC] Presets


agilgur5:
### Current Behavior

Right now, TSDX merges in any external configs with its own, i.e. users with `babelrc`, `eslintrc`, `jest.config.js`, etc don't have full agency over their configs if they want to opt-out of something.

The most confusion this has caused is with `tsdx lint`, e.g. #514, #498 . Linting tends to very project-specific, and when the rules don't line up 1-to-1 with the config (because it's merged), this is very noticeable and the solution is not intuitive.

The merge behavior also causes problems with Editor / IDE integrations, which has particularly impacted `tsdx test` as with #270 and its many duplicates and related issues.

This hasn't seemed to have affected `tsdx build` / `.babelrc` as much, but can be noticeable in issues like #383, where there have been inconsistencies between `babelrc` use in commands like `build` vs. `test`, as well as #543 where users are unable to change plugin order to that which they need and are stuck with whatever we choose because they can't opt-out easily either (without overriding `rollup-plugin-babel` in `tsdx.config.js`).

This merge behavior is not necessarily that uncommon, but some zero-config libraries do use presets instead as well.

### Desired Behavior

Allow full overridability of `.babelrc`, `.eslintrc`, and `jest.config.js` when a user wants to do. 
Make integrations with existing editor tools more straightforward.

### Suggested Solution

Add `babel-preset-tsdx`, `eslint-config-tsdx` (or `eslint-plugin-tsdx` that also has a config for easier installation), and `jest-preset-tsdx`. Or their equivalent `@tsdx/` versions.

Also add a default `.babelrc.js`, `.eslintrc.js`, and `jest.config.js` to all the templates that just use these presets and nothing else to make for instant editor integration.

The ONE issue with this is that the internal Babel config merge is a bit more complicated than a plain merge; it does some custom configuration of `preset-env` and a few plugins are dynamic based on the options passed to `tsdx build`. So we'll have to leave some of that custom merging behavior in and may need to account for the case of e.g. `babel-preset-tsdx` instead of `babel-preset-env`. Some things can be configured by just allowing users to pass options into the preset, but those won't be dynamic or reflect `tsdx build`.
Still need to think more on that specific case, will probably be more clear once the implementation is underway.

Could also add a `tsconfig.json` preset as well, which would make things like #504 not quite as breaking, but might not be so wanted as these get configured more frequently.

### Who does this impact? Who is this for?

Any users with custom Babel, ESLint, or Jest configs.
[...]

PR #705, opened 2020-05-01
Add size-limit bundle size analysis tool

andresz1:
## Performance budget tool
This PR is an **initial idea** (something that I thought that maybe could be useful after @jaredpalmer tweet) of how a **performance budget tool** could be added automatically in every template. Having this kind of tool helps you to keep your library **bundle size** as **small** as possible. I like [size-limit](https://github.com/ai/size-limit) because besides the size it can calculate the time it would take a browser to download and execute your JS.

I added `@size-limit/preset-small-lib` by default but in case that library size increases, it could be easily replaced by `@size-limit/preset-big-lib` (adding time).

### Bundle size check
```bash
yarn size
```
<div align="center">
Screenshot 2020-05-01 at 13 47 24
</div>

### Bundle analyze
```bash
yarn analyze
```
<div align="center">
Screenshot 2020-05-01 at 13 47 57
</div>

### GitHub action
The action added ([size-limit-action](https://github.com/andresz1/size-limit-action)) will execute `size-limit` a comment the PRs made with the comparison of it.

<div align="center">
pr
</div>

Any feedback would be appreciated and I'd be more than happy to iterate over a solution if you like the idea. Thanks in advance and for the great tool!


andresz1:
Hi @agilgur5! please let me know if you have any thoughts about this! 🙏

agilgur5:
Hey @andresz1 I definitely like the changes! I took a look when you first made the PR, but I just haven't had the time to dive deeper and review all the pieces of this.
There are a handful of related issues and some code that was commented out from removed features related to this that I need to sort through, understand how `size-limit` may differ from those, think through template vs. dependency route, and then actually review the code

andresz1:
Hi @agilgur5! I'm glad that you like the idea 🎉 and thank you for the reply. Let me know if you need help or any further explanation. Is there any `tsdx` discord or spectrum channel where we could discuss ?

agilgur5:
Sorry for the delay on reviewing this, I've been incredibly stressed and busy, so larger PRs to add features got pretty backlogged as they take a good bit of time to really dive into.

> There are a handful of related issues and some code that was commented out from removed features related to this that I need to sort through

Here's a few past references:
- #81
- https://github.com/formium/tsdx/blob/79dddc4ed6aeee89472999cede9e44ac951c548c/src/createRollupConfig.ts#L197-L199
- https://github.com/formium/tsdx/blob/5e5c3f804e68609083be978a080ab905dc7ec19d/src/env.d.ts#L15
- #496
- #741

Summarizing, `rollup-plugin-size-snapshot` was previously used and still has some unused / commented out code references to it, `rollup-plugin-visualizer` was previously considered, and `rollup-plugin-analyzer` was previously recommended

> understand how `size-limit` may differ from those

Did a quick glance through [on NPM trends](https://www.npmtrends.com/size-limit-vs-bundlesize-vs-cost-of-modules-vs-bundlewatch-vs-rollup-plugin-size-snapshot-vs-rollup-plugin-sizes-vs-rollup-plugin-visualizer-vs-rollup-plugin-analyzer) and `size-limit` seems to have a lot of nicer features:
- currently maintained (some alternatives are not)
- also does timing (none of the others do)
- also does visualization (most of the others don't)
- doesn't need to be added to Rollup build (some of the others do)

That being said, it does have more dependencies like `webpack` which I'm not sure are necessary for our use-case considering TSDX bundles everything with Rollup. It looks like `webpack` is a plugin though and can be disabled, but is included in `@size-limit/preset-small-lib`. The visualizations aren't as fancy as `rollup-plugin-visualizer` but probably nbd.

> think through template vs. dependency route

Can probably stick to template initially as a testing ground and then move to a dependency later if there is enough usage. Though with #634 we're moving a bit away from dependencies.

> investigate any potential impact on multi-entry and/or `preserveModules`

I think this will require something like webpack to bundle pieces. Multi-entry it would need to show each entry and `preserveModules` should theoretically all be bundled up from one entry.

> and actually review the code

In progress.

> Is there any `tsdx` discord or spectrum channel where we could discuss ?

No, and I don't think that would be helpful either. Keeping discussions async and transparent on a PR is ideal. Watching those channels would take more time as well.

templates/basic/.github/workflows/size.yml
```diff
@@ -0,0 +1,15 @@
+name: "size"
```
agilgur5, line 1:
```suggestion
name: size
```

YAML doesn't require quoting for the most part and the rest here and in the other actions in existing templates isn't quoted, so let's stay consistent

templates/basic/.github/workflows/size.yml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  pull_request:
+    branches:
+      - master
```
agilgur5, line 5:
```suggestion
on: [pull_request]
```

I don't think it normally matters what branch it is to, even to a `test` branch this would still be relevant/useful. Also main branches are not always named that and increasingly less nowadays.

templates/basic/.github/workflows/size.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  size:
+    runs-on: ubuntu-latest
+    env:
+      CI_JOB_NUMBER: 1
+    steps:
+      - uses: actions/checkout@v1
+      - uses: andresz1/size-limit-action@v1.3.0
```
agilgur5, line 13:
```suggestion
      - uses: andresz1/size-limit-action@v1
```

Let's keep consistent with all the other actions and general NPM Semver practice and not pin to a specific patch version.

templates/basic/.github/workflows/size.yml
```diff
@@ -0,0 +1,15 @@
+name: "size"
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  size:
+    runs-on: ubuntu-latest
+    env:
+      CI_JOB_NUMBER: 1
```
agilgur5, line 10: what's the purpose of this? I found https://github.com/ai/ci-job-number, but this doesn't run a matrix in any case; there is only one job here.

src/templates/basic.ts
```diff
@@ -2,7 +2,14 @@ import { Template } from './template';
...
 const basicTemplate: Template = {
   name: 'basic',
-  dependencies: ['husky', 'tsdx', 'tslib', 'typescript'],
+  dependencies: [
+    'husky',
+    'tsdx',
+    'tslib',
+    'typescript',
+    'size-limit',
+    '@size-limit/preset-small-lib',
```
agilgur5, line 11: Does this not need any configuration to use? The docs didn't specify how to add presets to the config

agilgur5:
Ah, also

Two references from above comment should be removed here too:
 - https://github.com/formium/tsdx/blob/79dddc4ed6aeee89472999cede9e44ac951c548c/src/createRollupConfig.ts#L197-L199
- https://github.com/formium/tsdx/blob/5e5c3f804e68609083be978a080ab905dc7ec19d/src/env.d.ts#L15

Should really add some docs somewhere about this. The template READMEs may be appropriate places for that since they mention CI

templates/basic/.github/workflows/size.yml
andresz1: Agree 👍

templates/basic/.github/workflows/size.yml
andresz1: Agree 👍

src/templates/basic.ts
andresz1: A default configuration is added in `package.json` file and IMO there's no need to add more stuff in this case. This could be edited by users if they want to

agilgur5: The `package.json` config doesn't reference `@size-limit/preset-small-lib` though -- does it automatically detect it or something?

templates/basic/README.md
```diff
@@ -30,6 +30,10 @@ Code quality is set up for you with `prettier`, `husky`, and `lint-staged`. Adju
 
 Jest tests are set up to run with `npm test` or `yarn test`.
 
+### Bundle analysis
+
+Calculates the real cost of your library using [size-limit](https://github.com/ai/size-limit) with `npm run size` and visulize it with `npm run analyze`.
```
agilgur5, line 35:
```suggestion
### Bundle Analysis

[`size-limit`](https://github.com/ai/size-limit) is set up to calculate the real cost of your library with `npm run size` and visualize the bundle with `npm run analyze`.
```

Consistency with capitalization and sentence structure with the rest of the sentences in this section. Also fixes a spelling mistake (visulize -> visualize)

templates/basic/README.md
```diff
@@ -57,10 +61,10 @@ TSDX uses [Rollup](https://rollupjs.org) as a bundler and generates multiple rol
 
 ### GitHub Actions
 
-A simple action is included that runs these steps on all pushes:
+Two actions are added by default:
 
-- Installs deps w/ cache
-- Lints, tests, and builds
+- `main` which runs installs deps w/ cache, lints, tests, and builds on all pushes
```
agilgur5, line 66:
```suggestion
- `main` which installs deps w/ cache, lints, tests, and builds on all pushes
```

Fixes grammatical issue

templates/basic/README.md
```diff
@@ -57,10 +61,10 @@ TSDX uses [Rollup](https://rollupjs.org) as a bundler and generates multiple rol
...
+- `size` which comments cost comparison of your library on every pull request using [size-limit](https://github.com/ai/size-limit)
```
agilgur5, line 67:
```suggestion
- `size` which comments cost comparison of your library on every pull request using [`size-limit`](https://github.com/ai/size-limit)
```

agilgur5:
@andresz1 I'd make the changes myself and merge but there were some unresolved questions remaining from previous review that I was hoping you could answer

templates/basic/.github/workflows/size.yml
andresz1: I wasn't able to remember why this was needed 😱  I tried without it and the problem is that size-limit doesn't work properly in the action (it throws an error), so for now is needed. I'll try to investigate more about this

src/templates/basic.ts
andresz1: Yep! you don't need to specify that you are using `@size-limit/preset-small-lib`, you just need to install it (check [docs](https://github.com/ai/size-limit#small-libraries))

andresz1:
@agilgur5 sorry I was on vacations and thank you so much for the great feedback! I just answered your questions and made so changes!

src/templates/basic.ts
agilgur5: Thanks for the confirmation. Seems like some new docs were added that specify the step-by-step and don't mention configuring which confirms, but would be great if the docs explicitly stated that 😕 . Might file a PR for that

templates/basic/.github/workflows/size.yml
agilgur5: Do you have a log of the error? Since this is the only job in this new workflow, making it go first shouldn't be necessary, but maybe the action relies on this variable being specified in order to work

agilgur5:
@andresz1 no worries on the delay or if you're on vacation, everyone does that. Was just following up, especially since I couldn't find the answers to some of these myself.

agilgur5:
@all-contributors please add @andresz1 for code, doc, example

PR #723, opened 2020-05-17
docs: add links to top with jump2header

strdr4605:
Added link to for so it will be easier to navigate in README.

[Demo](https://github.com/strdr4605/tsdx/blob/jump2header/README.md).

Generated with:

```
jump2header -slug toc -position end -emoji 2 --start Features -max-level 2
```

with small manual adjustments.

If you don't like the PR but interested in links to top see [jump2header](https://github.com/strdr4605/jump2header#readme) for more options.


README.md
```diff
@@ -58,6 +60,8 @@ TSDX comes with the "battery-pack included" and is part of a complete TypeScript
 - Jest test runner setup with sensible defaults via `tsdx test`
 - Zero-config, single dependency
 
+[🔝](#toc)<!-- Link generated with jump2header -->
```
agilgur5, line 63:
```suggestion
[🔝 Back to Top 🔝](#toc)<!-- Link generated with jump2header -->
```

Symbols are not the most accessible and this could be missed easily, so should add some text to each of these.

agilgur5 (changes requested):
Thanks for the PR and apologies for the delay.

I think this is useful, however it will be superseded by the fix to #848 and so is not necessary.

agilgur5:
Superseded by #848

agilgur5:
@allcontributors please add @strdr4605 for idea

Issue #438, opened 2020-01-22
[proposal] add github actions to templates


sw-yx:
### Current Behavior

gh actions are awesome. lets recommend and auto scaffold some for people to test their tsdx libraries? i know not everybody uses github but they could just delete the .github folder if so.

pull some from https://github.com/sdras/awesome-actions

what actions would people like?

e.g. i'd like matrix testing by default, as well as issue/pr templates

ambroseus:
action for caching  .npm  between builds:
```yaml
- uses: actions/cache@v1
   with:
      path: ~/.npm
      key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
      restore-keys: |
      ${{ runner.os }}-node-
```

ambroseus:
collect, summarize and upload coverage reports from packages in monorepo (for codeclimate):
```yaml
     - name: upload coverage
        run: |
          # collect coverage
          export GIT_BRANCH="${GITHUB_REF/refs\/heads\//}"
          curl -s -S -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./reporter
          chmod +x ./reporter
          for f in `find packages -maxdepth 2 -mindepth 1`
          do
            if [ -e $f/coverage/lcov.info ]; then
              echo $f;
              ./reporter format-coverage -t lcov -o cov/${f//\//-}.json $f/coverage/lcov.info;
            fi
          done;
          ./reporter sum-coverage -o cov/total.json cov/*.json
          ./reporter upload-coverage -i cov/total.json -r ${{ secrets.test_reporter_id }}
```

agilgur5:
I think basic issue/PR templates that can then be customized per library would be great (and are often under-prioritized) and are fairly universal too.

I'm ambivalent about Actions. I use Travis for my libraries and it's a lot more reliable, readable, and functional than GH (e.g. simple things like restarting a test) and has one-liner code coverage integration. On the one hand, it's much easier to delete stuff than to add stuff, but that can quickly spiral into having too many things to delete.

As more stuff gets added to the templates (e.g. actions, storybook, etc), it might make sense to have TSDX wrap more of that functionality internally instead of exposing all the config to users (that also means that instead of needing to delete something, you just don't use that functionality, which is even easier).

kylemh:
Is there a way to do this with some system better than templates? What if I want Storybook TSDX + GitHub Actions?

[...]

PR #733, opened 2020-05-23
Upgrade engines.node to >=12 in package.json for basic template

zenflow:
The `engines.node` field should match the version of node we are testing against in CI: https://github.com/jaredpalmer/tsdx/blob/21893b8e498985879bb6665dba279cf4853aabec/templates/basic/.github/workflows/main.yml

This change doesn't apply to the other templates

Thanks for the awesome package! You saved me a ton of time for sure! <3

### DIFF
--- a/src/templates/basic.ts
+++ b/src/templates/basic.ts
@@ -13,7 +13,7 @@ const basicTemplate: Template = {
     typings: `dist/index.d.ts`,
     files: ['dist', 'src'],
     engines: {
-      node: '>=10',
+      node: '>=12',
     },
     scripts: {
       start: 'tsdx watch',


agilgur5:
I appreciate that you were looking for some consistency here, but I'm sorry to say that I'll have to decline the PR in its current state.

The current GitHub Action was meant to just be a [starting](https://github.com/jaredpalmer/tsdx/issues/438#issuecomment-578283526) [point](https://github.com/jaredpalmer/tsdx/pull/457#issuecomment-578442500) and not the end state. The action in each template is also [meant to be the same](https://github.com/jaredpalmer/tsdx/pull/457#issue-367094544); it's a limitation of our current templating set-up that there are multiple files.

Node 10 is still LTS for another year, [until May 2021](https://nodejs.org/en/about/releases/), so I would certainly not recommend library authors only support Node 12+. It would be better to add [matrix testing](https://github.com/jaredpalmer/tsdx/issues/438#issue-553603931) for different Node versions and OSes like TSDX itself does, which is likely to be added in the near future. Per the proposal issue #438, the basic variant was meant not to significantly bloat the templates as an initial feature especially as we'd prefer partial scaffold commands that could set-up things like Actions _or_ TravisCI for instance.

PR #736, opened 2020-05-31
fix: preserve package name (fix #714)

### LINKED ISSUE #714: tsdx create support scope name
### Current Behavior
after running `tsdx create @byted/test`, tsdx creates a package with the following pakcage.json
```
// package.json
{
  name:"test" // I actually want this to be @byted/test
}
```

### Expected behavior
generate the following package.json
```
// package.json
{
  name:"@byed/test" // I actually want this to be @byted/test
}
```

### Suggested solution(s)
it's confusing scope means directory or scope, maybe a prompt for package name could help,which use @byted/test as default
image

### Your environment

| TSDX           |
| TypeScript       |
| Browser          |
| npm/Yarn         |
| Node             | 
| Operating System |

fi3ework:
Preserve inputted package name of `name` field in package.json.

## Before

input @byted/test,
```json
name: test
```

## After

input @byted/test,
```json
name: @byted/test
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -218,7 +218,11 @@ prog
       // Install deps
       process.chdir(projectPath);
       const safeName = safePackageName(pkg);
-      const pkgJson = generatePackageJson({ name: safeName, author });
+      const pkgJson = generatePackageJson({
+        safeName,
+        author,
+        pkg,
+      });
 
       const nodeVersionReq = getNodeEngineRequirement(pkgJson);
       if (
--- a/src/templates/utils/index.ts
+++ b/src/templates/utils/index.ts
@@ -1,17 +1,20 @@
 import { Template } from '../template';
 
 interface ProjectArgs {
-  name: string;
+  safeName: string;
   author: string;
+  pkg: string;
 }
+
 export const composePackageJson = (template: Template) => ({
-  name,
+  safeName,
   author,
+  pkg,
 }: ProjectArgs) => {
   return {
     ...template.packageJson,
-    name,
+    name: pkg,
     author,
-    module: `dist/${name}.esm.js`,
+    module: `dist/${safeName}.esm.js`,
   };
 };


src/index.ts
```diff
@@ -218,7 +218,11 @@ prog
       // Install deps
       process.chdir(projectPath);
       const safeName = safePackageName(pkg);
-      const pkgJson = generatePackageJson({ name: safeName, author });
+      const pkgJson = generatePackageJson({
+        safeName,
+        author,
+        pkg,
```
agilgur5, line 224:
```suggestion
        name: pkg,
        safeName,
        author,
```

It really should be called `pkgName` or something, but that's a refactor, so we're stuck with `pkg` in `create` for now, but should not use that poor naming elsewhere if avoidable. `pkg` itself is overloaded a few times within this file too.

src/templates/utils/index.ts
```diff
@@ -1,17 +1,20 @@
 import { Template } from '../template';
 
 interface ProjectArgs {
-  name: string;
+  safeName: string;
   author: string;
+  pkg: string;
```
agilgur5, line 6:
```suggestion
  name: string;
  safeName: string;
  author: string;
```

src/templates/utils/index.ts
```diff
@@ -1,17 +1,20 @@
...
 }
+
 export const composePackageJson = (template: Template) => ({
-  name,
+  safeName,
   author,
+  pkg,
```
agilgur5, line 12:
```suggestion
  name,
  safeName,
  author,
```

src/templates/utils/index.ts
```diff
@@ -1,17 +1,20 @@
...
 }: ProjectArgs) => {
   return {
     ...template.packageJson,
-    name,
+    name: pkg,
```
agilgur5, line 16:
```suggestion
    name,
```

agilgur5: So removing `safeName` here will actually create potential for an error -- there's no check that what you typed in is a valid `package.json.name`. What we can do to workaround this is add a check in `create` and ask for a name again if the name is invalid, so only valid names make it here.

agilgur5 (changes requested):
## Thanks, sorry

Thanks for the PR @fi3ework. I took a look at this as I was going through existing PRs and seeing what else can make it in with all the breaking changes ongoing.

Apologies for the delay as I've been extremely busy, now recently laid off, this required a good bit of digging to check the edge cases, and there are higher priority tasks to address.

## Required Changes and potential breakage

So this does not seem to impact `build` as it also uses `safePackageName(pkgJson.name)` basically, although this will likely break UMD for scoped names as there's a bug there that I discovered during #367: https://github.com/formium/tsdx/blob/2938ed91e71d4cbcde6b5d692254b920b28a0003/src/createRollupConfig.ts#L100

Fixing that bug is potentially a breaking change, but might be able to get the scope to a bugfix. That can be a totally separate PR as well, which might be a better option in fact.

There's also another major issue that I mentioned in my last comment below that needs to be resolved before this is ready to go.

### Potential to release as patch

I think this could be released as patch if everything is resolved within scope properly, as that way it would only affect `create` and only fix issues in `build`. I.e. only improvements/changes for new packages, doesn't affect existing ones and therefore backward-compatible (this could arguably be considered breaking as well, but then every template update would be, that would really require splitting out `create` into its own package, i.e. #31)

PR #745, opened 2020-06-12
Simplify getInputs function

xiaoxiaojx:


jaredpalmer:
LGTM

agilgur5:
@jaredpalmer would appreciate a heads-up before you come back and merge code very quickly. I have been handling this myself for some time and have been following a process...

I would also not have merged this because this actually further complicates it into a single return statement and further wraps the code, along with adding a cast in there.

I have had to _unwrap_ similar code in the codebase before, such as in `createBuildConfigs`. The more you wrap something the more brittle and harder to refactor it becomes, a simplification would have been to unwrap it further, with the ternaries being a particularly good place to start

PR #750, opened 2020-06-24
deps: upgrade actions/cache to v2 in all templates

devrelm:
[actions/cache@2](https://github.com/actions/cache/releases/tag/v2.0.0) has been released and comes with some useful updates.

### DIFF
--- a/templates/basic/.github/workflows/main.yml
+++ b/templates/basic/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
           node-version: 12.x
 
       - name: Use cached node_modules
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: node_modules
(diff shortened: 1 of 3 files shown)


agilgur5:
Thanks, will merge this in this weekend or next week when I'm back from vacation.

The current implementation is flawed btw, as it caches `node_modules` instead of `yarn-cache` or `~/.npm`.
I've been planning on changing the templates to use the `npm_install` action as TSDX uses as of #625, but that one is buggy on Mac and Windows and my issue report hasn't been responded to https://github.com/bahmutov/npm-install/issues/24. Planning to fork and fix that, but it relies on a fork of some of the GH actions since there weren't NPM packages of them -- now there are per the changelog so that should make that easier.

This is still an improvement over the current version so still makes sense to merge until that is ready.

agilgur5:
@allcontributors please add @devrelm for code

PR #758, opened 2020-07-12
optim: only check types once

Semperia:
Fixes https://github.com/formik/tsdx/issues/757

Change-Id: I998648be8181371ec14e23a3096dc3b4ac731957

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -176,7 +176,7 @@ export async function createRollupConfig(
               : {}),
           },
         },
-        check: !opts.transpileOnly,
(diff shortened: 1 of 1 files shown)


agilgur5:
The CI runs can vary highly and some came out slower than previous CI, but in other places there was a 10-20s improvement, which is big. I'll have to run it over a few times locally to get a more accurate estimate since CI runners aren't quite as stable.

agilgur5:
It's roughly another 10-15% perf boost, so together with #691 it's a 20-30% perf boost! YMMV of course though

agilgur5:
@allcontributors please add @Semperia for code

PR #762, opened 2020-07-16
Update help channels to point to formium org

jssee:
Currently, the URLs pointing to the repo/issues are not uniform or correct.
It's not a huge deal since it seems all redirects are current working correctly,
but consistency is nice

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -11,7 +11,7 @@ export const help = function() {
   return `
     Only ${chalk.green('<project-directory>')} is required.
     If you have any problems, do not hesitate to file an issue:
-      ${chalk.cyan('https://github.com/formik/tsdx/issues/new')}
+    ${chalk.cyan('https://github.com/formium/tsdx/issues/new')}
   `;
 };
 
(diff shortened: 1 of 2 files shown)


jssee:
@agilgur5 good catch, updated!

src/messages.ts
```diff
@@ -11,7 +11,6 @@ export const help = function() {
   return `
     Only ${chalk.green('<project-directory>')} is required.
     If you have any problems, do not hesitate to file an issue:
-      ${chalk.cyan('https://github.com/formik/tsdx/issues/new')}
```
agilgur5, line 14: think you might have accidentally made a mistake here as it got deleted instead of updated

jssee: sorry, this was an oversight. fixed now.

jssee:
@agilgur5 i cant rebase and fix merge conflicts, i  think it might have to do with permissions?

agilgur5:
@jssee ah so #765 introduced a conflict here.. Not sure what kind of error you're getting as command line `git` shouldn't normally give permissions errors. I also have no control over permissions in this repo in any case, so if that were actually the case, I wouldn't be able to resolve 😕

jssee:
@agilgur5 no problem, i figured out what the conflict was, i was editing a page that no longer existed. I dropped that commit, hopefully this one is good to go

agilgur5:
@allcontributors please add @jssee for code

PR #765, opened 2020-07-17
New docs site!

jaredpalmer:


agilgur5:
It would be good to state the purpose of a new docs site given that the old one was never linked to and never updated and users didn't really mind either.

PR #789, opened 2020-08-09
Upgrade from rollup-plugin-babel to @rollup/plugin-babel

vladdy-moses:


vladdy-moses:
Also fixed _Node CI / Lint on node 10.x and ubuntu-latest (pull_request)_ check.

jaredpalmer:
I thnk this closes #792

agilgur5:
@jaredpalmer er, not sure how those two are related... I just closed that one, it's a transient `core-js` postinstall error that shouldn't affect anything and is very upstream.
This PR doesn't change anything related to `core-js`. It happens to remove one of the deprecation notices logged in that issue but does not fix the central error there...

.github/workflows/nodejs.yml
```diff
@@ -19,6 +19,10 @@ jobs:
 
       - name: Install deps and build (with cache)
         uses: bahmutov/npm-install@v1
+      - name: Install deps and build (with cache) - website
+        uses: bahmutov/npm-install@v1
+        with:
+          working-directory: website
```
agilgur5, line 25: This seems very unrelated and the new website Jared added seems to have its own system set-up from CI (puts comments literally everywhere as seen in this PR, which is mildly annoying especially when no changes to the website were made...). I saw you wrote this:

> Also fixed _Node CI / Lint on node 10.x and ubuntu-latest (pull_request)_ check.

However if this fixes it, that doesn't seem to be intentional. There's also a change required to `src/index.ts` to fix the lint issue, so this seems to break things.
The correct fix is to just run `yarn lint --fix` -- I did so in a separate PR: #794

yarn.lock
```diff
@@ -1056,6 +1064,15 @@
...
     estree-walker "^1.0.1"
 
+"@rollup/pluginutils@^3.0.8":
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/@rollup/pluginutils/-/pluginutils-3.1.0.tgz#706b4524ee6dc8b103b3c995533e5ad680c02b9b"
+  integrity sha512-GksZ6pr6TpIjHm8h9lSQ8pi8BE9VeubNT0OMJ3B5uZJ8pz73NPiqOtCog/x2/QzM1ENChPKxMDhiQuRHsqc+lg==
+  dependencies:
+    "@types/estree" "0.0.39"
+    estree-walker "^1.0.1"
+    picomatch "^2.2.2"
```
agilgur5, line 1074: This might conflict / need to be de-duped with #678 ([specific change](https://github.com/formium/tsdx/pull/678/files#diff-8ee2343978836a779dc9f8d6b794c3b2R1059))

.github/workflows/nodejs.yml
agilgur5: Merged #794 now which fixes the lint errors properly. So if you remove this and rebase, Lint should still pass
```suggestion
```

agilgur5: Removed it myself

agilgur5:
Have taken over this myself since this has been stale for several weeks and is one of the last dep updates being added to [`v0.14.0`](https://github.com/formium/tsdx/milestone/2), which is just about ready otherwise. This also fits into the theme as an upgrade that requires Node 10+.

I've removed the change I requested to be removed, rebased the PR and fixed the `yarn.lock` and `package.json` conflicts, and tests are looking to pass now.

> I assume this is just to get rid of a deprecation notice, however it would be better not to assume.

Yup, seems to be so, got this warning on a fresh install of `v0.13.3`:
```bash
npm WARN deprecated rollup-plugin-babel@4.4.0: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-babel.
```

agilgur5:
Removed the two leftover references to `rollup-plugin-babel`, one was an old commented out import, another was a `declare module`, which is no longer necessary because `@rollup/plugin-babel` [ships its own typings](https://github.com/rollup/plugins/blob/master/packages/babel/CHANGELOG.md#v510).

agilgur5 (approved):

agilgur5:
@allcontributors please add @vladdy-moses for code

PR #800, opened 2020-08-15
fix GitHub Action templates' cache key by removing glob

### LINKED ISSUE #799: Github Actions templates cache doesn't hit due to hashing bug
### Current Behavior

Due to the fact that this [line](https://github.com/formium/tsdx/blob/master/templates/basic/.github/workflows/main.yml#L20) has a **glob**, it will create a wrong key!
Explanation, when the workflow starts it doesn't have any node_modules, therefore `hashFiles('**/yarn.lock')` will use only the package's `yarn.lock` (as expected), **but** at the end of the flow, node_modules are installed, `hashFiles('**/yarn.lock')` will contain **all** `yarn.lock` files from node_modules as well, therfore `hashFiles('**/yarn.lock')` will save the cache on a different key.

### Expected behavior

It should relay only on the package's `yarn.lock` file.

### Suggested solution(s)

Remove the glob from this line https://github.com/formium/tsdx/blob/master/templates/basic/.github/workflows/main.yml#L20.

### Additional context

Based on the research that @AllanChain did, https://github.com/AllanChain/blog/issues/98

### Your environment

| TSDX           | 
| TypeScript       |
| Browser          |
| npm/Yarn         |
| Node             | 
| Operating System |

felixmosh:


agilgur5:
per https://github.com/formium/tsdx/issues/799#issuecomment-695950119, this has been superseded by #882

PR #802, opened 2020-08-17
Use eslint-config-tsdx

### LINKED ISSUE #801: Use an eslint shareable config which can also be used independently of tsdx
### Current Behavior

Sometimes I want to use tsdx's wonderful eslint configuration, but don't need the other functions of `tsdx` (build & test), for whatever reason (see comment https://github.com/formium/tsdx/issues/634#issuecomment-673735730 for specific examples).

One way to do this is to just install the whole `tsdx` package but that will come with a large volume of unused dependencies.

Another way to do it is to just use [`eslint-config-tsdx`](https://www.npmjs.com/package/eslint-config-tsdx), an eslint shareable config with the rules of tsdx. This way we don't get the large volume of unused dependencies, but that package (right now) just mirrors the configuration that lives in `tsdx`, and the two can become out-of-sync.

### Desired Behavior

1. I would like to be able to install and use *just the eslint configuration* from tsdx, without the added weight of the other parts of `tsdx` (build & test).

2. I would also like to be sure I'm using the same eslint configuration in this scenario as when using tsdx itself.

### Suggested Solution

My proposed solution is simply to use `eslint-config-tsdx` (an eslint shareable config with the rules of tsdx, as mentioned above) here, in the `extends` field of the generated eslint config. This way we can achieve both things listed under "Desired Behavior".

See https://github.com/formium/tsdx/issues/634#issuecomment-673736473 for a more detailed/verbose description of my proposed solution.

I will put my proposed solution into the form of a PR to make it perfectly clear.

This is for users who want to use tsdx's linting functionality without using tsdx's other functionalities. This should only impact them; other users should not be impacted at all besides getting a simplified (but equivalent) output from `tsdx lint --write-file`.

Some options are described above (under Current Behavior) but have their drawbacks.

### Additional context

The use case for this feature is originally described here: https://github.com/formium/tsdx/issues/634#issuecomment-673735730

zenflow:
Closes #801 with the solution proposed there. This is the same solution originally proposed in this comment: https://github.com/formium/tsdx/issues/634#issuecomment-673736473

As I mentioned in the above linked comment, the eslint-config-tsdx package & repo should be transferred to the formium organization before merging this.

I don't believe this change to be breaking. We started a discussion on that here: https://github.com/formium/tsdx/issues/634#issuecomment-673747526


yarn.lock
```diff
@@ -9,6 +9,13 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
+"@babel/code-frame@^7.10.4":
```
agilgur5, line 12: It's not clear to me why this changed so much -- the tree of dependencies is the same aside from `eslint-config-tsdx`, so there shouldn't be that much change

package.json
```diff
@@ -67,14 +64,7 @@
...
     "enquirer": "^2.3.4",
     "eslint": "^6.1.0",
-    "eslint-config-prettier": "^6.0.0",
-    "eslint-config-react-app": "^5.0.2",
-    "eslint-plugin-flowtype": "^3.13.0",
-    "eslint-plugin-import": "^2.18.2",
-    "eslint-plugin-jsx-a11y": "^6.2.3",
-    "eslint-plugin-prettier": "^3.1.0",
-    "eslint-plugin-react": "^7.14.3",
-    "eslint-plugin-react-hooks": "^2.2.0",
```
agilgur5, line 77: I mentioned this in https://github.com/formium/tsdx/issues/634#issuecomment-674997895, but putting plugins and parsers as dependencies of a shareable config is explicitly not sanctioned by ESLint and against ESLint conventions, docs, etc. Other much more popular configs don't go against this convention either.

PR #823, opened 2020-08-23
feat/types: add main and types fields

cncolder:
- generate type difinition files
- tsdx is requireable now
  - so user can put `TsdxOptions` into tsdx.config.js JSDoc

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,15 +324,14 @@ The `options` object type definition in [src/externalTypes.ts](src/externalTypes
 #### Example: Adding Postcss
 
 ```js
-const tsdx = require('tsdx');
 const postcss = require('rollup-plugin-postcss');
 const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
 
 module.exports = {
   /**
-   * @param {tsdx.RollupOptions} config
-   * @param {tsdx.TsdxOptions} options
+   * @param {import('tsdx').RollupOptions} config
+   * @param {import('tsdx').TsdxOptions} options
    */
   rollup(config, options) {
     config.plugins.push(
--- a/test/integration/fixtures/build-withConfig/tsdx.config.js
+++ b/test/integration/fixtures/build-withConfig/tsdx.config.js
@@ -1,13 +1,11 @@
-/* eslint-disable-next-line no-unused-vars */
-const tsdx = require('..');
 const postcss = require('rollup-plugin-postcss');
 const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
 
 module.exports = {
   /**
-   * @param {tsdx.RollupOptions} config
-   * @param {tsdx.TsdxOptions} options
+   * @param {import('tsdx').RollupOptions} config
+   * @param {import('tsdx').TsdxOptions} options
    */
   rollup(config, options) {
     config.plugins.push(
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   "bugs": {
     "url": "https://github.com/formium/tsdx/issues"
   },
-  "main": "./dist/externalTypes.js",
   "typings": "./dist/externalTypes.d.ts",
   "bin": {
     "tsdx": "./dist/index.js"


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "importHelpers": true,
     "esModuleInterop": true,
     "outDir": "dist",
-    "declaration": false,
+    "declaration": true,
```
agilgur5, line 10:
```suggestion

```

[the default is `true`](https://www.typescriptlang.org/tsconfig#declaration)

src/externalTypes.ts
```diff
@@ -0,0 +1,27 @@
...
+  env: 'development' | 'production';
+  /** Module format */
+  format: ModuleFormat;
+  /** Is minifying? */
+  minify?: boolean;
+  /** Is this the very first rollup config (and thus should one-off metadata be extracted)? */
+  writeMeta?: boolean;
+  /** Only transpile, do not type check (makes compilation faster) */
+  transpileOnly?: boolean;
+}
```
agilgur5, line 27:
```suggestion
export { TsdxOptions } from './types'
```

Is there a reason you didn't just re-export that one type?

Otherwise, if just creating a single declaration file, the preference would be to just use a `.d.ts` file and not add any other declarations

cncolder: If re-export from './types'. User can explore more types which you don't want show public. e.g. Cmd+Click in vscode.

cncolder:
So now. What's your favor @agilgur5 ?

1. Active "declaration" in tsconfig.json. Then re-export `TsdxOptions` from types.ts.
2. Disable "declaration". And add a new file to npm like `types/index.d.ts`. But we still need main field in package.json

test/integration/fixtures/build-withConfig/tsdx.config.js
```diff
@@ -1,8 +1,14 @@
+/* eslint-disable-next-line no-unused-vars */
```
agilgur5, line 1: should add this to the example in the README too since they both use `tsdx lint`

README.md
```diff
@@ -349,11 +349,16 @@ export interface TsdxOptions {
 #### Example: Adding Postcss
```
agilgur5, line 349: I can't comment on the lines above this, but could you replace the typedef above with a relative link to the type file instead? That's what I meant when I said "Also replace the `TsdxOptions` snippet there to a link to the typing file"

src/externalTypes.ts
agilgur5: Yes but they can do that anyway if `declaration` is on. Nothing else is exported from `tsdx`, the rest have to be pulled from `tsdx/dist/` which is an undocumented and implicitly private API, so "use at your own risk" is implied there.

src/externalTypes.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export { RollupOptions } from 'rollup';
```
agilgur5, line 29: I was thinking of using the _exact_ type that TSDX modifies this too, although that may not be possible to override

test/integration/fixtures/build-withConfig/tsdx.config.js
cncolder: `tsdx lint` will skip `tsdx.config.js`

tsconfig.json
cncolder: Typescript docs is wrong. You can try delete this line then no `.d.ts` produced.

README.md
cncolder: done

tsconfig.json
agilgur5: Thanks for pointing this out, made a PR upstream to fix those docs: https://github.com/microsoft/TypeScript-Website/pull/1092

README.md
```diff
@@ -319,41 +319,21 @@ module.exports = {
...
-  // Only transpile, do not type check (makes compilation faster)
-  transpileOnly?: boolean;
-}
-```
+The `options` object type definition in [src/externalTypes.ts](src/externalTypes.ts).
 
 #### Example: Adding Postcss
 
 ```js
+const tsdx = require('tsdx');
```
agilgur5, line 327: So I recently found out that JSDoc actually [supports imports](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types) after seeing [some usage of it](https://github.com/developit/rollup-plugin-preserve-shebang/blob/cc98df81e83bb19d5e6e469bce77a93f577da045/index.mjs#L7), so we should be able to remove this entirely that way, and perhaps remove `package.json#main` as well
```suggestion
```

README.md
```diff
@@ -319,41 +319,21 @@ module.exports = {
...
 const postcss = require('rollup-plugin-postcss');
 const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
 
 module.exports = {
+  /**
+   * @param {tsdx.RollupOptions} config
+   * @param {tsdx.TsdxOptions} options
```
agilgur5, line 335:
```suggestion
   * @param {import('tsdx').RollupOptions} config
   * @param {import('tsdx').TsdxOptions} options
```

src/externalTypes.ts
agilgur5: Nevermind, TSDX itself uses `RollupOptions` internally.

agilgur5 (commented):
I'm working on a refactor of some of the types, and in the process seem to have found a better usage

src/externalTypes.ts
agilgur5: So I ran `tsc` over a `tsdx.config.js` with the proper import and a `// @ts-check` and `allowJs: "true"` etc and that actually failed because of the lack of exact typing. E.g.:

```bash
$ tsc --noEmit
tsdx.config.js:13:5 - error TS2532: Object is possibly 'undefined'.

13     config.plugins.push(
```

Now the problem is that I'm not entirely sure how to do this in TS or if it's possible to. TSDX's Rollup config should _conform_ to `RollupOptions` but should also be more specific than `RollupOptions` -- e.g. we should _know_ that `config.plugins.push` is possible without any type-casting.

Of course, TSDX Rollup config could have its own interface that _extends_ `RollupOptions`, but that would be the naive route and could still cause users some type-checking problems. The config is complex enough that I'd like it to be inferred (especially the plugins' types, since users may override them) while still conforming to `RollupOptions` (so we don't get problems like #371 ). I'm not sure if that's possible in TS or if I just haven't been able to figure out how to do it in TS.

test/integration/fixtures/build-withConfig/tsdx.config.js
```diff
@@ -3,6 +3,10 @@ const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
 
 module.exports = {
+  /**
+   * @param {import('tsdx').RollupOptions} config
+   * @param {import('tsdx').TsdxOptions} options
```
agilgur5, line 8:
```suggestion
   * @param {import('../../../../').RollupOptions} config
   * @param {import('../../../../').TsdxOptions} options
```

You had `require('..')` before, so it should either be that (`'..'`) or this, which works in VSCode but may not work during the test phase where this gets moved to the "staging" dir. But `tsdx build` can't type-check this anyway (otherwise we'd be using `tsdx.config.ts`). Might add `tsc` tests in the future for things like #871 though...

test/integration/fixtures/build-withConfig/tsdx.config.js
```diff
@@ -3,6 +3,10 @@ const autoprefixer = require('autoprefixer');
 const cssnano = require('cssnano');
```
agilgur5, line 3: Can't comment at the top of the file, but per my other comment, we probably want to add `// @ts-check` to the top of this and the example file.

Adding `allowJs` + `checkJs` to templates' and the tests' `tsconfig` probably makes more sense long-term, but there will probably be some ramifications to that and may want to wait to do that until #871 is fixed and creates two `tsconfig`s, one for `build` and one for type-checking. I definitely need to test that more before adding.

agilgur5 (commented):
Thanks for all the iterations despite the long wait in between! I think this _close_ to complete, but I've found more issues as I've dug into it and tried it myself 😐  .

I could take this over completing this myself now though (you'd of course be credited regardless for this work!), especially as some of these issues likely require some internal typing refactoring (which I've already started too), but if you know more about some TS nuances in the comments, that could be quite helpful!

PR #838, opened 2020-08-25
deps: upgrade Babel preset-env, remove now redundant plugins

agilgur5:
## Description

- optional-chaining and nullish-coalescing are now part of
  `@babel/preset-env`, so no need to explicitly install or configure
  them anymore
  - left in the regression tests though!
  - previously they were new syntax and TSDX adopted them as soon as
    they landed in TS
  - can view current compat table here:
    https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-compat-data/data/plugins.json#L24

- update autoprefixer integration test because browserslist and
  caniuse-lite were updated to more recent compat tables
  - `-webkit-input-placeholder` is no longer needed, but some others
    are, so switch to checking for `-moz-placeholder`
  - this probably fixes that "caniuse-lite is out-of-date" warning I
    was getting occassionally in CI
    - didn't get locally probably because I didn't install with
      `--frozen-lockfile` locally, unlike in CI

## Tags

#370 added this shortly after the syntax landed in TS.

## Misc Notes

Been meaning to do this for a while and finally got to it


agilgur5:
Huh... `rollup-plugin-postcss` integration test is failing now... instead of the syntax regression tests... well that's odd

agilgur5:
Yea it's just completely ignoring the CSS import in the output... wut...

agilgur5:
Ohhh it changed the output of autoprefixer a bit... I guess because the `browserslist` / `caniuse` compat table was updated to something more reflective of today and so `-webkit-input-placeholder` is no longer necessary.

PR #846, opened 2020-08-31
env/deps: remove greenkeeper.json, configure dependabot.yml

### LINKED ISSUE #839: Setup dependabot.yml
GitHub recently acquired Dependabot and there is a new GitHub-native way to do the setup.

https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

@agilgur5 docs seem pretty straightforward. I enabled it on tsdx via the dependabot app, but it appears that the yaml config allows for more granular control over things. I think we can safely set it to for weekly checks. As for rest of setup (ie auto-rebasing and automerging and pr limits), I will leave those decisions to you.

agilgur5:
## Description

- Greenkeeper has shut down, Snyk has been incredibly annoying and can't
  be configured in the codebase (installed by Jared and could only be
  configured by Jared), so use native Dependabot now
  - GitHub acquired Dependabot a little over a year ago and it is what
    powers GitHub vulnerability updates

- set-up sensible defaults with YAML anchor/alias
  - to only make PRs weekly, not spam daily
  - to only make PRs for deps, not devDeps
  - to only increase version when necessary, not for every patch and
    minor bump when a dep isn't pinned anyway
  - to use "deps:" prefix similarly to what I use

- set-up Dependabot to ignore /website entirely, for dep upgrades and
  vulnerabilities, as it is not a published package and doesn't really
  have an attack surface area
  - should only be updated as needed, not whenever a dep is upgraded

- temporarily ignore "/" as well because it's currently being updated so
  don't want duplication spam
  - but leave security PRs on, only dep upgrades off

## Tags

Fixes #839 
Follow-up to #815 which didn't work and [Snyk's removal](https://github.com/formium/tsdx/pull/770#issuecomment-679967132).

Closing a few PRs [as unnecessary](https://github.com/formium/tsdx/pull/807#issuecomment-679448859) since they update the `/website` dir, which isn't a published library and doesn't really have an attack surface:
- Closes #818 
- Closes #786


agilgur5:
Test failure is a timeout that's been happening occasionally on macOS runs on GitHub (not limited to this repo), overriding and merging

PR #858, opened 2020-09-16
Add `--max-warnings` flag to `tsdx lint`

### LINKED ISSUE #853: lint flag to make warnings exit with non-zero
### Current Behavior

If you have lint warnings and run `yarn lint` or `tsdx lint`, the warnings are printed but the command exits without an error

### Desired Behavior

It would be nice to have `tsdx lint` take a flag to exit with a non-zero code if warnings are found so that a pre-commit hook can prevent a commit from going through and require the user to fix the warnings.

This is for people who want to force all lint errors _and_ warnings to be cleared before committing code.

georgevarghese185:
Fixes #853 

Works in a similar way to the --max-warnings flag in ESLint.

If `tsdx lint` finds more than `--max-warnings` number of warnings, it exits with a non-zero code. Otherwise, it exits with 0.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -547,7 +547,8 @@ prog
   .example('lint src test --ignore-pattern test/foobar.ts')
   .option(
     '--max-warnings',
-    'Exits with non-zero error code if warnings exceed this number'
+    'Exits with non-zero error code if warnings exceed this number',
+    Infinity
   )
   .example('lint src test --max-warnings 10')
   .option('--write-file', 'Write the config file locally')
@@ -560,7 +561,7 @@ prog
       'ignore-pattern': string;
       'write-file': boolean;
       'report-file': string;
-      'max-warnings': number | undefined;
+      'max-warnings': number;
       _: string[];
     }) => {
       if (opts['_'].length === 0 && !opts['write-file']) {
@@ -603,10 +604,7 @@ prog
       if (report.errorCount) {
         process.exit(1);
       }
-      if (
-        opts['max-warnings'] !== undefined &&
-        report.warningCount > opts['max-warnings']
-      ) {
+      if (report.warningCount > opts['max-warnings']) {
         process.exit(1);
       }
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -545,6 +545,11 @@ prog
   .example('lint src test --fix')
   .option('--ignore-pattern', 'Ignore a pattern')
   .example('lint src test --ignore-pattern test/foobar.ts')
+  .option(
+    '--max-warnings',
+    'Exits with non-zero error code if warnings exceed this number'
+  )
+  .example('lint src test --max-warnings 10')
   .option('--write-file', 'Write the config file locally')
   .example('lint --write-file')
   .option('--report-file', 'Write JSON report to file locally')
--- a/src/index.ts
+++ b/src/index.ts
@@ -555,6 +555,7 @@ prog
       'ignore-pattern': string;
       'write-file': boolean;
       'report-file': string;
+      'max-warnings': number | undefined;
       _: string[];
     }) => {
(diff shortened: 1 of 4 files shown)


test/e2e/fixtures/lint/file-with-lint-warnings.ts
```diff
@@ -0,0 +1,4 @@
+const foo = () => {
+  const bar = 'baz';
+  const foobar = '';
+};
```
agilgur5, line 4:
```suggestion
// this file should have 3 unused var lint warnings
const unusedVar1 = () => {
  const unusedVar2 = 'baz';
  const unusedVar3 = '';
};
```

The lint fixtures don't seem to have comments, but some other fixtures do and it would be useful to have so it's easier to interpret at-a-glance. Changed naming to be similarly easier to interpret rather than random

test/e2e/tsdx-lint.test.ts
```diff
@@ -30,6 +30,48 @@ describe('tsdx lint', () => {
...
+    const testFile = `${lintDir}/file-with-lint-warnings.ts`;
+    const output = shell.exec(
+      `node dist/index.js lint ${testFile} --max-warnings 2`
+    );
+    expect(output.code).toBe(1);
+    expect(output.stdout.includes('@typescript-eslint/no-unused-vars')).toBe(
+      true
+    );
+  });
+
```
agilgur5, line 74: Let's move this under the `tsx` tests but before "`should not lint`" test so that the lint error tests are grouped together as are the lint warning tests

src/index.ts
```diff
@@ -555,6 +560,7 @@ prog
       'ignore-pattern': string;
       'write-file': boolean;
       'report-file': string;
+      'max-warnings': number | undefined;
```
agilgur5, line 563:
```suggestion
      'max-warnings': Infinity;
```

Seems like an accurate default to me

src/index.ts
```diff
@@ -545,6 +545,11 @@ prog
   .example('lint src test --fix')
   .option('--ignore-pattern', 'Ignore a pattern')
   .example('lint src test --ignore-pattern test/foobar.ts')
+  .option(
+    '--max-warnings',
+    'Exits with non-zero error code if warnings exceed this number'
```
agilgur5, line 550:
```suggestion
    'Exits with non-zero error code if number of warnings exceed this number',
    Infinity
```

sets the default to go along with the interface below

src/index.ts
```diff
@@ -597,6 +603,12 @@ prog
       if (report.errorCount) {
         process.exit(1);
       }
+      if (
+        opts['max-warnings'] !== undefined &&
+        report.warningCount > opts['max-warnings']
+      ) {
```
agilgur5, line 609:
```suggestion
      if (report.warningCount > opts['max-warnings']) {
```

if `Infinity` is the default for `max-warnings` then we can simplify this a bit

agilgur5 (changes requested):
Thanks for the PR @georgevarghese185 ! Great work and comprehensive testing! Have a few suggestions below to improve this, but is otherwise good.

Was going to add that you need to have the `option` and `example`, but I see you figured out that piece yourself.

One thing missing is that the API Reference section in the README needs to be updated with the new output of `tsdx lint --help` with this change

georgevarghese185:
Hey, @agilgur5 !

Thanks for the suggestions (and for being super polite 😃)

I think I've addressed everything. Take another look and let me know if I've missed something or if it can be improved further.

src/index.ts
agilgur5: ah I see you can't use default args in an interface like this, sorry for the incorrect recommendation

agilgur5:
@allcontributors please add @georgevarghese185 for code, docs, test

PR #864, opened 2020-09-17
Add skipLibCheck, noEmit, consistentCasing, and lots of comments to template tsconfigs

### LINKED ISSUE #529: noEmit and skipLibCheck should be on by default in TS compiler options
This is a related issue to #352. The build script works with rollup.js, not `tsc`, so the `noEmit` option should be enabled by default. Users running `npx tsc` shouldn't expect to have to specify `--noEmit`, because using `tsc` is a normal part of TS development. Currently `npx tsc` will clutter the `src` directory with transpiled files, which can be difficult to clean up if there are uncommitted changes mixed in.

This is important because without `tsc` there is no way to do project-wide type checking. The build script will only stop at the first error. The test runner will report errors, but only for files that are tested, and it's also the wrong place for type checking (#521).

### LINKED ISSUE #619: Redundant compiler options
There are a number of redundant compiler options in the current `tsconfig.json`, and it could be cleaned up or simplified by removing them.

`strict` already enables `noImplicitAny`, `noImplicitThis`, `alwaysStrict`, `strictNullChecks`, `strictFunctionTypes` and `strictPropertyInitialization`.

`noUnusedLocals` and `noUnusedParameters` overlap with the `@typescript-eslint/no-unused-vars` ESLint rule, which results in duplicate errors in the editor (after the user fixes [this notable issue](https://github.com/jaredpalmer/tsdx/issues/354)):

image

This means, for example, that using the Go to Next Problem or Go to Previous Problem in the editor require being used twice to move past the same error.

agilgur5:
#### feat: add skipLibCheck and forceConsistentCasingInFileNames
- as defaults to the templates' tsconfigs
  - these are now recommended by TS to be set to true, which can be
    seen in the TSConfig Reference and @tsconfig/recommended
    - c.f. https://www.typescriptlang.org/tsconfig and
      https://www.npmjs.com/package/@tsconfig/recommended
  - also add to all fixtures
    - yea some centralization/extends or dogfooding create would help
      reduce all this duplication

- skipLibCheck is a big perf gain, though it seems to only affect
  `tsc --noEmit` usage and rollup-plugin-typescript2 seems to ignore it
  - I have an existing issue open on the matter
- forceConsistentCasingInFileNames is a no-brainer to help avoid a bug
  that is easy to make and hard to debug

- also add a trailing comma for ease of use and better diffing
  - not to fixtures though, because there is a test for this

#### feat: add noEmit to templates' tsconfigs
- for ease of use when using `tsc` for type-checking, which a lot of
  users already do
  - it doesn't fully type-check everything though, since we'd need to
    split off a tsconfig.build.json in order to let builds only run on
    `src` while type-checking would run on all files
    - but people already using `tsc --noEmit`, so might as well start
      supporting this as it's not uncommon

- also add to all fixtures
  - yea some centralization/extends or dogfooding create would help
    reduce all this duplication

#### docs: add comments for nearly all tsconfig options in use
- since we can use comments in tsconfig.json, add them to give simple,
  one-liner explanations as to what each setting does
- also add a link to the TSConfig Reference for further details
- and specify which options are recommended as true by TS but not on by
  default for backwards compat or other reasons

- a few leftover that I still need to think about

#### docs: add comment that noUnused* overlaps with an ESLint rule
- the checks overlap with the @typescript-eslint/no-unused-vars rule
- per user issue, this can result in duplicative errors inside an IDE,
  so explicitly call this out in a comment for `tsdx lint` / ESLint
  users in case they'd like to disable this and leave only one on

- change the ordering of linter checks in the templates as well to make
  the comments easier to read/understand

## Tags

- Fixes #529 as requested, since, per https://github.com/formium/tsdx/issues/529#issuecomment-679746098 these are now the defaults upstream as well
- Fixes #619 , the second part about duplication with ESLint, by adding a comment as suggested in https://github.com/formium/tsdx/issues/619#issuecomment-600227877

- Comments and trailing comma are a result of #483 / #489 , since I didn't realize this was possible before
  - I've written in several issues in a few places (can't remember them all now) that I think comments would be useful, so have finally gotten to it for _most_ of the `tsconfig` (but not quite all, and not including some forced or commented out options, i.e. https://github.com/formium/tsdx/issues/760#issuecomment-692979270)
  - `tsc --init` has comments too as existing precedent

## Review Notes

Key thing to check would be that all 3 templates and all 6 fixtures are similarly updated

### DIFF
--- a/templates/basic/tsconfig.json
+++ b/templates/basic/tsconfig.json
@@ -14,10 +14,11 @@
     // stricter type-checking for stronger correctness. Recommended by TS
     "strict": true,
     // linter checks for common issues
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    // noUnused* overlap with @typescript-eslint/no-unused-vars, can disable if duplicative
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
     // use Node's module resolution algorithm, instead of the legacy TS one
     "moduleResolution": "node",
     // transpile JSX to React.createElement
--- a/templates/basic/tsconfig.json
+++ b/templates/basic/tsconfig.json
@@ -17,5 +17,6 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
+    "noEmit": true,
   }
 }
--- a/templates/basic/tsconfig.json
+++ b/templates/basic/tsconfig.json
@@ -1,22 +1,34 @@
 {
+  // see https://www.typescriptlang.org/tsconfig to better understand tsconfigs
   "include": ["src", "types"],
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
     "importHelpers": true,
+    // output .d.ts declaration files for consumers
     "declaration": true,
+    // output .js.map sourcemap files for consumers
     "sourceMap": true,
+    // match output dir to input dir. e.g. dist/index instead of dist/src/index
     "rootDir": "./src",
+    // stricter type-checking for stronger correctness. Recommended by TS
     "strict": true,
+    // linter checks for common issues
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    // use Node's module resolution algorithm, instead of the legacy TS one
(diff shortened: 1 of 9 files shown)


agilgur5 (commented):
LGTM 👍  Diff was fairly straightforward to go through due to spacing between comments making it easier on the eyes

agilgur5:
Failing CI check is another one of those GH Actions macOS timeouts, not an actual test failure. Should be good to go!

PR #870, opened 2020-09-19
fix: remove redundant CI env from workflow steps

leonardodino:
Thanks for adding GH actions to the templates,
it's awesome and greatly reduces overhead of adding CI on new packages.
I always end up needing to read the damm docs again for every project 🙈 

**nitpick** fix on this PR:
`CI=true` is already declared at runner level.

docs: [default-environment-variables](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)

### DIFF
--- a/templates/basic/.github/workflows/main.yml
+++ b/templates/basic/.github/workflows/main.yml
@@ -23,20 +23,12 @@ jobs:
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
-        env:
-          CI: true
 
       - name: Lint
         run: yarn lint
-        env:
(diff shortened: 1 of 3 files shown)


agilgur5:
Hmm, yea not sure why these were added in #457. It's not used in [TSDX's own internal actions](https://github.com/formium/tsdx/blob/22133ce025f5f3a80cc797b15cdec3960b2ba3ae/.github/workflows/nodejs.yml).

Good catch and thanks for the helpful link to the docs @leonardodino !

agilgur5 (approved):

PR #805, opened 2020-08-18
update react-with-storybook template for Storybook v6

HipsterBrown:
With the release of [Storybook v6](https://storybook.js.org/releases/6.0), there are [updates to be made](https://storybook.js.org/releases/6.0) to migrate from v5.3 and allow the `react-with-storybook` template take advantage of the latest features, i.e. built-in Typescript support and new story format.

PR #876, opened 2020-09-21
feat: type-check stories during Storybook build

### LINKED ISSUE #668: stories dir is not being type checked during Storybook build
### Current Behavior

If I have bad typescript code in `/stories/button.stories.tsx` (like the following code) I will not get a typescript build error when I'm either watching storybook or building storybook:

```
const falseAlways = false;
falseAlways = true; // <-- shouldn't able to reassign a const & also a false literal can't be set to true.
```

### Expected behavior

I should get a build error when running `build-storybook` that says: `Cannot assign to 'falseAlways' because it is a constant.ts(2588)`

### Suggested solution(s)

Modify the `.storybook/main.js` in the templates and or the `tsconfig.json` so that `ts-loader` knows that it should be type checking the files.

### Additional context

This might be a problem due to the `noEmit` flag in `tsconfig.json`; however, I've tried turning file emitting back on and just simple directing it (via `outDir`) to a different directory.

I've used the default main.js but I've also tried adding to it to see if I can get better feedback:
```js
const WebpackNotifierPlugin = require('webpack-notifier');
const WarningsToErrorsPlugin = require('warnings-to-errors-webpack-plugin');

module.exports = {
  stories: ['../stories/**/*.stories.(ts|tsx)'],
  addons: [
    '@storybook/addon-actions',
    '@storybook/addon-links',
    '@storybook/addon-docs',
  ],
  webpackFinal: async config => {

    console.log(`#####################################
      Please forgive the wait. The startup time is slow but the watcher time should be quite quick.
    #####################################`);

    const compilerOptions = require("../tsconfig.build.json").compilerOptions;

    if(!compilerOptions){
      throw new Error("Could not find tsconfig.json");
    }

    // include the stories so we can get type checking
    compilerOptions.rootDirs = compilerOptions.rootDirs || [];
    if(compilerOptions.rootDir){
      console.warn("Modifying rootDir to be a rootDirs array so we can get type checking for the stories");
      compilerOptions.rootDirs.push(compilerOptions.rootDir);
      compilerOptions.rootDir = "..";
    }
    compilerOptions.rootDirs.push("./stories");
    compilerOptions.outDir = "./storybook-static";
    compilerOptions.isolatedModules = false;
    // When building for storybook we need to override this setting or we'd get an error
    delete compilerOptions.moduleResolution;


    config.module.rules.push({
      test: /\.(ts|tsx)$/,
      use: [
        {
          loader: require.resolve('ts-loader'),
          options: {
            compilerOptions
          }
        },
        {
          loader: require.resolve('react-docgen-typescript-loader'),
          options: {
            compilerOptions
          }
        },
      ],
    });
    config.module.rules.push({
      test: /\.s[ac]ss$/i,
      use: [
        // Creates `style` nodes from JS strings
        'style-loader',
        // Translates CSS into CommonJS
        'css-loader',
        // Compiles Sass to CSS
        'sass-loader',
      ],
    });

    config.resolve.extensions.push('.ts', '.tsx');

    config.plugins = config.plugins || [];
    config.plugins.push(
      new WebpackNotifierPlugin({ title: 'Custom Storybook Webpack' })
    );
    config.plugins.push(new WarningsToErrorsPlugin());

    return config;
  },
};

```

### Your environment

| TSDX           | 0.13.1 |
| TypeScript       | 3.8.3 |
| Browser          | n.a. but Chrome |
| npm/Yarn         | npm 6.13.4 |
| Node             |  12.15.1 |
| Operating System | Windows 10 |

kylemh:
Resolves #668

### DIFF
--- a/templates/react-with-storybook/.storybook/main.js
+++ b/templates/react-with-storybook/.storybook/main.js
@@ -1,6 +1,7 @@
 module.exports = {
   stories: ['../stories/**/*.stories.@(ts|tsx|js|jsx)'],
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  // https://storybook.js.org/docs/react/configure/typescript#mainjs-configuration
   typescript: {
     check: true, // set to false if you don't care about your stories being type-checked.
   }
--- a/templates/react-with-storybook/.storybook/main.js
+++ b/templates/react-with-storybook/.storybook/main.js
@@ -1,4 +1,7 @@
 module.exports = {
   stories: ['../stories/**/*.stories.@(ts|tsx|js|jsx)'],
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  typescript: {
+    check: true, // set to false if you don't care about your stories being type-checked.
+  }
 };
--- a/templates/react-with-storybook/.storybook/main.js
+++ b/templates/react-with-storybook/.storybook/main.js
@@ -3,6 +3,6 @@ module.exports = {
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
   // https://storybook.js.org/docs/react/configure/typescript#mainjs-configuration
   typescript: {
-    check: true, // set to false if you don't care about your stories being type-checked.
+    check: true, // type-check stories during Storybook build
   }
 };


templates/react-with-storybook/.storybook/main.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   stories: ['../stories/**/*.stories.@(ts|tsx|js|jsx)'],
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  typescript: {
```
agilgur5, line 4:
```suggestion
  // https://storybook.js.org/docs/react/configure/typescript#mainjs-configuration
  typescript: {
```
Add a link to the docs for this config as we did with `preview.js` in #805 ; think docs links like these are generally useful, particularly for examples like the templates.

templates/react-with-storybook/.storybook/main.js
```diff
@@ -1,4 +1,7 @@
...
+    check: true, // set to false if you don't care about your stories being type-checked.
```
agilgur5, line 5:
```suggestion
    check: true, // type-check stories during Storybook build
```

"if you don't care" is really strong wording that I think we should generally avoid, but also it's not particularly helpful -- why would one not care? (a perf need is the main one). 

But rather than enumerating all reasons to not set it, let's just explain what the option does for now

agilgur5 (changes requested):
Thanks for the quick patch @kylemh 🙂  Some small comments changes below

agilgur5 (approved):
Thanks!

PR #878, opened 2020-09-21
deps: update extractErrors plugin to Babel 7 and more

agilgur5:
## Description

This plugin was out-of-date by a good bit, so have updated a good bit to a more current version.

#### license: add FB License header to all extractErrors code
- this plugin is largely borrowed from the React monorepo, consistently
  add FB's License headers to each file, not just to extractErrors.ts

- also add comments pointing to the exact file and commit that each
  file was copied from
  - most of these are slightly out of date now:
    - evalToString is missing a line
    - extractErrors needs to be updated to Babel 7
      - babylon -> @babel/parser, babel-traverse -> @babel/traverse
    - transformErrorMessages has several new constructs added :/...
      - keeping a pseudo-fork up-to-date is pretty tedious and not very
        maintainble (it's also broken per the tests I've added)
        - would be better if we could ask FB to split out theirs as a
          separate package...

#### clean: bad whitespace and Flow types in extractErrors
- whitespace errors in extractErrors
- unnecessary Flow comments in invertObject when we're using TS
  - also inconsistent, it looks like they've been removed and replaced
    with TS elsewhere
    - here TS types were added but Flow was also left in

#### deps: update extractErrors Babel plugins to Babel 7
- this was using Babel 6's Babylon and babel-traverse, which meant two
  different versions of Babel were being used and a lot of unnecessary
  deps were in the tree
  - now there should no longer be any Babel 6 deps in the tree
  - also transformErrorMessages was already using Babel 7
    (`@babel/helper-module-imports`) while extractErrors was on Babel 6
    just for extra inconsistency 😬
    - these two files seemed to be updated at the same time to Babel 7
      as far as I can tell though 🤨

- `@babel/parser` ships its own typings now, so no need to use the
  `declare module` workaround for it anymore
  - but did have to add a type-cast as its usage doesn't entirely fit
    the type 😕

- this also helps remove an old version of core-js v2, the addition of
  which would give a deprecation warning on install:
  - "npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3."

#### deps: update extractErrors plugin's evalToString file
- a line was added upstream in the last year, so add it here too to be
  up-to-date
  - now only transformErrorMessages is out-of-date ...but there's a
    little too many changes for me to really prioritize right now

## Tags

Follow-up to #138 I suppose

Some React PRs that updated this part of their monorepo:
- Babel 7: https://github.com/facebook/react/pull/16297 
- Babylon -> `@babel/parser`: https://github.com/facebook/react/pull/17988
- `Literal` addition in `evalToString`: https://github.com/facebook/react/pull/17599

TODO to be added to `transformErrorMessages`... well I'd prefer asking FB/React to split out their code as a separate package than maintaining this though:
- https://github.com/facebook/react/pull/16963
- https://github.com/facebook/react/pull/17063

In tandem with #795 which removed the unused `@babel/polyfill` dep (and its `core-js` v2 dep), this should remove the last deprecation warning I know of that TSDX `v0.13.3` has from https://github.com/formium/tsdx/pull/789#issuecomment-695854530 by removing the last `core-js` v2 usage:

```bash
npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3. // @babel/polyfill and babel-traverse
```


agilgur5 (commented):
LGTM 👍  . Dependency removal gr8

PR #879, opened 2020-09-21
deps: upgrade several more deps' majors

agilgur5:
## Description

- @types/node v13 -> v14
  - Node 13.x was EoL at the beginning of June per
    https://github.com/nodejs/Release
- @types/fs-extra was recently updated to v9, so can bump it to match
  the recent upgrade of fs-extra itself to v9
- rollup-plugin-sourcempas v0.5 -> v0.6
  - requires Node 10+, which we've already bumped to in the last dep
    upgrade, otherwise no breaking changes

- @rollup/plugin-node-resolve v7 -> v9
  - requires Node 10+, which we've already bumped to in the last dep
    upgrade, other breaking changes don't affect our usage
  - use the new `DEFAULTS` variables to pull out the extensions instead
    of having a comment referencing them
    - rename this `DEFAULTS` to `RESOLVE_DEFAULTS` and Babel's
     `DEFAULT_EXTENSIONS` to `DEFAULT_BABEL_EXTENSIONS` so that there's
      less naming confusion around these

- fix execa being pinned to v4.0.0 exactly
  - this was probably accidentally added as such because sometimes
    `yarn add dep@x` sets it exactly instead of to the caret version

## Tags

This is basically a follow-up to #678 , including fixing the accidental exact `execa` pin created. Also adds the `@types/fs-extra` upgrade that didn't exist back then.

Closes a few related Greenkeeper PRs:
- Closes #703 
- Closes #717 
- Closes #721 
- Closes #726


src/createRollupConfig.ts
```diff
@@ -116,8 +118,7 @@ export async function createRollupConfig(
           'main',
           opts.target !== 'node' ? 'browser' : undefined,
         ].filter(Boolean) as string[],
-        // defaults + .jsx
-        extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
+        extensions: [...RESOLVE_DEFAULTS.extensions, '.jsx'],
```
agilgur5, line 121: so this uses `.ext` while Babel uses `ext`. that's confusing, but correct

yarn.lock
```diff
@@ -3449,6 +3435,21 @@ execa@^3.2.0:
     signal-exit "^3.0.2"
     strip-final-newline "^2.0.0"
 
+execa@^4.0.0, execa@^4.0.3:
```
agilgur5, line 3438: this was moved to the correct place; previously it was above the ones that had numbers lower than it

yarn.lock
```diff
@@ -7173,13 +7174,20 @@ resolve@1.1.7:
   resolved "https://registry.yarnpkg.com/resolve/-/resolve-1.1.7.tgz#203114d82ad2c5ed9e8e0411b3932875e889e97b"
   integrity sha1-IDEU2CrSxe2ejgQRs5ModeiJ6Xs=
 
-resolve@1.15.1, resolve@1.x, resolve@^1.1.6, resolve@^1.10.0, resolve@^1.11.0, resolve@^1.12.0, resolve@^1.13.1, resolve@^1.14.2, resolve@^1.15.1, resolve@^1.3.2, resolve@^1.5.0:
+resolve@1.15.1:
```
agilgur5, line 7177: this is because rpts2 pins exact deps... and bundles them too... https://github.com/ezolenko/rollup-plugin-typescript2/pull/80 😕 😕

yarn.lock
```diff
@@ -7578,6 +7586,14 @@ source-map-resolve@^0.5.0, source-map-resolve@^0.5.3:
     source-map-url "^0.4.0"
     urix "^0.1.0"
 
+source-map-resolve@^0.6.0:
```
agilgur5, line 7589: `rollup-plugin-sourcemaps` v0.6 uses this version as can be seen above

agilgur5 (commented):
Some confusing changes, but they're all correct. LGTM 👍

agilgur5:
Deleted all the stale old Greenkeeper branches associated with all versions of the deps upgraded: `@types/node`, `@types/fs-extra`, `rollup-plugin-sourcemaps`, `@rollup/plugin-node-resolve`

PR #880, opened 2020-09-21
ci: update matrix to use Node 14 instead of Node 13

agilgur5:
## Description

- Node 13.x was EoL at the beginning of June per
  https://github.com/nodejs/Release
  - similarly we just recently updated @types/node to v14

## Tags

`@types/node` was updated to v14 in #879

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -31,7 +31,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


agilgur5 (commented):
👍  Couldn't find any other place Node 13 might be referenced.

Also, it be working:

Screen Shot 2020-09-21 at 1 55 09 AM

Issue #759, opened 2020-07-16
UMD format with lodash breaks due to lodash-es substitution


zhaoyao91:
### Current Behavior

To export umd module, I would like to include all modules into the bundle, so I config the rollup

```
    if (config.output.format === 'umd') {
      delete config.external;
    }
```

but when building, it crashes:

```
(babel plugin) SyntaxError: /Users/bytedance/workspace/ee_web_apps-menu/node_modules/lodash-es/isBuffer.js: 'import' and 'export' may only appear at the top level (4:0)

  2 |
  3 | var isBuffer_1 = commonjsHelpers.createCommonjsModule(function (module, exports) {
> 4 | import root from './_root.js';
    | ^
  5 | import stubFalse from './stubFalse.js';
  6 |
  7 | /** Detect free variable `exports`. */

at undefined:4:0
```

### Suggested solution(s)

when building umd module, it should not rename lodash to lodash-es

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| TSDX           | ^0.13.2
| TypeScript       |
| Browser          |
| npm/Yarn         |
| Node             | 
| Operating System |

agilgur5:
> when building umd module, it should not rename lodash to lodash-es

Ah yep, that's probably required for UMD. Currently the code only checks for CJS. Should be a one-liner fix here: https://github.com/formium/tsdx/blob/8b148cea34c1852f6fcf6e94a3758d7a9f46d9b2/src/babelPluginTsdx.ts#L72

Morphexe:
How do I disable this feature ?

agilgur5:
@Morphexe not sure what you mean, this is a bug, not a feature.

I think one can do a partial workaround for this by configuring `babelrc` with a dummy entry:

```js
  // ...
  plugins: [
  // ...
  [ 'babel-plugin-transform-rename-import', replacements: [{ original: '', replacement: '' }]
```

But conversely, that'll probably break ESM builds

agilgur5:
So https://github.com/formium/tsdx/issues/912#issuecomment-717002753 revealed a similar issue that points to upstream https://github.com/rollup/plugins/issues/304 which would seem to confirm my suspicion that there's a bug upstream.

Per my comment there: 
[...]

Issue #254, opened 2019-10-15
Release automation


jaredpalmer:
Continuing on this https://github.com/jaredpalmer/tsdx/issues/85

It would be great to automate releases so that @sw-yx can cut releases as well.

jaredpalmer:
I think intuit has a cool thing with labels now called like `auto`. Worth a try?

jaredpalmer:
https://github.com/intuit/auto

jaredpalmer:
and perhaps this: https://github.com/auto-it/autobot

sw-yx:
don't we already use circleci? i think kentcdodds uses circle to do his release automation. i've never actually added this to any of my npm projects tho, would be cool to add this

puregarlic:
I'd love to help out with this if possible. I've previously used [`semantic-release`](https://www.npmjs.com/package/semantic-release) to do automated package releases, and it would be pretty easy to configure a GitHub Actions job based on either that or one of the other tools mentioned by @jaredpalmer. I'd say `auto` is the best option, since it's probably not reasonable to hold everyone to conventional commits.

If we were to use `auto` and a continuous release process, I imagine that it would look something like this:

1. A contributor opens a pull request to `tsdx` master
2. The [Labeler](https://github.com/actions/labeler) GitHub Action runs, matching against different file paths to determine the PR's changelog section. We won't be able to determine whether it's `major`, `minor`, or `patch` from this check, since that requires maintainer discretion, but we'll at least be able to sort out most `internal` or `documentation` changes.
3. A required "Check Release Status" GitHub Action runs, failing if there's no `auto`-sanctioned label attached to the PR. See [here](https://intuit.github.io/auto/pages/autorc.html#changelog-titles) for that list.
4. Repeat the usual review -> request changes -> re-review workflow
5. A maintainer applies a version label to the pull request once it's ready to merge
6. The "Check Release Status" Action runs again, this time passing because `auto`'s requirements for a release are now met
7. A maintainer merges the branch
8. A "Release" Action is initiated that runs `auto shipit` on the master branch

That's my initial vision for the process, at least. Is that kind of what you guys are looking for?

agilgur5:
[`dripip`](https://github.com/prisma-labs/dripip) from Prisma has a really interesting flow where basically all PRs can be pre-released without needing a full publish/release process (deciding semver, release notes, batching multiple changes into one release, etc), allowing people to unblock themselves when necessary.

[...]

PR #884, opened 2020-09-21
fix: don't replace lodash/fp imports with lodash-es/fp

altrim:
When importing modules from `lodash/fp` we get breaking imports since `babel-plugin-transform-rename-import` renames all `lodash` imports to `lodash-es` for the esm build.

e.g `import mergeAll from 'lodash/fp/mergeAll';` ends up being imported like `import mergeAll from 'lodash-es/fp/mergeAll'` and we get an error since `lodash-es/fp` doesn't exist.

With this fix we rename all `lodash-es/fp` imports to `lodash/fp` and we get the correct imports in the final esm build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
   "devDependencies": {
     "@types/eslint": "^6.1.2",
     "@types/fs-extra": "^9.0.1",
+    "@types/lodash": "^4.14.161",
     "@types/node": "^14.11.1",
     "@types/react": "^16.9.11",
     "@types/rollup-plugin-json": "^3.0.2",
--- a/src/babelPluginTsdx.ts
+++ b/src/babelPluginTsdx.ts
@@ -10,7 +10,8 @@ export const isTruthy = (obj?: any) => {
   return obj.constructor !== Object || Object.keys(obj).length > 0;
 };
 
-const replacements = [{ original: 'lodash', replacement: 'lodash-es' }];
+// replace lodash with lodash-es, but not lodash/fp
+const replacements = [{ original: 'lodash(?!/fp)', replacement: 'lodash-es' }];
 
 export const mergeConfigItems = (type: any, ...configItemsToMerge: any[]) => {
   const mergedItems: any[] = [];
--- a/test/e2e/fixtures/build-default/src/index.ts
+++ b/test/e2e/fixtures/build-default/src/index.ts
@@ -6,6 +6,9 @@ import './syntax/jsx-import/JSX-import-JSX';
 import './syntax/async';
 export { testGenerator } from './syntax/generator';
 
+export { kebabCase } from 'lodash';
+export { merge, mergeAll } from 'lodash/fp';
+
 export { foo } from './foo';
 
 export const sum = (a: number, b: number) => {
--- a/test/e2e/tsdx-build-default.test.ts
+++ b/test/e2e/tsdx-build-default.test.ts
@@ -67,6 +67,14 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(matched).toBeFalsy();
   });
 
+  it("shouldn't replace lodash/fp", () => {
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    const matched = grep(/lodash\/fp/, ['dist/build-default.*.js']);
+    expect(matched).toBeTruthy();
+  });
+
   it('should clean the dist directory before rebuilding', () => {
     let output = execWithCache('node ../dist/index.js build');
     expect(output.code).toBe(0);
--- a/test/e2e/tsdx-build-default.test.ts
+++ b/test/e2e/tsdx-build-default.test.ts
@@ -67,6 +67,22 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(matched).toBeFalsy();
   });
 
+  it('should use lodash for the CJS build', () => {
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    const matched = grep(/lodash/, ['dist/build-default.cjs.*.js']);
+    expect(matched).toBeTruthy();
+  });
+
+  it('should use lodash-es for the ESM build', () => {
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    const matched = grep(/lodash-es/, ['dist/build-default.esm.js']);
+    expect(matched).toBeTruthy();
+  });
+
   it("shouldn't replace lodash/fp", () => {
     const output = execWithCache('node ../dist/index.js build');
     expect(output.code).toBe(0);
--- a/yarn.lock
+++ b/yarn.lock
@@ -1314,6 +1314,11 @@
   dependencies:
     "@types/node" "*"
 
+"@types/lodash@^4.14.161":
+  version "4.14.161"
+  resolved "https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.161.tgz#a21ca0777dabc6e4f44f3d07f37b765f54188b18"
+  integrity sha512-EP6O3Jkr7bXvZZSZYlsgt5DIjiGr0dXP1/jVEwVLTFgg0d+3lWVQkRavYVQszV7dYUwvg0B8R0MBDpcmXg7XIA==
+
 "@types/minimatch@*", "@types/minimatch@^3.0.3":
   version "3.0.3"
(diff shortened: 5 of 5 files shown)


src/babelPluginTsdx.ts
```diff
@@ -10,7 +10,10 @@ export const isTruthy = (obj?: any) => {
   return obj.constructor !== Object || Object.keys(obj).length > 0;
 };
 
-const replacements = [{ original: 'lodash', replacement: 'lodash-es' }];
+const replacements = [
+  { original: 'lodash', replacement: 'lodash-es' },
+  { original: 'lodash-es/fp', replacement: 'lodash/fp' },
```
agilgur5, line 15: does this actually work? seems very hacky and would rely on the plugin using a serial two-pass replace. In any case, we'd want to add a test for this

src/babelPluginTsdx.ts
```diff
@@ -10,7 +10,10 @@ export const isTruthy = (obj?: any) => {
...
+];
```
agilgur5, line 16: `babel-plugin-transform-rename-import` [supports RegEx](https://github.com/laat/babel-plugin-transform-rename-import#babelrc) -- I think using [a negative lookahead](https://stackoverflow.com/a/9594067/3431180) would be a cleaner fix for this

```suggestion
// replace lodash with lodash-es, but not lodash/fp
const replacements = [{ original: 'lodash(?!\/fp)', replacement: 'lodash-es' }];
```

agilgur5 (changes requested):
Thanks for the bug report and fix @altrim ! I've proposed a cleaner fix below using RegEx.

While I've tested the RegEx I've provided [against the source](https://github.com/laat/babel-plugin-transform-rename-import/blob/7d1365f5377f926ab0fd03a30da92ed8335cbaa6/src/index.js#L7-L10), it would also be good to add an automated test for this behavior to ensure it always works properly. A test could be added to `test/e2e/tsdx-build-default.test.ts` with fixture in `test/e2e/fixtures/build-default/src/index.ts`

I'd also wonder if we should add a check for any other potential uses of `lodash` packages, but afaict, `lodash/fp` is the only other official one.

src/babelPluginTsdx.ts
altrim: It works, but I agree your solution seems better. 👍

altrim:
> Thanks for the bug report and fix @altrim ! I've proposed a cleaner fix below using RegEx.
> 
> While I've tested the RegEx I've provided [against the source](https://github.com/laat/babel-plugin-transform-rename-import/blob/7d1365f5377f926ab0fd03a30da92ed8335cbaa6/src/index.js#L7-L10), it would also be good to add an automated test for this behavior to ensure it always works properly. A test could be added to `test/e2e/tsdx-build-default.test.ts` with fixture in `test/e2e/fixtures/build-default/src/index.ts`
> 
> I'd also wonder if we should add a check for any other potential uses of `lodash` packages, but afaict, `lodash/fp` is the only other official one.

I've updated the PR to include the required changes. Anything else left or is this ok?

package.json
```diff
@@ -53,6 +53,7 @@
     "@rollup/plugin-node-resolve": "^9.0.0",
     "@rollup/plugin-replace": "^2.2.1",
     "@types/jest": "^25.2.1",
+    "@types/lodash": "^4.14.161",
```
agilgur5, line 56: This should be a devDep -- it's only used during testing and not part of actual builds

test/e2e/fixtures/build-default/src/index.ts
```diff
@@ -4,6 +4,7 @@ import './syntax/optional-chaining';
 import './syntax/jsx-import/JSX-import-JSX';
 
 import './syntax/async';
+export { merge, mergeAll } from 'lodash/fp';
 export { testGenerator } from './syntax/generator';
```
agilgur5, line 8:
```suggestion
export { testGenerator } from './syntax/generator';

export { kebabCase } from 'lodash';
export { merge, mergeAll } from 'lodash/fp';
```

The syntax imports are all grouped together intentionally. Please move the `lodash` imports below them. I'm also refactoring some of the syntax tests so this will ensure less merge conflicts

I added plain `lodash` here because we also want to add a regression test to ensure that hasn't broken with this change, since unfortunately that test didn't exist beforehand 😕

test/e2e/tsdx-build-default.test.ts
```diff
@@ -105,6 +105,15 @@ describe('tsdx build :: zero-config defaults', () => {
...
 
+  it('should replace lodash/fp correctly', () => {
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    // ensures correct lodash/fp import was used
+    const matched = grep(/lodash\/fp/, ['dist/build-default.*.js']);
+    expect(matched).toBeTruthy();
+  });
+
```
agilgur5, line 116: Could you move this test above the `clean` test, but below the `regeneratorRuntime` test? Basically 1 up. The clean test is a cache buster, so I'd prefer for that to stay at the end.

test/e2e/tsdx-build-default.test.ts
```diff
@@ -105,6 +105,15 @@ describe('tsdx build :: zero-config defaults', () => {
     shell.mv('package-og.json', 'package.json');
   });
 
+  it('should replace lodash/fp correctly', () => {
```
agilgur5, line 108:
```suggestion
  it("shouldn't replace lodash/fp", () => {
```

"should replace [...] correctly" makes it sound like it's supposed to perform a replace. But it's _not_ supposed to perform a replace and the check is to make sure it's still intact

test/e2e/tsdx-build-default.test.ts
```diff
@@ -105,6 +105,15 @@ describe('tsdx build :: zero-config defaults', () => {
...
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+
+    // ensures correct lodash/fp import was used
```
agilgur5, line 112:
```suggestion
```

the other tests don't have comments in here and I don't think this comment is necessary; the `it` block describes the check already

test/e2e/tsdx-build-default.test.ts
```diff
@@ -105,6 +105,15 @@ describe('tsdx build :: zero-config defaults', () => {
...
+    const matched = grep(/lodash\/fp/, ['dist/build-default.*.js']);
```
agilgur5, line 113: Could you add another test or two (above this test) for plain `lodash` and `lodash-es` as well? Would be the same format as this, but you'd want to test the specific format (`*.cjs.js`, `.esm.js`) since that's part of the replacement logic.

Like I mentioned above, this would ensure the plain `lodash` import still works since unfortunately a test for that didn't exist beforehand

agilgur5 (changes requested):
Thanks for the iteration @altrim !

There's some small changes I'd like to see in the test, but would also be great if you could add tests for the original plain `lodash` replacement to ensure that still works. Could write that as a separate PR though if you want since it's existing behavior.

altrim:
Hey @agilgur5 anything else left here or is it ok to merge?

agilgur5 (approved):
Great work on this @altrim ! Thanks for all the iterations

> Hey @agilgur5 anything else left here or is it ok to merge?

If you're blocked by this, I'd recommend using `patch-package` temporarily as until #254 is complete, I'm going to be batching releases so folks don't get spammed (even with #254 though, stable releases will still be batched, but canaries could go out faster).

agilgur5 (approved):
I rebased to current `master` and made some slight modifications to test layout and naming, but did not change any real content otherwise. Also squashed up some commits and added more details in the commit messages.

Apparently GitHub automatically marked my review as stale so re-adding it here. Will merge this as soon as tests pass!

agilgur5:
@allcontributors please add @altrim for bug, code, test

agilgur5:
Ack, while writing a test for #896 I got this in the logs during a UMD build:

```txt
No name was provided for external module 'lodash-es' in output.globals – guessing 'lodashEs'
No name was provided for external module 'lodash/fp' in output.globals – guessing 'fp'
No name was provided for external module 'lodash-es' in output.globals – guessing 'lodashEs'
```

`lodash-es` I assume would use the same global `lodash` (or maybe UMD shouldn't use the `-es` variant #759 ), not sure what `lodash/fp` should use... jotting this down, haven't quite looked into it yet.
It's an existing "bug" though (it's not an error, just a warning), this just added tests that show these warnings.

Issue #635, opened 2020-03-23
[RFC] Shareable Plugins (for TSDX Build/Config, i.e. `tsdx.config.js` or Rollup in general)


agilgur5:
### Current Behavior

Currently, `tsdx.config.js` allows many internals to be changed however a user so chooses. As the warning I added to the README states, this can be very fragile/brittle against updates and can break easily, since this exposes a large part of TSDX as an API that's normally only used internally.

That warning not withstanding, there are still many users of `tsdx.config.js`, and both official and unofficial documentation (e.g. #379 ) of ways of extending it.

Features like #367 , at least one optimization mentioned there, and likely many more would be entirely non-breaking, but because they change the Rollup config, they are breaking for certain users of `tsdx.config.js` (which is one of the primary reasons I added the warning -- it's very easy to break this API as the internal Rollup config is a core portion of `tsdx build` which is the core of TSDX).

### Desired Behavior

Document some sort of API that allows users to share `tsdx.config.js` plugins together for interoperability and better maintenance -- especially through version-locked peerDependencies on TSDX itself (that current copy/paste solutions don't have).

It also allows for users and orgs to re-use their configs between projects more easily.

Plugins could also be written in TypeScript for better safety, adhering to an interface set by TSDX, using whatever other library types are available, and transpiled down to JS for use by `tsdx.config.js` (which can only be written in CJS right now, as it's `require`'d).

### Suggested Solution

The kicker is that I'm 100% sure what this should look like yet, there are pros/cons to each option. The main proposal is:

1. The plugin just accepts a Rollup `config` and returns a new `config`. That would look like:
    
    ```js
    const tsdxRollupPluginPostCSS = require('tsdx-rollup-plugin-postcss');
  
    // tsdx.config.js
    module.exports = {
      rollup (config, options) {
         config = tsdxRollupPluginPostCSS(config);
         return config;
      }
    }
    ```
    
    This has some limited extensibility to general Rollup users, but not that much because any overrides would be specific to some plugins. Extensibility to general Rollup users is preferred.
  
    The major problem with it is that it doesn't account for if we add new configurations to `tsdx.config.js`. Top of mind is being able to more easily re-configure existing plugins similarly to how `customize-cra` works, so that one doesn't have to do hacky `config.plugins.map((plugin) => plugin.name !== myPlugin.name ? plugin : ...` and so that they don't need to match the internal config and can just override it where necessary

2. The plugin operates on the entire `tsdx.config.js`, looking like:

[...]

PR #889, opened 2020-09-28
Update to Rollup 2.

ludovicofischer:
- Update Rollup
 - Update plugin-commonjs, plugin-json, plugin-replace, plugin-terser to versions
   officially supporting Rollup 2


ludovicofischer:
Looks like the MacOS failure is anetwork problem on the test machine. I believed this would be harder, am I missing something? the only change to the code itself is that the terser plugin does not accept a `sourcemap` option anymore, instead it infers it from the rollup options (see https://github.com/TrySound/rollup-plugin-terser/releases/tag/v6.0.0)

agilgur5:
> I believed this would be harder, am I missing something?

Rollup has not been upgraded for lack of effort, it's because it is a very breaking change, likely affecting a good portion of `tsdx.config.js` users. See https://github.com/formium/tsdx/issues/821#issuecomment-678745312.
As mentioned there, previously it was also not well supported, and I believe there are still issues with TS 3.8+ as I mentioned in that comment and where it links to: https://github.com/formium/tsdx/pull/679#issuecomment-631478298
It also drops Node 8 support, as does `plugin-terser`. TSDX v0.14.0 was just released [a week ago](https://github.com/formium/tsdx/releases/tag/v0.14.0) that now requires Node 10+ -- Node 8 was still supported until a week ago.

I also wrote in the release notes of v0.14.0:
> I also wanted to push Rollup 2, TS 4.0, Prettier 2, and ESLint upgrades into this, but the breaking changes in the changelog started getting too big, so I decided to wait a bit to split those changes across more releases to not throw too much breakage at users at once.

I'm not planning on releasing Rollup 2 at least for a few weeks to spread out aforementioned breakage, and I was planning to batch that in with [v0.15.0's changes](https://github.com/formium/tsdx/milestone/3). But as TS 4, Prettier 2, and ESLint were left out of v0.14.0 intentionally, I was thinking of shipping those as v0.15.0 and pushing what's currently slated as that until v0.16.0. Those could be interchanged though.

ludovicofischer:
Yes, as I was making the changes I realized maintaining a compiler tool like this must be horribly complicated as it’s hard to check if and how you’re breaking other projects. It’s even harder as you’re not really in control of what upstream dependencies decide to release.

agilgur5:
@ludovicofischer indeed. I appreciate the acknowledgement and understanding.

Breaking changes really require a lot of thought and planning of maintainers and lot of factors in TSDX don't make that easier. That's one of the reasons why the Unix methodology of small modules doing one thing well is a good one to follow -- the smaller the API surface, the less potential for breakage there is, and the more known that breakage looks like. The scoping problem becomes larger with the introduction of any new API, which is why those decisions should really be made quite carefully.
Some maintainers can widen out APIs too quickly and also break APIs unknowingly frequently, one makes the other an easier mistake to make.

Being a compiler tool as you say increases the scope and potential impact of breakage to be quite large; it can literally break the correctness of your built code (see also the problems with [`async-to-promises`](https://github.com/formium/tsdx/labels/topic%2Fasync-to-promises) and the recent switch to regenerator in #795).
`tsdx.config.js` opens up basically the whole Rollup API as a TSDX API that can be broken. That's why I discourage it and say to only use if necessary as an escape hatch, per the warning I added in #400. It's very difficult to promise that we'll know if anything breaks that, and nearly any plugin change or minor change could be considered breaking, so that warning is also a warning that unforeseen breakage is to be expected with `tsdx.config.js`. This breakage is quite foreseen though, so I think good faith should be made to plan it as a breaking change and notify users.
Being an all-in-one tool further increases this scope. A breaking change to any one of `tsdx build`, `tsdx test`, `tsdx lint`, `tsdx create`, etc also mean that I have to release a breaking change for the entirety of TSDX since they're all currently combined as one big API of TSDX.

#635 is one of my approaches to improve the `tsdx.config.js` problem (make maintainable and testable plugins) and #634 is an approach to start splitting out some of the API. I'd like to split out each command into a separate package as well, that way I can work on improving, say, `@tsdx/test` independently and release a breaking change there without affecting `tsdx test` until that is upgraded. Can then batch breaking changes together more easily and also allow users to opt-out or opt-in to different pieces of TSDX without requiring the whole thing.

package.json
```diff
@@ -122,7 +122,7 @@
     "styled-components": "^5.0.1",
     "tiny-invariant": "^1.1.0",
     "tiny-warning": "^1.0.3",
-    "yarn-deduplicate": "^2.1.1"
+    "yarn-deduplicate": "^3.0.0"
```
agilgur5, line 125: This isn't mentioned in the PR/commit description and doesn't seem to be related. 

The breaking changes don't seem to [affect our usage](https://github.com/atlassian/yarn-deduplicate/blob/master/CHANGELOG.md#300---2020-10-29) and it doesn't seem to have any new features or bugfixes, so I don't think there's much need to upgrade this.

If you still want to make this change, I would say please make a separate PR for that.

package.json
```diff
@@ -122,7 +122,7 @@
     "styled-components": "^5.0.1",
```
agilgur5, line 122: Can't comment on unchanged lines in GitHub's code review, but the line above this, `rollup-plugin-postcss`, should also be updated to a version beyond [v3](https://github.com/egoist/rollup-plugin-postcss/releases/tag/v3.0.0) as that supports Rollup 2 (and not Rollup 1).

That would also resolve the final PR needed for Rollup 2, #693 (also mentioned in https://github.com/formium/tsdx/issues/821#issue-684122764).

package.json
```diff
@@ -48,10 +48,10 @@
     "@babel/preset-env": "^7.11.0",
     "@babel/traverse": "^7.11.5",
     "@rollup/plugin-babel": "^5.1.0",
-    "@rollup/plugin-commonjs": "^11.0.0",
-    "@rollup/plugin-json": "^4.0.0",
+    "@rollup/plugin-commonjs": "^15.1.0",
```
agilgur5, line 51: See https://github.com/formium/tsdx/pull/727#issuecomment-695894307 : upgrading `commonjs` apparently required Rollup 2 without any note in the Changelog

package.json
```diff
@@ -48,10 +48,10 @@
...
+    "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
-    "@rollup/plugin-replace": "^2.2.1",
+    "@rollup/plugin-replace": "^2.3.3",
```
agilgur5, line 54: [v2.3.2](https://github.com/rollup/plugins/blob/master/packages/replace/CHANGELOG.md#v232) is the first to add it in the peerDep range, hence this change. Older versions are forward-compatible but will produce a peerDep warning

package.json
```diff
@@ -48,10 +48,10 @@
     "@babel/preset-env": "^7.11.0",
     "@babel/traverse": "^7.11.5",
     "@rollup/plugin-babel": "^5.1.0",
-    "@rollup/plugin-commonjs": "^11.0.0",
-    "@rollup/plugin-json": "^4.0.0",
+    "@rollup/plugin-commonjs": "^15.1.0",
+    "@rollup/plugin-json": "^4.1.0",
```
agilgur5, line 52: [v4.0.3](https://github.com/rollup/plugins/blob/master/packages/json/CHANGELOG.md#v403) is the first to add it in the peerDep range, hence this change. Older versions are forward-compatible but will produce a peerDep warning

package.json
```diff
@@ -86,9 +86,9 @@
     "prettier": "^1.19.1",
     "progress-estimator": "^0.2.2",
     "regenerator-runtime": "^0.13.7",
-    "rollup": "^1.32.1",
+    "rollup": "^2.28.2",
     "rollup-plugin-sourcemaps": "^0.6.2",
-    "rollup-plugin-terser": "^5.1.2",
+    "rollup-plugin-terser": "^7.0.2",
```
agilgur5, line 91: [v6.0.0](https://github.com/TrySound/rollup-plugin-terser/releases/tag/v6.0.0) requires Rollup 2 and Node 10+. c.f. https://github.com/formium/tsdx/pull/731#issuecomment-672463413

src/createRollupConfig.ts
```diff
@@ -202,7 +202,6 @@ export async function createRollupConfig(
       sourceMaps(),
       shouldMinify &&
         terser({
-          sourcemap: true,
```
agilgur5, line 205: As of [v6.0.0](https://github.com/TrySound/rollup-plugin-terser/releases/tag/v6.0.0), inferred automatically from Rollup's `output.sourcemap`

yarn.lock
```diff
@@ -4946,42 +5118,43 @@ jest-watch-typeahead@^0.5.0:
...
+  version "26.3.0"
+  resolved "https://registry.yarnpkg.com/jest-worker/-/jest-worker-26.3.0.tgz#7c8a97e4f4364b4f05ed8bca8ca0c24de091871f"
+  integrity sha512-Vmpn2F6IASefL+DVBhPzI2J9/GJUsqzomdeN+P+dK8/jKxbh8R3BtFnx3FIta7wYlPU62cpJMJQo4kuOowcMnw==
   dependencies:
+    "@types/node" "*"
     merge-stream "^2.0.0"
     supports-color "^7.0.0"
 
 jest@^25.3.0:
-  version "25.3.0"
```
agilgur5, line 4978: I'm not sure why this was changed as it seems unrelated and I couldn't find a dep that was upgraded that has `jest` itself as direct dependency nor one that was updated.

This and a few other changes, like to `@babel/core` make me suspect you ran a `yarn upgrade` which I do not recommend as it just upgrades everything -- including minors and patches unnecessarily -- instead of the exact dependencies that were changed or need upgrading.

If so, I would throw out these changes to `yarn.lock` and just run `yarn` + `yarn deduplicate` against the new `package.json`, which should condense to only the changes that are necessary.

agilgur5 (changes requested):
Thanks for the PR @ludovicofischer !

I've made several comments here. Highlights are:
- Per https://github.com/formium/tsdx/pull/789#pullrequestreview-463885368, writing a rationale and linking to the changelog for each change and describing them in the PR and commit makes it much, _much_, easier to review. I've added comments here that describe, from my perspective, why it seems like each of these changes were made. 
- The devDep, `rollup-plugin-postcss` should be upgraded as well. There's an integration test for it
- `yarn.lock` changes seem to be too broad

package.json
ludovicofischer: Sorry, I updated this trying to fix some errors but it turned out I did not understand how yarn-deduplicate worked. I will update the pull request tomorrow as it is time to go to bed where I live.

yarn.lock
ludovicofischer: If I understand correctly, I should revert `yarn.lock` to how it looks on the current master, then run the `yarn` commands? I don’t think I ran `yarn upgrade` but I am more familiar with `npm` than `yarn` so I could have made some mistake.

package.json
agilgur5: Ok, no rush since I'm gonna have to hold on merging this anyway as it's breaking per our conversation above.

yarn.lock
agilgur5: > I should revert `yarn.lock` to how it looks on the current master, then run the `yarn` commands?

Yep, exactly.

> I don’t think I ran `yarn upgrade` but I am more familiar with `npm` than `yarn` so I could have made some mistake.

Yea same. `npm upgrade rollup@"^2.28.2"` I believe would only update Rollup and anything affected, whereas `yarn upgrade rollup@"^2.28.2"` basically ignores the argument, at least in my experience. `yarn add rollup@"^2.28.2"` I _think_ only upgrades it, but there's some quirks I've found with that as well. `yarn-deduplicate` itself exists because of a bug/bugs with Yarn that it doesn't dedupe properly (despite `yarn dedupe` saying otherwise). 

Upgrading `package.json` by hand and then running `yarn` + `yarn deduplicate` has been the safest and most predictable method I've found of updating the lockfile 😅

ludovicofischer: `rollup-plugin-terser` depends on `jest-worker`

package.json
agilgur5: Looks like this was removed in the latest update to the PR

package.json
agilgur5: Looks like this was added in the most recent update to the PR

yarn.lock
agilgur5: Yes, but `jest-worker` doesn't depend on `jest` itself (which can be seen above this line), so that shouldn't have changed.

It looks like you followed the directions in your latest update to the PR and now `yarn.lock` looks correctly changed. It's about 10% the changes of the previous version.

agilgur5 (approved):
Thanks for the iteration @ludovicofischer ! Seems like all the issues I pointed out are now resolved. 
Tests all pass and no stray warnings or errors in the logs.

I've also deleted ~56 unnecessary Greenkeeper branches that were associated with the related PRs that this consolidates and updates.

This is ready to go, but per the version label, please to any other folks with write access **DO NOT MERGE** until the v0.15.0 release is being prepared, as, to repeat, this is quite **BREAKING**.

ludovicofischer:
Great! I’ve been happy to help.

PR #890, opened 2020-09-28
Use compatible eslint-config-react-app and eslint-plugin-react-hooks.

ludofischer:
This solves the following warning on yarn install

eslint-config-react-app@5.0.2" has incorrect peer dependency "eslint-plugin-react-hooks@1.x".

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "enquirer": "^2.3.4",
     "eslint": "^6.1.0",
     "eslint-config-prettier": "^6.0.0",
-    "eslint-config-react-app": "^5.0.2",
+    "eslint-config-react-app": "^5.2.1",
     "eslint-plugin-flowtype": "^3.13.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-jsx-a11y": "^6.2.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3121,10 +3121,10 @@ eslint-config-prettier@^6.0.0:
   dependencies:
     get-stdin "^6.0.0"
 
-eslint-config-react-app@^5.0.2:
-  version "5.0.2"
-  resolved "https://registry.yarnpkg.com/eslint-config-react-app/-/eslint-config-react-app-5.0.2.tgz#df40d73a1402986030680c040bbee520db5a32a4"
-  integrity sha512-VhlESAQM83uULJ9jsvcKxx2Ab0yrmjUt8kDz5DyhTQufqWE0ssAnejlWri5LXv25xoXfdqOyeDPdfJS9dXKagQ==
+eslint-config-react-app@^5.2.1:
+  version "5.2.1"
+  resolved "https://registry.yarnpkg.com/eslint-config-react-app/-/eslint-config-react-app-5.2.1.tgz#698bf7aeee27f0cea0139eaef261c7bf7dd623df"
+  integrity sha512-pGIZ8t0mFLcV+6ZirRgYK6RVqUIKRIi9MmgzUEmrIknsn3AdO0I32asO86dJgloHq+9ZPl8UIg8mYrvgP5u2wQ==
   dependencies:
     confusing-browser-globals "^1.0.9"
 


ludovicofischer:
I guess this is a breaking change for people who relied on the previous React hooks lint rules.

agilgur5:
Yeaaa, it's probably breaking, but I'm not really sure why `eslint-config-react-app` didn't release a breaking change for it, it seems to change the rules in patches and minors [per the changelogs](https://github.com/facebook/create-react-app/releases) 😕  There's also no individual changelog for this package [in its directory](https://github.com/facebook/create-react-app/blob/025f2739ceb459c79a281ddc6e60d7fd7322ca24/packages/eslint-config-react-app/).

Per the milestone added here and my comments in #889 , I was planning to release all the ESLint and Prettier upgrades as one breaking change together.

agilgur5 (changes requested):
So the CI logs have a different warning now:

```bash
warning " > eslint-config-react-app@5.2.1" has incorrect peer dependency "eslint-plugin-react-hooks@1.x || 2.x".
```

🤔 

But oddly enough they don't seem to show warnings for all the other peer deps that need upgrading if `config-react-app` is upgraded. Several more plugins would need to bumped [per its current peerDeps](https://github.com/facebook/create-react-app/blob/025f2739ceb459c79a281ddc6e60d7fd7322ca24/packages/eslint-config-react-app/package.json#L19)

agilgur5:
Ah, here's what the [v5.2.1 peerDeps look like](https://github.com/facebook/create-react-app/blob/7641a3c1c6d02cd707ac11e7b8b8907acf3df7c0/packages/eslint-config-react-app/package.json#L17), which explains both the current warning and the lack of warnings. All the other changes (including addition of `eslint-plugin-jest`) seem to have been made after v5.2.1 and I guess not yet released?

ludovicofischer:
In light of the discussion in #634 I wonder if it would not be better in the future to inline  eslint rules into tsdx own config. That way it would not be forced to wrangle with dependency combinations which in this case are somewhat arbitrary (as package versions reflect rule changes more often than compatibility with eslint versions). Also you would have more control over which new warnings to introduce.

agilgur5:
I'm not sure that would really solve anything and may make TSDX actually have to make _more_ breaking changes, because most rule changes (at least those to `error`s) are breaking. I also personally don't want to have to wrangle with opinionated rules, the feedback therein, etc -- folks who make configs already deal with that and I'd prefer to use their work instead of adding more work for myself (which is already too much).

package.json
```diff
@@ -68,13 +68,13 @@
...
     "eslint-config-prettier": "^6.0.0",
-    "eslint-config-react-app": "^5.0.2",
+    "eslint-config-react-app": "^5.2.1",
     "eslint-plugin-flowtype": "^3.13.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-react": "^7.14.3",
-    "eslint-plugin-react-hooks": "^2.2.0",
+    "eslint-plugin-react-hooks": "^2.5.0",
```
agilgur5, line 77: Did this actually need changing? Because I don't believe it does -- `^2.2.0` is compatible with `2.5.0`

On the same note, `eslint-config-react-app` shouldn't need changing either but it does seem to install `5.0.2` by default for some reason 🤔

agilgur5 (changes requested):

package.json
ludofischer: I updated the pull request as you suggested, but think we were both getting confused. We need to distinguish what happens when a `tsdx` developer runs `yarn install` in the `tsdx` repository from the end user case.
The warning appears when _developing_ tsdx itself, because the `yarn.lock` in the repository pins `eslint-config-react-app` to `5.0.2`. 
If a _user_ installs `tsdx` for the first time, they won’t get any error, as it will resolve the `eslint-config-react-app` to the latest compatible version, right now `5.2.1`.
The user _will_ see an error message if they run ` yarn add tsdx` when a new version comes out and they first installed `tsdx` when `eslint-config-react-app` was still at `5.0.2`, because in that case their own  `yarn.lock` would pin `eslint-config-react-app` (I tested this by setting the `eslint-config-react-app` dependency to exactly `5.0.2`, installing, then setting it to `^5.0.2`, installing again, then finally setting it to `^5.2.1` and installing a third time).
In the end, the minimal solution is still to do just as you suggest, and update only `eslint-config-react-app` to the latest version both in `package.json` and `yarn.lock`.
1. tsdx `yarn.lock` update (`yarn upgrade eslint-config-react-app` in tsdx repository) takes care of the warning while developing tsdx itself
2. For an upgrading user, `tsdx`’s own `yarn.lock` has no effect, so getting rid of the warning requires to also update `eslint-config-react-app` in `package.json`, otherwise the installer will keep using the version set in their own lockfile
3. Updating `eslint-plugin-react-hooks` is unnecessary because `2.2.0` is already compatible with the latest `eslint-config-react-app`

agilgur5: > The warning appears when _developing_ tsdx itself, because the `yarn.lock` in the repository pins `eslint-config-react-app` to `5.0.2`.

Yea, was aware of that one, but that's not the one I meant. This one's not particularly important as it's not visible to end users.

> The user _will_ see an error message if they run ` yarn add tsdx` when a new version comes out and they first installed `tsdx` when `eslint-config-react-app` was still at `5.0.2`, because in that case their own `yarn.lock` would pin `eslint-config-react-app` (I tested this by setting the `eslint-config-react-app` dependency to exactly `5.0.2`, installing, then setting it to `^5.0.2`, installing again, then finally setting it to `^5.2.1` and installing a third time).

Ah that would make sense that their lockfile has the transitive dep set. What I was referencing actually was that I was getting this error when I was installing _globally_. I guess there's a global lockfile somewhere too that makes this happen 

> 2\. so getting rid of the warning requires to also update `eslint-config-react-app` in `package.json`, otherwise the installer will keep using the version set in their own lockfile

👍  . I thought this might be relevant to vulnerabilities too like https://github.com/formium/tsdx/pull/731#issuecomment-688246341, but an `npm audit fix` will automatically resolve those lockfile updates, whereas a warning like this one needs manual intervention

Great analysis @ludofischer !

agilgur5 (approved):
Minimally required change LGTM now. Double-checked and confirmed that there were no other breaking changes and basically no other changes in [between 5.0.2 and 5.2.1](https://github.com/facebook/create-react-app/commits/d2f813f8897ffcd2f0b0d2da75d0c44924c92f4d/packages/eslint-config-react-app/package.json)

Thanks for the contribution!

agilgur5:
@allcontributors please add @ludofischer for code

Issue #888, opened 2020-09-24
v0.14 No longer supports yarn workspaces


jaredpalmer:
### Current Behavior

TSDX isn't building when used with yarn workspaces. Seems like it may be a rollup-plugin-typescript-2 cache issue? 

```shell
(typescript) Error: ENOENT: no such file or directory, rename '/Users/jared/workspace/github/jaredpalmer/formik-actual/packages/formik/node_modules/.cache/rollup-plugin-typescript2/rpt2_48f7284bed271946d808c741a74a6d872af7a18e/code/cache_' -> '/Users/jared/workspace/github/jaredpalmer/formik-actual/packages/formik/node_modules/.cache/rollup-plugin-typescript2/rpt2_48f7284bed271946d808c741a74a6d872af7a18e/code/cache'
Error: ENOENT: no such file or directory, rename '/Users/jared/workspace/github/jaredpalmer/formik-actual/packages/formik/node_modules/.cache/rollup-plugin-typescript2/rpt2_48f7284bed271946d808c741a74a6d872af7a18e/code/cache_' -> '/Users/jared/workspace/github/jaredpalmer/formik-actual/packages/formik/node_modules/.cache/rollup-plugin-typescript2/rpt2_48f7284bed271946d808c741a74a6d872af7a18e/code/cache'
    at Object.renameSync (fs.js:643:3)
    at RollingCache.roll (/Users/jared/workspace/github/jaredpalmer/formik-actual/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:24003:12)
    at TsCache.done (/Users/jared/workspace/github/jaredpalmer/formik-actual/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:24845:24)
    at Object._ongenerate (/Users/jared/workspace/github/jaredpalmer/formik-actual/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29098:21)
    at Object.generateBundle (/Users/jared/workspace/github/jaredpalmer/formik-actual/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29076:18)
    at /Users/jared/workspace/github/jaredpalmer/formik-actual/node_modules/rollup/dist/shared/node-entry.js:13117:25
[...]

Issue #892, opened 2020-10-01
v0.14.0 rpts2 cache issue when building `umd` with `cjs`


MagnusBrzenk:
### Current Behavior

Right out of the box, running `tsdx build --format umd,cjs,esm` causes following error:

```text
> tsdx build --format esm,cjs,umd

✓ Creating entry file 1.1 secs
(typescript) Error: ENOENT: no such file or directory, rename '/Users/magnus/work/javascript/myapp/node_modules/.cache/rollup-plugin-typescript2/rpt2_4c61ae4392b9bd24f4d43d13a3f56419b8b5805d/code/cache_' -> '/Users/magnus/work/javascript/myapp/node_modules/.cache/rollup-plugin-typescript2/rpt2_4c61ae4392b9bd24f4d43d13a3f56419b8b5805d/code/cache'
Error: ENOENT: no such file or directory, rename '/Users/magnus/work/javascript/myapp/node_modules/.cache/rollup-plugin-typescript2/rpt2_4c61ae4392b9bd24f4d43d13a3f56419b8b5805d/code/cache_' -> '/Users/magnus/work/javascript/myapp/node_modules/.cache/rollup-plugin-typescript2/rpt2_4c61ae4392b9bd24f4d43d13a3f56419b8b5805d/code/cache'
    at Object.renameSync (fs.js:660:3)
    at RollingCache.roll (/Users/magnus/work/javascript/myapp/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:24003:12)
    at TsCache.done (/Users/magnus/work/javascript/myapp/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:24845:24)
    at Object._ongenerate (/Users/magnus/work/javascript/myapp/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29098:21)
    at Object.generateBundle (/Users/magnus/work/javascript/myapp/node_modules/tsdx/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29076:18)
    at /Users/magnus/work/javascript/myapp/node_modules/rollup/dist/shared/node-entry.js:13117:25
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
[...]

PR #896, opened 2020-10-06
fix/deps: upgrade rollup-plugin-typescript2 to fix cache issue

tanem:
Bumped into #888/#892 when creating a new repo, and the workaround mentioned [here](https://github.com/formium/tsdx/issues/892#issuecomment-702449484) didn't work for me.

This patch brings in the permanent fix from ezolenko/rollup-plugin-typescript2#243. Tested by yarn linking tsdx across repos, and the build looks to be working ok with this change.

Guess dependabot would have eventually updated this dep, but figured it was important enough to do ASAP?

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "rollup": "^1.32.1",
     "rollup-plugin-sourcemaps": "^0.6.2",
     "rollup-plugin-terser": "^5.1.2",
-    "rollup-plugin-typescript2": "^0.26.0",
+    "rollup-plugin-typescript2": "^0.27.3",
     "sade": "^1.4.2",
     "semver": "^7.1.1",
     "shelljs": "^0.8.3",
--- a/test/e2e/tsdx-build-options.test.ts
+++ b/test/e2e/tsdx-build-options.test.ts
@@ -1,0 +1,53 @@
+import * as shell from 'shelljs';
+
+import * as util from '../utils/fixture';
+import { execWithCache } from '../utils/shell';
+
+shell.config.silent = false;
+
+const testDir = 'e2e';
+const fixtureName = 'build-default';
+// create a second version of build-default's stage for concurrent testing
+const stageName = 'stage-build-options';
+
+describe('tsdx build :: options', () => {
+  beforeAll(() => {
+    util.teardownStage(stageName);
+    util.setupStageWithFixture(testDir, stageName, fixtureName);
+  });
+
+  it('should compile all formats', () => {
+    const output = execWithCache(
+      'node ../dist/index.js build --format cjs,esm,umd,system'
+    );
+
+    expect(shell.test('-f', 'dist/index.js')).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.cjs.development.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.cjs.production.min.js')
+    ).toBeTruthy();
+    expect(shell.test('-f', 'dist/build-default.esm.js')).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.umd.development.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.umd.production.min.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.system.development.js')
+    ).toBeTruthy();
+    expect(
+      shell.test('-f', 'dist/build-default.system.production.min.js')
+    ).toBeTruthy();
+
+    expect(shell.test('-f', 'dist/index.d.ts')).toBeTruthy();
+
+    expect(output.code).toBe(0);
+  });
+
+  afterAll(() => {
+    util.teardownStage(stageName);
+  });
+});
--- a/yarn.lock
+++ b/yarn.lock
@@ -3559,13 +3559,13 @@ fill-range@^7.0.1:
   dependencies:
     to-regex-range "^5.0.1"
 
-find-cache-dir@^3.2.0:
-  version "3.2.0"
-  resolved "https://registry.yarnpkg.com/find-cache-dir/-/find-cache-dir-3.2.0.tgz#e7fe44c1abc1299f516146e563108fd1006c1874"
-  integrity sha512-1JKclkYYsf1q9WIJKLZa9S9muC+08RIjzAlLrK4QcYLJMS6mk9yombQ9qf+zJ7H9LS800k0s44L4sDq9VYzqyg==
+find-cache-dir@^3.3.1:
+  version "3.3.1"
+  resolved "https://registry.yarnpkg.com/find-cache-dir/-/find-cache-dir-3.3.1.tgz#89b33fad4a4670daa94f855f7fbe31d6d84fe880"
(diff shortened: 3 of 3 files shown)


agilgur5:
Thanks for the patch @tanem -- was actually looking to bump this myself once v0.28.0 is released, because that will contain my fix that was recently merged: https://github.com/ezolenko/rollup-plugin-typescript2/pull/221 (which resolves a few issues and does a root cause fix for a PR of mine in TSDX)

> and the workaround mentioned [here](https://github.com/formium/tsdx/issues/892#issuecomment-702449484) didn't work for me.

If that didn't work for you could you provide more details on that? Neither nuking the cache nor all of `node_modules` worked?
That might suggest https://github.com/ezolenko/rollup-plugin-typescript2/pull/243 isn't a root cause fix then. Though now I think I understand why it's occurring, it's because the cache dir was changed/set to the default in #691 . But if everything is nuked that should be the same as a new install / fresh slate, which should be handled... And #691 was released in v0.13.3... And cache only gets rolled _after_ a build, not before... 🤔 

> Tested by yarn linking tsdx across repos, and the build looks to be working ok with this change.

💯 

> Guess dependabot would have eventually updated this dep

I've temporarily disabled it because all the dep updates were spammy (duplicate PRs, unnecessarily updating patches/minors, updating peerDeps that can't be updated in isolation, etc). Need to add a bunch of filters to it, but even then, it's not feasible, not always necessary, and may actually be a bad idea stability-wise to update deps immediately after release.
Whereas a PR like this or bugs reported actually demonstrate a _need_ to update.

agilgur5 (dismissed):
This looks good to me, though I'd like to understand root cause and why workarounds didn't work per my above comment

tanem:
Hey @agilgur5 👋 

Here's what I did to get the issue:

```
$ npx tsdx create my-lib
[chose react template]
$ cd my-lib
$ yarn build --format cjs,esm,umd
[error]
```

Then I tried:

```
$ rm -rf node_modules/.cache/rollup-plugin-typescript2/
$ yarn build --format cjs,esm,umd
[error]
```

Finally:

```
$ rm -rf node_modules
$ yarn
$ yarn build --format cjs,esm,umd
[error]
```

Once I applied the update in this PR to my local copy of `tsdx`, then linked that over to `my-lib`, `yarn build --format cjs,esm,umd` ran without error.

---

On the topic of dependabot, have you considered using [Renovate](https://github.com/renovatebot/renovate) instead? I find it gives a lot more control in terms of how much noise you have to deal with as a maintainer. It also has a [`stabilityDays`](https://docs.renovatebot.com/configuration-options/#stabilitydays) config option which addresses your point about updating immediately following a release.

One thing it doesn't do that dependabot does do is update nested dependency versions for security updates (hopefully I described that correctly, don't want to misrepresent Renovate 😅). In those cases, I've seen other open-source repos use dependabot just for security patches, and renovate for the rest. For my own repos I usually manually update if I need a security patch applied promptly.

agilgur5:
Thanks for the additional details @tanem 

>```shell
> $ npx tsdx create my-lib
> [chose react template]
> $ cd my-lib
> $ yarn build --format cjs,esm,umd
> [error]
>```

This is pretty intriguing; I thought most folks, like Jared, were getting this during an _upgrade_ of TSDX, but this is happening to you on a _fresh install_.... yea that's even more confusing since rpts2 hasn't been updated since TSDX v0.13.0 / #506 . And #691 / #758 were the only changes to the options and were in v0.13.3... but for some reason this wasn't discovered until v0.14.0 🤔 🤔  
The cache code initiates [both directories in its constructor](https://github.com/ezolenko/rollup-plugin-typescript2/blob/fc2274d2eaa76b0726638cdc6db466e27ac8c498/src/rollingcache.ts#L21) which would by definition be called before its `roll()` call... weird.
#691 made it so it was a single cache instead of a per-format cache (because the config is hashed anyway) so I guess it may have created a race?? 

Will have to follow up on this upstream

agilgur5:
> On the topic of Dependabot, have you considered using [Renovate](https://github.com/renovatebot/renovate) instead?

I have not. I didn't realize Renovate did dependencies, I thought it was an automated merge bot for some reason based on past usage I'd seen.
It seems to have an [enormous list of configuration options](https://docs.renovatebot.com/configuration-options/) that are potentially useful, though even those don't handle cases like https://github.com/formium/tsdx/pull/889#discussion_r496208822 . Dependencies are pretty manual in my experience; updating all the time without a defined need just [adds a lot of](https://github.com/formium/tsdx/issues?q=author%3Asnyk-bot+label%3Asolution%2Funnecessary) [noise](https://github.com/formium/tsdx/pull/807#issuecomment-676816960).

Also recently switched from Greenkeeper to have even more issues with Snyk (as linked above), then to Dependabot (#846 ), so I'm not clamoring to change and learn another potentially problematic bot, especially if I need to use two for security issues (which are the few that are quite important / aren't noise) 😕  But I'll keep it in my mind for the future

agilgur5:
I had checked this already when looking to update previously, but confirming here that I see no relevant breaking changes in rpts2 [v0.27.0](https://github.com/ezolenko/rollup-plugin-typescript2/releases/tag/0.27.0). The release notes aren't particularly detailed but [the diff](https://github.com/ezolenko/rollup-plugin-typescript2/compare/0.26.0...0.27.0) was small enough to read through and check.

agilgur5:
Reproduced and confirmed #892 as well as did some more investigations there, including finding that this bug did start with TSDX v0.13.3, just with less popular combinations of formats than v0.14.0 for unknown reason.

I'm not going to wait on the release of https://github.com/ezolenko/rollup-plugin-typescript2/pull/221 / rtps2 v0.28.0 anymore since this issue seems to be quite widespread, #898 and #900 hit into it as well.

I'll probably merge and release TSDX v0.14.1 tomorrow, but I'd like to add a smoke test that tests a build of all formats simultaneously to confirm this works and doesn't regress. Apparently there weren't any tests of `umd` or `system` before, can't remember why I didn't add them when I built out a lot of the test suite in #627... maybe I was just sticking to integration tests then.. 🤔

agilgur5 (approved):
New test passes, no weird logs in the output (aside from https://github.com/formium/tsdx/pull/884#issuecomment-707347469 which is an existing issue). Test conforms to existing test style and pretty straightforward so that LGTM to me as well.

agilgur5:
@allcontributors please add @tanem for bug, code

JustFly1984:
@agilgur5 
I have the feedback.
I had released 2 new versions of my library @react-google-maps/api

1.12.1 includes tsdx update to 0.14.1
1.13.0 includes updated webpack@5.0.0

I see the bundle size grow from 15.2kb to 17.5kb for no obvious reason.
https://bundlephobia.com/result?p=@react-google-maps/api@1.12.1 - bundle size is grown after update tsdx
https://bundlephobia.com/result?p=@react-google-maps/api@1.13.0 - bundle size did not changed by updating webpack version

There is definitely some degradation in build tool optimizations.

agilgur5:
> for no obvious reason.
> There is definitely some degradation in build tool optimizations.

@JustFly1984 you upgraded from v0.13.3 to v0.14.1, i.e. two versions, one of which was a breaking change. The [v0.14.0 Release Notes](https://github.com/formium/tsdx/releases/tag/v0.14.0) are very explicit as to possible bundle size increases and even give directions on reductions, please read those.

As such, this is unrelated to v0.14.1 or this PR, so this is not the right place either.

JustFly1984:
@agilgur5 thank you for clarifications!

tanem:
Thanks for the merge @agilgur5 🙏

Issue #206, opened 2019-09-09
Who's using TSDX?


jaredpalmer:
Drop your company name, svg logo, link to your website, and links to any repos using TSDX. Going to add these to the website. 

From just scanning through the GitHub dependents tab, found some notable companies using it...

- https://github.com/microsoft/ability-attributes
- https://github.com/pelotoncycle/frame-throttle
- https://github.com/artsy/arc
- https://github.com/mobxjs/mobx-react-lite
- https://github.com/react-spring/react-use-gesture
- https://github.com/expo/copy-package

JustFly1984:
https://github.com/JustFly1984/react-google-maps-api/tree/master/packages/react-google-maps-api
https://github.com/JustFly1984/react-google-maps-api/tree/master/packages/react-google-maps-api-marker-clusterer
https://github.com/JustFly1984/react-google-maps-api/tree/master/packages/react-google-maps-api-infobox
https://raw.githubusercontent.com/JustFly1984/react-google-maps-api/master/logo.png

sw-yx:
Netlify [react-netlify-identity-widget](https://github.com/sw-yx/react-netlify-identity-widget) [Logos](https://www.netlify.com/press/)

netzwerg:
I use it in my latest lib: https://github.com/netzwerg/react-svg-timeline – thanks for making TSDX ✨

liyuanqiu:
TSDX helps me creating libs in this github org:
https://github.com/final-state
👍👍👍🙏🙏🙏

stolinski:
https://github.com/leveluptuts/fresh migrated from Create React Library

elado:
[Lyft](http://lyft.com) 
<img src=https://user-images.githubusercontent.com/51488/65556306-0471fe00-dee4-11e9-8ee5-8615878cf906.png height=200 />

whinc:
I use tsdx to startup my new github project:

https://github.com/whinc/api-sharp

👍🙏

zachariahtimothy:
@jaredpalmer once again from the govvy world :) ...
http://www.koniaggss.com/
And very shortly: https://www.aphis.usda.gov/aphis/home/

lookfirst:
https://github.com/lookfirst/mui-rff

zcuric:
Built my first TS project with `tsdx`, a logging utility. 

https://github.com/zcuric/sql-log-prettifier

alexvcasillas:
[...]

PR #899, opened 2020-10-09
Create CODE_OF_CONDUCT.md

jaredpalmer:
In response to the back and forth today. This is our CoC going forward. Period.

https://github.com/formium/tsdx/issues/206#issuecomment-706270555
https://github.com/formium/tsdx/issues/206#issuecomment-706373507

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
+
```
agilgur5, line 33: If you want to use the Contributor Covenant, I think several of these should have a clear "Good/Bad" example that specifically illustrates said behavior so there is less ambiguity.

This was very painful, but I went through some abusive instances in TSDX's history that occurred to multiple maintainers and paraphrased to make examples:
```md
* Snarky, sarcastic, condescending, or passive-aggressive behavior:
❌  Bad: "Will you fix this issue, or should I just not use this library?"
✅  Good: "I'd like to use this library, but this issue is a blocker for me right now."

❌  Bad: "How many more breaking bugs do you need?"
✅  Good: "These bugs have been very frustrating for me, so I would appreciate if changes respected SemVer more carefully and testing were a higher priority".

❌  Bad: "_Really_?"
✅  Good: "I believe this line doesn't match up with what you said, but correct me if I'm wrong"

* Trolling, insulting/derogatory comments, and personal or political attacks
❌  Bad: "You shouldn't have closed my issue, you're a jerk"
✅  Good: "I disagree that it should be closed and think this issue is substantial for these reasons [...]. Maybe this will be reconsidered down the line"

❌  Bad: "Who put _you_ in charge??"
✅  Good: "I disagree with this approach and think it could be improved in these ways [...]. Maybe this can be reconsidered on that basis"

❌  Bad: "You're uninformed and don't know what you're talking about. Your opinion is wrong. I'll take my chances with other libraries."
✅  Good: "Here are some examples that illustrate my point [...]. I disagree with this opinion and think the reverse trade-offs are better to make. I hope these examples can better inform the community's decisions, but I feel strongly enough about these trade-offs that I'll be exploring other approaches."
```

On the positive side, here's a few other examples:
```md
❌  Bad: "My word is final."
✅  Good: "Here's a few reasons why I don't think that is a good approach [...]. The examples and evidence here are not currently enough to sway my opinion, but if there is more community support, maybe it can be reconsidered then."

❌  Bad: "I'm not going to use that"
❌  Bad: "I don't like that"
❌  Bad: "That's a bad option"
✅  Good: "I don't agree with that option because of these trade-offs: [...]. If more positive trade-offs outweigh the cons, maybe I'll reconsider"
```

agilgur5 (commented):
I've been considering adding a CoC to TSDX, all of my repos, and all of the templates for quite a while now, but there is a reason that I didn't do so myself: I'd like to mix good parts from various CoCs out there to draft something that addresses several areas and is more detailed, concrete, and specific.

The Contributor Covenant, for example, has really short, somewhat abstract examples that leave it open to significant interpretation or misinterpretation. There's even an example in TSDX itself (edited out specific issue) of using the Contributor Covenant while simultaneously being abusive to several maintainers and core contributors.
In mine and some of my colleagues' experience in management, we've found that concrete examples add a lot more clarity and directly encourage or discourage certain behaviors by illustrating them directly and offering better alternatives from which to improve.

The [GitHub Community Forum's Code of Conduct](https://docs.github.com/en/free-pro-team@latest/github/site-policy/github-community-forum-code-of-conduct), for example, is still somewhat abstract, but lists more examples and is quite descriptive on both the positive and unacceptable sides (including staying on-topic, tidiness, discrimination, bullying, etc)

agilgur5 (commented):
> Period.

This is an unnecessary and condescending remark that I believe violates the Contributor Covenant... in a PR that adds the Contributor Covenant. I don't appreciate the condescending and hypocritical remarks you've made a few times to me now.

> [#206 (comment)](https://github.com/formium/tsdx/issues/206#issuecomment-706270555)
> [#206 (comment)](https://github.com/formium/tsdx/issues/206#issuecomment-706373507)

That user made a passive-aggressive comment after I simply marked their issue as duplicate. They then went to a totally unrelated thread (which I don't believe is "best for the community") to misleadingly attack me. We both agreed those were attacks as well.

I defended myself, pointed out the evidence and effort against their inaccuracies, pointed out the attacks. I did not attack that user. I think that user violated this CoC a number of times while I pointed out their violations.

Defending myself is not the same attacking someone, so I don't appreciate this false equivalence and victim-blame. I believe victim-blaming is quite against the intent of the Contributor Covenant. I'm not sure what I did to deserve that, and I don't think anyone deserves to be victim-blamed.

I'm asking again that you please consider how your words and actions have and continue to significantly damage my mental health and what that means for me and the community.

PR #903, opened 2020-10-14
Refactor build-default tests for more explicitness and correctness checks

agilgur5:
#### refactor: move unbundled regenerator test to build-options
- I accidentally put it into build-default before build-options existed,
  but the test very specifically tests an option, `--target node`, so
  it should be in build-options as that's not a zero-config default

#### refactor: use a less noisy side-effect for async regression test
- the previous console.log statement would actually log out because
  one of the tests imports (`require`s) the actual built library in
  order to ensure correctness

#### fix/docs: basic README should reference .ts, not .tsx files
- it uses normal .ts files and no JSX (that's what the React template
  is for after all), so README shouldn't mention .tsx

- woops, this was probably a copy+paste error on my behalf when I was
  improving the basic template's docs

This one is unrelated to the rest, just found it as I was refactoring for the previous commit

#### refactor/test: test for correctness of syntax, not just parsing
- have most of the syntax tests export a function that returns true
  if the syntax is evaluated correctly, and test that it does in fact
  return true

## Tags

- Fixes everything in https://github.com/formium/tsdx/pull/795#pullrequestreview-492994343 that I missed in initial review of #795 it. `build-options` was only _just_ added with my smoke test in #896, so I must've forgot to split out files before.
- Basic template using ".tsx" in the README was a mistake I made in #706 
- #46 previously removed some swear words, but not all
- Syntax tests are from several PRs over time

## Review Notes

Been wanting to do the explicitness for a while. Consistency in the syntax testing I thought would help (for my own use writing tests too) and correctness seemed to make more sense and was something I already did with the generator test (so this makes the others consistent in checking for correctness as well). I didn't actually add correctness _tests_ (i.e. checking the value) until now, maybe because I forgot I already had a place in the test suite to do so.

This doesn't change any actual source code or meaningful template code.

Will **rebase** these in.


agilgur5 (commented):
LGTM. A lot of files changed, but mostly one-liners. Biggest change was moving the regenerator test to `build-options`.

Not sure why Windows test on Node 10 is failing, one of the integration tests for error extraction (which doesn't totally work, so is only a partial test) failed, but that wasn't changed (only `build-default` was). re-running..

agilgur5:
Test passes now, not sure why it failed that one time. First time I've seen that error and it's in a section of the test suite that I didn't change and a not so important test, so going to merge.

PR #904, opened 2020-10-14
deps: ignore dependabot for peer dependencies

agilgur5:
## Description

- several of the dependencies in the tree are only there because they
  are peerDeps, and not because we directly depend on them
  - they should only be upgraded in tandem with the dep(s) they are
    peers of and pretty much never in isolation

- Greenkeeper made several PRs for peers which previously caused an
  erroneous/buggy update of `@types/jest` to 25 many months before Jest
  was upgraded to 25
  - and honestly those Greenkeeper PRs for peers had confused me a
    good deal too

## Tags

- #462 was a previous erroneous PR due to Greenkeeper updating peers, which was reverted in #470 
- This directly addresses some of my previous confusion from `eslint-config-react-app` peers in https://github.com/formium/tsdx/pull/640#issuecomment-702894150 / https://github.com/formium/tsdx/pull/642#issuecomment-702892887 / https://github.com/formium/tsdx/pull/728#issuecomment-702892139 / https://github.com/formium/tsdx/pull/729#issuecomment-702892420 / https://github.com/formium/tsdx/pull/890#issuecomment-700292943 
- #709 updated `jest-watch-typeahead` to a version for Jest 26 before we even upgraded to Jest 25

Let's close some of the confusing Greenkeeper peer update PRs now that this has been configured:
- Closes #640
- Closes #642
- Closes #728
- Closes #729
- Closes #709

## Review Notes

I straight copy+pasted the `eslint-config-react-app` peerDeps from [its `package.json`](https://github.com/facebook/create-react-app/blob/7641a3c1c6d02cd707ac11e7b8b8907acf3df7c0/packages/eslint-config-react-app/package.json#L17), so they're alphabetized the same way and shouldn't have any spelling mistakes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,20 @@ updates:
       # only upgrade prod deps (not devDeps)
       - dependency-name: '*'
         dependency-type: production
+    ignore:
+      # ignore eslint-config-react-app's peerDeps
+      - dependency-name: '@typescript-eslint/eslint-plugin'
+      - dependency-name: '@typescript-eslint/parser'
+      - dependency-name: 'babel-eslint'
+      - dependency-name: 'eslint-plugin-flowtype'
+      - dependency-name: 'eslint-plugin-import'
+      - dependency-name: 'eslint-plugin-jsx-a11y'
+      - dependency-name: 'eslint-plugin-react'
+      - dependency-name: 'eslint-plugin-react-hooks'
+      # ignore Jest's "peers" that should be upgraded simultaneously with Jest
+      - dependency-name: '@types/jest'
+      - dependency-name: 'jest-watch-typeahead'
+      - dependency-name: 'ts-jest'
     # temporarily disable dep upgrade PRs for / as they're being updated
     open-pull-requests-limit: 0
 


agilgur5 (commented):
peerDeps all look to line up to me, but I think `tslib` should probably be included in here too because of https://github.com/formium/tsdx/pull/719#issuecomment-683470971.

But somewhat problematically, it [doesn't seem to have a peerDep on `typescript`](https://github.com/microsoft/tslib/blob/76bedfddc2cad38aa11060a32f56e9484c279c98/package.json), meaning one wouldn't get a peerDep mismatch error if an incorrect version were installed. And that means it could be easy to forget to check/upgrade `tslib` in tandem with `typescript`... Hmmm, need to think about that.

This might be unintentional, so should check upstream in `tslib` and perhaps file a PR.

agilgur5:
Seems like https://github.com/microsoft/tslib/issues/22 exists that goes into this. https://github.com/microsoft/tslib/issues/22#issuecomment-318492123 seemed against using peerDeps but without an example that was difficult to understand how the use-case actually plays out (does it mean in the dependency tree or a single package?)

agilgur5:
Since older versions of `tslib` seem to be forward-compatible, it seems like the decision to upgrade `tslib` would be independent of TS upgrade as a result, so might not make sense to ignore it then.

Added a comment in the thread upstream, but think this is good to go as is due to above. Can always add more to the list as well.

agilgur5:
Cleaned up ~74 Greenkeeper branches associated with `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `eslint-plugin-flowtype`, `eslint-plugin-import`, `eslint-plugin-react`, `eslint-plugin-react-hooks`, `@types/jest`, and `jest-watch-typeahead`.

Think nearly all the rest can be cleaned up now too. I think they're mostly minor version upgrades (hence no PR) or devDeps or both. Has taken a while to get here, have cleaned up _100s_ of Greenkeeper branches 😕

agilgur5:
Deleted another ~12 Greenkeeper branches. They were mostly patch bumps of devDeps actually, a few minors and a few prod. Remaining ones seem to be all Prettier/`eslint-config-prettier`/`eslint-plugin-prettier` deps or template version bumps in the `example` dir

PR #909, opened 2020-10-20
Refactor/Improve some typings

agilgur5:
#### types: add remaining external typings
- added @types/babel__core, @types/babel__traverse, and
  @types/lodash.merge and removed their declarations in env.d.ts
  - also removed rollup-plugin-terser's declaration there because it
    has types built-in
  - remaining actually don't have types per in-line comments

- this improved type-checking allowed TS to do stricter checks and so
  exposed a few type errors that were fixed as well
  - @rollup/plugin-babel depends on @types/babel__core as well, so the
    declaration here had previously overridden it
    - `custom` had to be augmented as it's not an option of
      `@rollup/plugin-babel`
    - `passPerPreset` did too though... it's not in `@types/babel__core`
      but it is a (deprecated) Babel option
      - logs seemed to show it was `false` regardless of what was
        passed in though...
  - Rollup's `plugins` now started complaining that `false` wasn't a
    valid Plugin either, so added a `.filter(Boolean)` but that didn't
    fully fix it so had to type-cast

#### refactor/fix: consistently use capital TSDX everywhere
- Jared had asked in a review comment back in December to consistently
  use capital TSDX (not `tsdx`) so this establishes that throughout the
  rest of the codebase
  - I don't think any of these lines/files/variables were written by me,
    so I'm just fixing up existing ones

- this came up as I was refactoring typings, so wanted to make sure
  everything was consistent as I went

#### refactor/types: add a TSDXConfig type
- another PR by a contributor exposes the types for `tsdx.config.js` in
  order to be used in JSDoc, so this makes it so only a single type is
  needed and exported

## Tags

- Using capital TSDX instead of `tsdx`, "tsdx", etc was requested in https://github.com/formium/tsdx/pull/400#discussion_r361858698. Maybe in an earlier issue/PR too, but I couldn't find or remember it.
- The "at least one" type confusion was in https://github.com/formium/tsdx/pull/401#discussion_r361774071
- `TSDXConfig` type is pretty much just to simplify #823 

## Review Notes

The `passPerPreset` augmentation for `@rollup/plugin-babel` / `@types/babel__core` that is mentioned above is pretty confusing. I decided to leave it as is for now so as to potentially change anything, but my logs seem to show it isn't used 😕 🤔  :

<details>
  <summary>Babel log output:</summary>

```ts
PartialConfig {
  options:
   { sourceMaps: true,
     caller:
      { name: '@rollup/plugin-babel',
        supportsStaticESM: true,
        supportsDynamicImport: true,
        supportsTopLevelAwait: true },
     filename:
      '[redacted]/tsdx/stage-build-default/src/syntax/jsx-import/JSX-B.jsx',
     babelrc: false,
     configFile: false,
     passPerPreset: false,
     envName: 'test',
     cwd:
      '[redacted]/tsdx/stage-build-default',
     root:
      '[redacted]/tsdx/stage-build-default',
     plugins: [],
     presets: [] },
  babelignore: undefined,
  babelrc: undefined,
  config: undefined }
{ sourceMaps: true,
  caller:
   { name: '@rollup/plugin-babel',
     supportsStaticESM: true,
     supportsDynamicImport: true,
     supportsTopLevelAwait: true },
  filename:
   '[redacted]/tsdx/stage-build-default/src/syntax/jsx-import/JSX-B.jsx',
  babelrc: false,
  configFile: false,
  passPerPreset: false,
  envName: 'test',
  cwd:
   '[redacted]/tsdx/stage-build-default',
  root:
   '[redacted]/tsdx/stage-build-default',
  plugins:
   [ ConfigItem {
       value: [Function],
       options: {},
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] },
     ConfigItem {
       value: [Function: _default],
       options: {},
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] },
     ConfigItem {
       value: [Function],
       options: {},
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] },
     ConfigItem {
       value: [Function],
       options: [Object],
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] },
     ConfigItem {
       value: [Function],
       options: [Object],
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] } ],
  presets:
   [ ConfigItem {
       value: [Function],
       options: [Object],
       dirname:
        '[redacted]/tsdx/stage-build-default',
       name: undefined,
       file: [Object] } ] }
PartialConfig {
  options:
   { sourceMaps: true,
     caller:
      { name: '@rollup/plugin-babel',
        supportsStaticESM: true,
        supportsDynamicImport: true,
        supportsTopLevelAwait: true },
     filename:
      '[redacted]/tsdx/stage-build-default/src/syntax/jsx-import/JSX-B.jsx',
     babelrc: false,
     configFile: false,
     passPerPreset: false,
     envName: 'test',
     cwd:
      '[redacted]/tsdx/stage-build-default',
     root:
      '[redacted]/tsdx/stage-build-default',
     plugins: [],
     presets: [] },
  babelignore: undefined,
  babelrc: undefined,
  config: undefined }
```
</details>

### DIFF
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -1,6 +1,6 @@
 import { safeVariableName, safePackageName, external } from './utils';
 import { paths } from './constants';
-import { RollupOptions } from 'rollup';
+import { RollupOptions, Plugin } from 'rollup';
 import { terser } from 'rollup-plugin-terser';
 import { DEFAULT_EXTENSIONS as DEFAULT_BABEL_EXTENSIONS } from '@babel/core';
 import commonjs from '@rollup/plugin-commonjs';
@@ -213,6 +213,6 @@ export async function createRollupConfig(
           toplevel: opts.format === 'cjs',
           warnings: true,
         }),
-    ],
+    ].filter(Boolean) as Plugin[],
   };
 }
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -14,8 +14,8 @@ import typescript from 'rollup-plugin-typescript2';
 import ts from 'typescript';
 
 import { extractErrors } from './errors/extractErrors';
-import { babelPluginTsdx } from './babelPluginTsdx';
-import { TsdxOptions } from './types';
+import { babelPluginTSDX } from './babelPluginTSDX';
+import { TSDXOptions } from './types';
 
 const errorCodeOpts = {
   errorMapFilePath: paths.appErrorsJson,
@@ -25,7 +25,7 @@ const errorCodeOpts = {
 let shebang: any = {};
 
 export async function createRollupConfig(
-  opts: TsdxOptions,
+  opts: TSDXOptions,
   outputNum: number
 ): Promise<RollupOptions> {
   const findAndRecordErrorCodes = await extractErrors({
@@ -71,7 +71,7 @@ export async function createRollupConfig(
     // Rollup has treeshaking by default, but we can optimize it further...
     treeshake: {
       // We assume reading a property of an object never has side-effects.
-      // This means tsdx WILL remove getters and setters defined directly on objects.
+      // This means TSDX WILL remove getters and setters defined directly on objects.
       // Any getters or setters defined on classes will not be effected.
       //
       // @example
@@ -184,7 +184,7 @@ export async function createRollupConfig(
         check: !opts.transpileOnly && outputNum === 0,
         useTsconfigDeclarationDir: Boolean(tsCompilerOptions?.declarationDir),
       }),
-      babelPluginTsdx({
+      babelPluginTSDX({
         exclude: 'node_modules/**',
         extensions: [...DEFAULT_BABEL_EXTENSIONS, 'ts', 'tsx'],
         passPerPreset: true,
--- a/src/types.ts
+++ b/src/types.ts
@@ -8,6 +8,7 @@ interface SharedOpts {
 }
 
 export type ModuleFormat = 'cjs' | 'umd' | 'esm' | 'system';
+export type AtLeastOneModuleFormat = [ModuleFormat, ...ModuleFormat[]];
 
 export interface BuildOpts extends SharedOpts {
   name?: string;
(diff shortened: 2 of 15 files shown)


src/createRollupConfig.ts
```diff
@@ -213,6 +213,6 @@ export async function createRollupConfig(
           toplevel: opts.format === 'cjs',
           warnings: true,
         }),
-    ],
+    ].filter(Boolean) as Plugin[],
```
agilgur5, line 216: This is a simple one and doesn't change any behavior, but the type-cast seemed to have eroded some typing requirements that exist without it. E.g. Rollup requires each `Plugin` to have a `name` property, and some of these custom ones here do not.

src/types.ts
```diff
@@ -60,3 +64,9 @@ export interface PackageJson {
     node?: string;
   };
 }
+
+export type RollupOptionsWithOutput = RollupOptions & { output: OutputOptions };
+
+export interface TSDXConfig {
+  rollup: (config: RollupOptions, options: TSDXOptions) => RollupOptions;
+}
```
agilgur5, line 72: This still has the same problem as in https://github.com/formium/tsdx/pull/823#discussion_r503081245 . Should probably try being more specific here and in `createRollupConfig.ts`.

agilgur5 (commented):
Mostly looks good to me. Mostly just straightforward renames.

The external types portion is the most potentially confusing bit, since the type-checking just ignored some Babel and Rollup stuff before due to the lack of declarations. Babel stuff is necessary and fairly straightforward at least.

A few things I'd like to take a closer look at though

src/createRollupConfig.ts
agilgur5: Ahh think I need a type predicate here per https://github.com/microsoft/TypeScript/issues/20812. I've been using these type-casts after `filter` like #54 did, but a type predicate / type guard is the right way to do it that I believe will prevent eroding of types

agilgur5: So I tried the below:
```typescript
].filter((e): e is Plugin => Boolean(e))
```

but that didn't fully work either. It also eroded some types, like the `name` property requirement mentioned above 😕 

With the `TSDXOptions` cast it didn't work because it doesn't seem to be able to infer the `AtLeastOne` portion 😕 

For this Rollup config piece, I was able to rewrite the `condition && plugin` to `...(condition ? [plugin] : [])` which I've done before elsewhere, and then that worked, but it is quite verbose... maybe I can abstract that into a function...

PR #914, opened 2020-10-27
change: by default lint all dirs, not just src and test

agilgur5:
## Description

- BREAKING: change the default lint args from 'src test' to './'

- this is the same arg that's used internally and there isn't really a
  reason to skip some files arbitrarily; it's best practice to lint
  everything
- this also lines up with changes to make `tsc` type-check all files,
  instead of just 'src' and 'types'

## Tags

- I made the same change internally in #648
- Goes along with `tsc` changes mentioned in https://github.com/formium/tsdx/issues/871#issuecomment-695822562 et al

## Notes

This is a breaking change. While it probably affects few users due to the fact that it gives a warning when nothing is specified, it is still breaking. Would make sense to add in [v0.16.0](https://github.com/formium/tsdx/milestone/4) to go with the other ESLint / formatting changes.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -540,17 +540,17 @@ prog
 prog
   .command('lint')
   .describe('Run eslint with Prettier')
-  .example('lint src test')
+  .example('lint ./')
   .option('--fix', 'Fixes fixable errors and warnings')
-  .example('lint src test --fix')
+  .example('lint ./ --fix')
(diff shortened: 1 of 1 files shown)


agilgur5 (commented):
LGTM, pretty small change 👍 

Now to wait some time before this can actually be merged...

agilgur5:
CI is failing because GitHub Actions appears to be down per the logs

Issue #848, opened 2020-08-31
Strip down readme and update website


jaredpalmer:
### Current Behavior

I forgot to strip down the readme when doing the site. It's confusing to have docs on both the readme and website as there is no single source of truth. 
### Expected behavior

The website is the source of truth for documentation. 

### Suggested solution(s)

Strip down readme to logo, one-liner description, badges, links to website, plus contributors. Update website to remove any discrepancies w/ current readme.

agilgur5:
I agree it's confusing if they're different (they are slightly right now), but if they're the same then 🤷  Could do the same thing as the Releases/Changelog page and just pull the README into the website.

I'm not sure I agree with the suggested solution. It's always annoying when you're on GitHub and you find some interesting repo but there's little to nothing in the README and it just points to a separate website. It requires another click which is generally not good UX (there's drop-off/that requires another conversion) and isn't great for SEO either

agilgur5:
Per my last email, I'd also appreciate if the Vercel bot were turned off as it's giving me a massive amount of unnecessary notifications instead of using GitHub status checks and only running on website PRs/commits like Netlify did.

jaredpalmer:
Gonna have to disagree with you about docs. As someone that has dealt with this before, Google figures out rather quickly when documentation is moved over. 

The readme should look like this: 

<img width="1151" alt="Screen Shot 2020-08-31 at 12 37 12 PM" src="https://user-images.githubusercontent.com/4060187/91743967-b307cf80-eb86-11ea-8289-626f3f167fcc.png">

agilgur5:
> As someone that has dealt with this before, Google figures out rather quickly when documentation is moved over.

That's not what I said -- I'm saying that it de-prioritizes the actual repo since the content is elsewhere. Things like this can also happen:

here's `mobx-state-tree`'s results:

![Screen Shot 2020-08-31 at 12 42 39 PM](https://user-images.githubusercontent.com/4970083/91744519-a0da6100-eb87-11ea-8fd6-0c80ca8acf73.png)

vs. Jest's results:

![Screen Shot 2020-08-31 at 12 42 58 PM](https://user-images.githubusercontent.com/4970083/91744636-c5ced400-eb87-11ea-8e30-953d827537de.png)


[...]

PR #923, opened 2020-11-04
fix: relative links in website to absolute links

felixmosh:
_Maintainer edit_:

Closes #844

### DIFF
--- a/website/pages/customization.md
+++ b/website/pages/customization.md
@@ -73,12 +73,12 @@ module.exports = {
 
 ## Babel
 
-You can add your own `.babelrc` to the root of your project and TSDX will **merge** it with [its own Babel transforms](./src/babelPluginTsdx.ts) (which are mostly for optimization), putting any new presets and plugins at the end of its list.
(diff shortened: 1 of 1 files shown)


agilgur5:
Supersedes #844 with a valid solution to another issue with #765 , though my remarks in https://github.com/formium/tsdx/pull/844#issuecomment-683764069 still stand, especially as the website has no versioning. The website is also a bit outdated now because it was never synced with the README (c.f. https://github.com/formium/tsdx/pull/844#issuecomment-683448861, #848).

This is still a good change though, so will accept this as incremental improvement even though it does not solve all the problems.

agilgur5 (approved):
URLs are correct and I've tested that they work in the preview site 👍 

Tests seem to be timing out again... will have to re-run those

agilgur5:
Another test timeout but only 1 this time. I'll merge since this is quite safe to change, doesn't affect any actual code, and was manually tested.

agilgur5:
@allcontributors please add @felixmosh for docs

PR #925, opened 2020-11-04
Update gitignore template to add coverage

### LINKED ISSUE #922: coverage not part of .gitignore
### Current Behavior

1. run `yarn test --coverage` to generate a coverage report.
1. run `git status`
1. `/coverage/` is part of the list of changed files.

### Expected behavior

It should be ignored by git.

### Suggested solution(s)

Append `coverage` to `.gitignore`

mimshwright:
fixes #922

### DIFF
--- a/templates/basic/gitignore
+++ b/templates/basic/gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 node_modules
 dist
+coverage


templates/basic/gitignore
```diff
@@ -2,3 +2,4 @@
 .DS_Store
 node_modules
 dist
+coverage
```
agilgur5, line 5:
```suggestion
# Jest test coverage reports
coverage
```

Could we add a comment here? I'm planning on adding comments to other portions of this file similar to #864

agilgur5 (changes requested):
@mimshwright good find that the templates were missing this! 

You've changed the `basic` template, but this change should also be added to the other two templates as well for consistency (and because they all use `tsdx test` / Jest) -- could you please add this change in those templates as well?

I've also requested a small change in-line that should go in all 3.

PR #942, opened 2020-12-05
Add option to not show progress animations in logs

garrettjstevens:
Fixes https://github.com/formium/tsdx/issues/689. 

Adds a --noProgress flag (based on lerna's `--no-progress`, but made camelCase to match existing flags). When used, the `progress-estimator` logger is replaced by a simple `console.log` of the message. The progress animations will also be removed when the environment is not a TTY and when running in CI.

Thanks for a great tool, and let me know if there's anything I can do to improve this PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -382,11 +382,19 @@ prog
   .example(
     'build --extractErrors=https://reactjs.org/docs/error-decoder.html?invariant='
   )
+  .option('--noProgress', "Don't show progress animations")
+  .example('build --noProgress')
   .action(async (dirtyOpts: BuildOpts) => {
     const opts = await normalizeOpts(dirtyOpts);
     const buildConfigs = await createBuildConfigs(opts);
     await cleanDistFolder();
-    const logger = await createProgressEstimator();
+    const progress =
+      process.stdout.isTTY && !process.env.CI && !opts.noProgress;
+    const logger = progress
+      ? await createProgressEstimator()
+      : (_promise: unknown, message: string) => {
+          console.log(message);
+        };
     if (opts.format.includes('cjs')) {
       const promise = writeCjsEntryFile(opts.name).catch(logError);
       logger(promise, 'Creating entry file');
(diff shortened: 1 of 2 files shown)


jaredpalmer (commented):
LGTM

agilgur5 (changes requested):
To clarify, this is _not_ a full, formal review, and normally I would not review until it is full, but there's a handful of issues (and improvements) with the PR as is:
- Docs were not updated
- The comment in `SharedOptions` is not correct (changing `noProgress` in the config does not change it, so the word "Should" is confusing and potentially misleading; "Is output progress being shown?" would be better)
- Only 1 command was updated (`build` and `watch` should both be updated)
- The `isTTY` portion is a breaking change and therefore can only go out in a breaking release. I would recommend that part be separated into a different PR as such.
- I think this check should be embedded into `createProgressEstimator` instead of in individual uses. That will also make it useful when other functions use `createProgressEstimator`
- Typings could be improved/more specific (`Promise<any>` instead of `unknown`)
- Is there a way to infer Lerna's current progress output status? That way there would not necessarily be a need for a new flag, which is what I was hoping to avoid in the issue

garrettjstevens:
Thanks @agilgur5, I'll work on addressing those issues.

garrettjstevens:
A few comments on my last commit:

> - The comment in SharedOptions is not correct (changing noProgress in the config does not change it, so the word "Should" is confusing and potentially misleading; "Is output progress being shown?" would be better)

Since the option is a negative, I went with "Are output progress animations disabled?". Is that ok?

> - Only 1 command was updated (build and watch should both be updated)

I've added the option to `watch` as well. I did this by using the `isEnabled` option of the `ora` spinner.

> - The isTTY portion is a breaking change and therefore can only go out in a breaking release. I would recommend that part be separated into a different PR as such.

It's now removed from this PR. I'll open a draft PR with that change separately.

> - Is there a way to infer Lerna's current progress output status? That way there would not necessarily be a need for a new flag, which is what I was hoping to avoid in the issue

In my use case I'm not using Lerna, so I'm hoping we can add this new flag.

PR #944, opened 2020-12-07
fix: pin husky so create does not fail on dep install

FaberVitale:
Pins husky version to 4.3.0

issue:  https://github.com/formium/tsdx/issues/943

husky issue: https://github.com/typicode/husky/issues/822

### Update

Husky has been [rolled back to 4.3.0](https://github.com/typicode/husky/issues/822#issuecomment-739760031), feel free to disregard this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "babel-plugin-replace-identifiers": "^0.1.1",
     "cssnano": "^4.1.10",
     "doctoc": "^1.4.0",
-    "husky": "^4.2.2",
+    "husky": "4.3.0",
     "np": "^6.4.0",
     "pretty-quick": "^2.0.0",
     "react": "^16.8.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "babel-plugin-replace-identifiers": "^0.1.1",
     "cssnano": "^4.1.10",
     "doctoc": "^1.4.0",
-    "husky": "^4.2.2",
```
agilgur5, line 116: This is the devDep that is used internally, and not the one that is in the templates. The one in the templates is here: https://github.com/formium/tsdx/blob/462af2d002987f985695b98400e0344b8f2754b7/src/templates/basic.ts#L6

agilgur5 (changes requested):
So the changes made here do not actually resolve or affect the issue, as I've written in my review comments. CI also did not pass due to a dep issue.

In any case, a change is not necessary as Husky has reverted this change, so I will be closing this.

> Husky has been [rolled back to 4.3.0](https://github.com/typicode/husky/issues/822#issuecomment-739760031), feel free to disregard this PR.

Not sure if this is a typo, [_v4.3.5_](https://github.com/typicode/husky/releases/tag/v4.3.5) reverts the change in v4.3.4; can see [the diff and commit log here](https://github.com/typicode/husky/compare/v4.3.4...v4.3.5). I'm not sure it's equivalent to v4.3.0 and a library can't really roll back per se, it can only roll forward and publish a new version.

PR #957, opened 2021-01-18
feature: shebang banner

### LINKED ISSUE #338: Insert bin prefix for script packages
### Current Behavior

The output of `index.js` begins with `use strict` which if the package is intended to be a script, an error outputs.

### Desired Behavior

When a package.json includes a bin parameter, prefix the `index.js` file with `#!/usr/bin/env node`

This is for users who wish to use tsdx for creating packages that are intended to be used in the command line.

### Additional context

If it doesn't make sense to base the behavior on the bin parameter in the package.json, it might make sense instead to make this a new template type, but it's so similar to node, that this might be more effort than necessary.

huozhi:
Use `rolloup.output.banner` to add shebang header into bundle. Resolves #338 

### Changes

* add `--bin` flag
* if `bin` is specified in package.json, output with shebang header

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,6 +5,8 @@ interface SharedOpts {
   tsconfig?: string;
   // Is error extraction running?
   extractErrors?: boolean;
+  // Is executive?
+  bin?: boolean;
 }
 
 export type ModuleFormat = 'cjs' | 'umd' | 'esm' | 'system';
@@ -52,6 +54,7 @@ export interface TsdxOptions extends SharedOpts {
 export interface PackageJson {
   name: string;
   source?: string;
+  bin?: string | { [binary: string]: string };
   jest?: any;
   eslint?: any;
   dependencies?: { [packageName: string]: string };
--- a/test/e2e/tsdx-build-default.test.ts
+++ b/test/e2e/tsdx-build-default.test.ts
@@ -41,6 +41,13 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(output.code).toBe(0);
   });
 
+  it('should add shebang', async () => {
+    const output = execWithCache('node ../dist/index.js build');
+    expect(output.code).toBe(0);
+    const matched = grep(/#!\/usr\/bin\/env/, ['dist/build-default.*.js']);
+    expect(matched).toBeTruthy();
+  });
+
   it('should create the library correctly', async () => {
     const output = execWithCache('node ../dist/index.js build');
 
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -45,6 +45,7 @@ export async function createRollupConfig(
   ]
     .filter(Boolean)
     .join('.');
+  const banner = opts.bin ? `#!/usr/bin/env node` : undefined;
 
   const tsconfigPath = opts.tsconfig || paths.tsconfigJson;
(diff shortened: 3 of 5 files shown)


huozhi:
not sure why it's failed with node 12 + macOS, I'm currently using those as my local development env and installation + tests are passed locally.. 😂

test/e2e/tsdx-build-default.test.ts
```diff
@@ -41,6 +41,13 @@ describe('tsdx build :: zero-config defaults', () => {
     expect(output.code).toBe(0);
   });
 
+  it('should add shebang', async () => {
+    const output = execWithCache('node ../dist/index.js build');
```
agilgur5, line 45: To test shebangs, you'd want to exec `. ./dist/index.js` so that the shebang gets used

test/e2e/tsdx-build-default.test.ts
```diff
@@ -41,6 +41,13 @@ describe('tsdx build :: zero-config defaults', () => {
...
+    expect(output.code).toBe(0);
+    const matched = grep(/#!\/usr\/bin\/env/, ['dist/build-default.*.js']);
```
agilgur5, line 47: This test isn't checking the most important file, `index.js`, which will not have it, which is why this PR will not work

src/types.ts
```diff
@@ -5,6 +5,8 @@ interface SharedOpts {
   tsconfig?: string;
   // Is error extraction running?
   extractErrors?: boolean;
+  // Is executive?
```
agilgur5, line 8:
```suggestion
  // Is executable?
```
"executive" means something different

src/types.ts
```diff
@@ -52,6 +54,7 @@ export interface TsdxOptions extends SharedOpts {
 export interface PackageJson {
   name: string;
   source?: string;
+  bin?: string | { [binary: string]: string };
```
agilgur5, line 57: should be `Boolean` like the docs and like `transpileOnly`

agilgur5 (changes requested):
Sorry but this approach does not work -- per  the comments #338 , the central problem is that the dev/prod entry, `index.js` was not written as a Rollup plugin and as such is written out-of-band without knowledge of shebangs.
Per the in-line comments below, `index.js`, which is the `bin` entry is not changed and the tests do not check for it properly

The correct approach, as I wrote there, is to convert that to a Rollup plugin and do shebang detection as well (no need for a `--bin` flag, though it could potentially be useful for newbies). Shebang detection was also partially removed at some point in time.

The plugin I'm working on does both of these things

huozhi:
got it, thanks for clarifying @agilgur5

huozhi:
@agilgur5 is that ok to use rollup-plugin-preserve-shebang to implement this feature? I can file another PR for it if the idea is acceptable for you

agilgur5:
@huozhi unfortunately that will run into the same issue, it won't fix the `index.js` file. You might notice from the codebase/repo that TSDX actually used it at one point (before it was partially removed).

The optimal solution is a good bit more complicated as it requires a nuanced Rollup plugin and refactoring several internals. I have a branch that does it, but haven't gotten around to testing it or prioritizing it (mostly because of issues written in my current status)

huozhi:
understood, manually generating index.js of cjs bundle is definitely a blocker for that solution. thanks for your patient explanation! 🙏

Issue #926, opened 2020-11-05
Support Typescript 4


vberlier:
### Current Behavior

Creating a new project with the example code from the [official release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types) doesn't work:

```ts
type Arr = readonly any[];

export function concat<T extends Arr, U extends Arr>(arr1: T, arr2: U): [...T, ...U] {
  return [...arr1, ...arr2];
}
```

```bash
$ yarn build
semantic error TS1256: A rest element must be last in a tuple type.
```

```bash
$ yarn lint
Parsing error: Cannot read property 'map' of undefined
```

### Workaround

For the moment I managed to work around it by updating `typescript` and `typescript-eslint` and forcing the resolutions:

```bash
$ yarn add --dev typescript@latest @typescript-eslint/eslint-plugin@latest @typescript-eslint/parser@latest
```
```json
// package.json

"resolutions": {
  "**/typescript": "^4.0.5",
  "**/@typescript-eslint/eslint-plugin": "^4.6.1",
  "**/@typescript-eslint/parser": "^4.6.1"
}
```

agilgur5:
Hmm.. So the linting portion duplicates #810 which just recently got unblocked by a new release of `eslint-config-react-app` upstream.

But oddly enough #810 doesn't report issues with any TS syntax. Perhaps there's a subset that's compatible but some of it isn't, per the TS release notes. 

TSDX normally uses [whatever version of TS is available in `node_modules`](https://github.com/formium/tsdx/blob/995099fd1d60104bca9ba334f8cd8c0835122840/src/createRollupConfig.ts#L153), which has allowed users to use newer versions of TS before TSDX upgraded to them. But I think since all versions were TS v3 previously, it would just get deduplicated / flattened in `node_modules` as a single TS version.
Since TS v4 released as a major (notably, TS doesn't follow SemVer as each minor has breaking changes in it too), I'm guessing this made two versions of TS appear in the `node_modules` tree and so TSDX ended up using the v3 which matches its [internal version](https://github.com/formium/tsdx/blob/995099fd1d60104bca9ba334f8cd8c0835122840/package.json#L99).
I actually wasn't sure exactly how this behavior would play out, so thanks for showing an example.

[...]

PR #985, opened 2021-03-05
fix: should respect user's typescript, tslib

lbwa:
Shortly, typescript and tslib should always be a member of `peerDependencies`, rather than `dependencies`. Otherwise, tsdx user never use their own tslib and typescript due to [nodejs module resolution rules](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).

## reproduce

1. install

```plaintext
npx tsdx create tsdx-demo
# choose one template
cd tsdx-demo
```

2. add some typescript v4 language feature in src/\*.ts file, for example, in `src/index.ts`：

```typescript
// src/index.ts
type Method = 'GET'
// Lowercase is typescript v4 language feature
export function oops(method: Lowercase<Method>) {
  return method
}
```

3. build

```bash
yarn build
# semantic error TS2304: Cannot find name 'Lowercase'.
```

But [Lowercase](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype) actually is typescript `Intrinsic String Manipulation Types` in v4

4. check all dependencies

```bash
// yarn v1.2.x
yarn list --pattern typescript
```

```plaintext
 yarn list --pattern typescript
 yarn list v1.22.5
 ├─ @typescript-eslint/eslint-plugin@2.34.0
 ├─ @typescript-eslint/experimental-utils@2.34.0
 ├─ @typescript-eslint/parser@2.34.0
 ├─ @typescript-eslint/typescript-estree@2.34.0
 ├─ rollup-plugin-typescript2@0.27.3
 ├─ tsdx@0.14.1
 │  └─ typescript@3.9.9
 └─ typescript@4.2.3
```

## why

Everything go back to normal when I delete typescript dir in `<PROJECT_ROOT>/node_modules/tsdx/node_modules/typescript`.  I wrote some code only supported by typescript v4, but tsdx only require typescript v3x.

**tsdx always require tsdx's typescript and tslib, rather than user's own version due to [nodejs module resolution rules](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).**

[This code](https://github.com/formium/tsdx/blob/v0.14.1/src/createRollupConfig.ts#L14) always requires typescript from tsdx's dependencies, rather than project's dependencies, [that code](https://github.com/formium/tsdx/blob/v0.14.1/src/templates/basic.ts#L9) typescript version **never works**!!

### conclusion

typescript should always be a member of peerDependencies. tslib too.

## solution

1. move typescript, tslib into peerDependencies field

1. throw importation error when typescript isn't installed, but tslib only throws a warning.

### for tsdx template user

Nothing needs to be done (because tsdx template will install typescript and tslib automatically)

### for standalone user (without tsdx template)

1. should always install typescript, tslib manually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,9 +94,7 @@
     "semver": "^7.1.1",
     "shelljs": "^0.8.3",
     "tiny-glob": "^0.2.6",
-    "ts-jest": "^25.3.1",
-    "tslib": "^1.9.3",
-    "typescript": "^3.7.3"
+    "ts-jest": "^25.3.1"
   },
   "devDependencies": {
     "@types/eslint": "^6.1.2",
@@ -123,7 +121,13 @@
     "styled-components": "^5.0.1",
     "tiny-invariant": "^1.1.0",
     "tiny-warning": "^1.0.3",
+    "tslib": "^1.9.3",
+    "typescript": "^3.7.3",
     "yarn-deduplicate": "^2.1.1"
+  },
+  "peerDependencies": {
+    "tslib": "^1.9.3 || ^2.0.0",
+    "typescript": "^3.7.3 || ^4.0.0"
   },
   "husky": {
     "hooks": {
--- a/src/createRollupConfig.ts
+++ b/src/createRollupConfig.ts
@@ -11,7 +11,6 @@ import resolve, {
 } from '@rollup/plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import typescript from 'rollup-plugin-typescript2';
-import ts from 'typescript';
 
 import { extractErrors } from './errors/extractErrors';
 import { babelPluginTsdx } from './babelPluginTsdx';
@@ -23,6 +22,22 @@ const errorCodeOpts = {
 
 // shebang cache map thing because the transform only gets run once
 let shebang: any = {};
+let ts: typeof import('typescript');
+try {
+  ts = require('typescript');
+} catch (error) {
+  throw new Error(
+    `The module "typescript" was not found. tsdx requires that you include it in "dependencies" of your "package.json". To add it, run "npm install typescript"`
+  );
+}
+
+try {
+  require.resolve('tslib');
+} catch (error) {
+  console.warn(
+    `The module "tslib" was not found. We use it to reduce your bundle size.`
+  );
+}
 
 export async function createRollupConfig(
   opts: TsdxOptions,


lbwa:
## why not just bump up typescript to v4 https://github.com/formium/tsdx/issues/926 ?

The key is tsdx **shouldn't** help user to make a choice which version user should use. In other words, tsdx shouldn't provide typescript and tslib for better compatibility(eg, [practice in next.js](https://github.com/vercel/next.js/blob/v10.0.7/packages/next/bin/next.ts#L6-L15)).

## use resolution field in package.json ? just like https://github.com/formium/tsdx/issues/926#issuecomment-751936109

It's not an elgant solution, and more mental burden. The important thing is that only works with yarn, not npm.

agilgur5:
@lbwa I appreciate the detailed write-up, but this is a known issue and not as simple as you suggest. For changes as ranging as this, I would definitely recommend making an issue first before jumping into a PR.
Please see https://github.com/formium/tsdx/issues/952#issuecomment-757523826 where I've previously answered the question as to why peerDeps aren't used (history, "all-in-one") and as to why peerDeps are not necessarily an optimal solution. In summary, lots of deps and TSDX itself can and do break easily and often due to TS changes, and TS does not follow SemVer and has a breaking change every minor (this would most likely result in more issues, not less). It is not quite like the React example you point out in Next -- React is one of the most, if not the most, stable package out there, even React's breaking changes tend to not break much and be straightforward to upgrade through if necessary.

Compatibility is also the most common type of issue in TSDX by a fairly wide margin, so we generally do not want to make that scenario worse. That is why several of my RFCs and pending changes try to break out TSDX into more components that can be separately updated, better overrided, and even opted-out. But we're not there yet, so this kind of change is particularly early. Even if I wanted to merge this, TSDX itself is not ready or compatible with it.

This change is also _very_ breaking and _itself_ is not compatible with all of the other dependencies TSDX has that rely on TypeScript and rely on a specific version. #926 (comment) shows some of that particular issue -- it does not _solely_ bump TS.

As such, at this time I'll be closing this PR as won't fix. Maybe in the future TSDX will be ready for such a change, but it is not at this time.

> use resolution field in package.json ? just like [#926 (comment)](https://github.com/formium/tsdx/issues/926#issuecomment-751936109)

Yes, that's meant to be a workaround and is listed as such, not a permanent solution.

<hr>

`tslib` is also practically not used, as we transpile to ESNext only and then the rest with Babel. I can't remember if it's used by other deps, but I also wasn't the one that added it to the templates and it's not entirely clear why it's there (very possible it's a mistake as other issues have suggested so).

lbwa:
Ok. Thanks for your response and contributions. Your explanation is very helpful.

thany:
@lbwa 

> It's not an elgant solution, and more mental burden. The important thing is that only works with yarn, not npm.

See my [other comment](https://github.com/jaredpalmer/tsdx/issues/1121#issuecomment-1066599523) where I explain how it works fine with NPM.

PR #1057, opened 2021-07-06
feat: full ESM-CJS interop

ctjlewis:
WIP.

Does not block `require()` with a `"type": "module"` package.json field, uses `main`, `module`, and `exports` fields to support CJS and ESM module resolution in complicated contexts (provoked by errors in Metro with React Native).


src/templates/basic.ts
```diff
@@ -15,8 +15,13 @@ const basicTemplate: Template = {
...
     // author: author,
-    main: 'dist/index.js',
-    // module: `dist/${safeName}.esm.js`,
+    main: './dist/index.cjs',
+    module: './dist/index.mjs',
+    exports: {
+      import: './dist/index.mjs',
+      require: './dist/index.cjs',
+    },
+    // module: `dist/${safeName}.mjs`,
```
ctjlewis, line 24: Standardized entry points.

